PR #7, opened 2018-08-17
Python version and ledger fixes

instagibbs:
I was getting failure to detect p2sh inputs, please test that this still works for other versions. I'm not a python expert esp wrt bytes.

Second commit is fixing the various issues with it, including grabbing the wrong "master key" since ledger has no way of directly asking for xpubs. 

Manually tested against a single input transaction spending:
1) (p2sh-)p2wpkh
2) (p2sh-)pwsh: 2 of 3 multisig

And all worked. It correctly identified p2sh-p2wpkh change outputs as well.


instagibbs:
broke stuff in the rebase, oops. My copy was pretty old...

instagibbs:
Just tested on bigger transaction:
Inputs:
1) p2sh-p2wsh (2 of 3 multisig)
2) p2sh-p2wpkh
3) p2wpkh
Outputs:
1) p2wpkh
2) p2wpkh(change)

And everything worked again as expected.

PR #12, opened 2018-08-23
Updates and fixes to getkeypool and keypaths

achow101:
Instead of using the actual master seed id, use the fingerprint and pack that into a uint160. Otherwise wallets which do not support fetching the master pubkey or its id will not work. By packing it into a uint160, we can still ensure that the fingerprint is actually used.

Also allow using `h` and `H` to indicate hardened to avoid having to escape `'` in the terminal.

Built on top of #11 so I could test the thing.

### DIFF
--- a/hwi.py
+++ b/hwi.py
@@ -187,6 +187,10 @@ def process_commands():
         path_base = command_args[0]
         start = int(command_args[1])
         end = int(command_args[2])
+
+        if device_type == 'digitalbitbox':
+            if '\'' not in path_base and 'h' not in path_base and 'H' not in path_base:
+                print(json.dumps({'error' : 'The digital bitbox requires one part of the derivation path to be derived using hardened keys'}))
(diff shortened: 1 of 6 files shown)


hwi.py
```diff
@@ -184,11 +184,11 @@ def process_commands():
         # args[0]: path base (e.g. m/44'/0'/0')
         # args[1]; start index (e.g. 0)
         # args[2]: end index (e.g. 1000)
-        master_xpub = json.loads(client.get_pubkey_at_path('m/'))['xpub']
-        master_fpr = get_xpub_id(master_xpub)
         path_base = command_args[0]
         start = int(command_args[1])
         end = int(command_args[2])
+        master_xpub = json.loads(client.get_pubkey_at_path('m/0h'))['xpub']
```
Sjors, line 194: Rather late question: why the change from `m/` to `m/0h`?

PR #24, opened 2018-08-30
Derive the keypool locally except for the parent key

achow101:
Grab the parent pubkey from the device and derive the child keys locally
from the parent. This is much faster than fetching all of the keys from
the device.


instagibbs:
tACK, addresses and pubkeys are matching behavior in master

Would prefer no additional dependencies, but we can always swap it out later.

PR #27, opened 2018-09-01
display address functionality

instagibbs:
This helps partially validate a deposit address is valid.

Even better would be a signing challenge in addition to this, but we'd have to have `verifymessage` type support also baked in python.


instagibbs:
rebased, no time to test this time, so please do

PR #31, opened 2018-09-07
Enriched with examples to use the extracted xpub

gabridome:
The syntax of the BIP32 is tricky:
BIP32 utils wants "m" in the beginning.


gabridome:
> Instead of place holders, can you use an actual example?

Absolutely. I just have to configure a new ledger, so probably not today :P.

> Also, I think this would be better suited as it's own document instead of as part of the ledger instructions.

Like an examples.md in the doc directory?

docs/examples.md
```diff
@@ -0,0 +1,150 @@
+# Examples
+
+This Example has been taken with a Ledger nano.
+
+```
+./hwi.py enumerate
+[{"type": "ledger", "path": "IOService:/AppleACPIPlatformExpert/PCI0@0/AppleACPIPCI/XHC1@14/XHC1@14000000/HS02@14200000/Nano S@14200000/Nano S@0/IOUSBHostHIDDevice@14200000,0", "serial_number": "0001"}, {"type": "ledger", "path": "IOService:/AppleACPIPlatformExpert/PCI0@0/AppleACPIPCI/XHC1@14/XHC1@14000000/HS02@14200000/Nano S@14200000/Nano S@1/IOUSBHostHIDDevice@14200000,1", "serial_number": "0001"}]```
```
achow101: That's... interesting. On Ubuntu, the device paths are much much shorter. I see things like `0003:0001:00` for the path. I guess then that you should also say what OS you are using as the paths may be different depending on OS.

gabridome: Ok no problem. i also have to clean up the modified ledger.md.

gabridome:
I didn't know how to cancel my editing to Ledger.md so I removed them. If you want I could close and delete the PR and resubmit it if you tell me how.

achow101:
Just use `git rebase -i` to squash/drop/reorder your commits as you want and the force push (`git push -f`) the changes.

gabridome:
Thank you.

PR #33, opened 2018-09-08
Implement a debug mode and debug logging

achow101:
Have a switch for debug mode.

Debug statements should now only be printed when debug mode is enabled.


instagibbs:
ACK for ledger e5d302b55bc5923ba83d8abfe0ce663e6bf9bf94

PR #37, opened 2018-09-14
Report devices that return an error during enumeration

instagibbs:
Partially fixes https://github.com/achow101/HWI/issues/36 and should allow easier debugging in the future.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -98,12 +98,15 @@ def enumerate():
 def find_device(fingerprint):
     devices = enumerate()
     for d in devices:
-        client = get_client(d['type'], d['path'])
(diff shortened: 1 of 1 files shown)


Sjors:
This reduces the error message I'm getting from a long crash stack to just an `error` field with the problem description, so it seems an improvement.

instagibbs:
Pushed another fix so having an error opening a device while finding the prescribed fingerprint doesn't result in an erroneous error.

e.g., I want my Ledger to sign, but the Trezor is first in the list, so it throws an error rather than skipping over it.

achow101:
ACK 92cc16a407d9930ef27db0596efe2f95b7ccbc45

PR #40, opened 2018-10-03
Bitbox fixes

instagibbs:


instagibbs:
with these fixes I've been able to spend p2sh-p2wpkh and p2sh-p2wsh-multisig funds

achow101:
utACK

PR #41, opened 2018-10-05
HWI Examples: Updates, typos and simplification

jonathancross:
Nice project :-)

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,48 +1,47 @@
 # Examples
 
-This Example has been taken with a Ledger nano S.
+Example using a Ledger Nano S:
 
 ```
 ./hwi.py enumerate
 [{"type": "ledger", "path": "IOService:/AppleACPIPlatformExpert/PCI0@0/AppleACPIPCI/XHC1@14/XHC1@14000000/HS02@14200000/Nano S@14200000/Nano S@0/IOUSBHostHIDDevice@14200000,0", "serial_number": "0001"}, {"type": "ledger", "path": "IOService:/AppleACPIPlatformExpert/PCI0@0/AppleACPIPCI/XHC1@14/XHC1@14000000/HS02@14200000/Nano S@14200000/Nano S@1/IOUSBHostHIDDevice@14200000,1", "serial_number": "0001"}]
 ```
-The OS in this case is OSX (17.7.0 Darwin Kernel Version 17.7.0). In Linux the 
+The OS in this case is macOS v10.13.6  (Darwin Kernel Version 17.7.0). In Linux the
 "path" is shorter.
 
-## Useful xpubs to extract
+## Extracting xpubs
 
-Starting from version Bitcoin Core v.0.17, It is possible to to retrieve 
-the Unspent transaction outputs relevant for a set of [Output Descriptors][1] 
-From Bitcoin Core with the `scantxoutset` funcion.
+Bitcoin Core v0.17.0 and later allows you to retrieve the unspent transaction outputs (utxo)
+relevant for a set of [Output Descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md) with the `scantxoutset` RPC call.
 
-To retrieve the outputs relevant for a specific hardware wallet it is 
+To retrieve the outputs relevant for a specific hardware wallet it is
 necessary:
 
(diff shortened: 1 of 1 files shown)


docs/examples.md
```diff
@@ -122,8 +121,8 @@ bitcoin-cli scantxoutset start '[{"desc":"sh(wpkh(xpub6DP8WTA5cy2qWzdtjMUpLJHkzo
 {"xpub": "xpub6DP9afdc7qsz7s7mwAvciAR2dV6vPC3gyiQbqKDzDcPAq3UQChKPimHc3uCYfTTkpoXdwRTFnVTBdFpM9ysbf6KV34uMqkD3zXr6FzkJtcB"}
 ```
 
-2. With this xpub it is possible  extract the relevant UTXOs using the 
-`scantxoutset` in Bitcoin Core (from 0.17).
+2. With this xpub it is possible  extract the relevant UTXOs using the
+`scantxoutset` RPC call in Bitcoin Core.
```
instagibbs: it's only in 0.17, still pertinent to tell users

docs/examples.md
```diff
@@ -1,48 +1,47 @@
...
+The OS in this case is macOS v10.13.6  (Darwin Kernel Version 17.7.0). In Linux the
 "path" is shorter.
 
-## Useful xpubs to extract
+## Extracting xpubs
 
-Starting from version Bitcoin Core v.0.17, It is possible to to retrieve 
-the Unspent transaction outputs relevant for a set of [Output Descriptors][1] 
-From Bitcoin Core with the `scantxoutset` funcion.
+Bitcoin Core v0.17[¹](#bitcoin-core-version) allows you to retrieve the unspent transaction outputs (utxo)
```
achow101: Please don't use superscripts.

docs/examples.md
```diff
@@ -138,8 +137,9 @@ bitcoin-cli scantxoutset start '[{"desc":"wpkh(xpub6DP9afdc7qsz7s7mwAvciAR2dV6vP
 ```
 
 
-[1]: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md
-[2]: https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki
-[3]: https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki
+## Notes
+
+<a name="bitcoin-core-version" title="Note on Bitcoin Core version">1</a> "Bitcoin Core" refers to v0.17.x and later.
```
achow101: Please don't use html. Anchors like these should be doable with just markdown. Also, I'm not sure that we really need these anchors.

docs/examples.md
jonathancross: Am I understanding this correctly that you want it repeated on each of the 3 lines that mention `scantxoutset`? (as it used to be)

The purpose of the footnote was to reduce redundancy, but I can revert if you like.

docs/examples.md
jonathancross: > Please don't use html. Anchors like these should be doable with just markdown.

Unfortunatly GitHub Flavored Markdown [doesn't support footnotes](https://stackoverflow.com/a/25584585/1117929) (named anchors)

> Also, I'm not sure that we really need these anchors.

The idea was to reduce redundancy, but I can remove if preferred and just use inline links.

docs/examples.md
jonathancross: I'll address this after we know if footnotes are being removed or not.

docs/examples.md
instagibbs: maybe it's ok, fine with as-is I guess

docs/examples.md
jonathancross: I went ahead and replaced the superscript numbers with normal numbers.

docs/examples.md
jonathancross: @achow101 Do you want to leave this as-is or should I convert to inline links to the info?

achow101: Can you do inline links?

jonathancross:
@instagibbs  @achow101 Anything else you'd like to see here?

instagibbs:
LGTM?

achow101:
ACK 621d2bfffdd3e2aba3a6b30f896316450e72ab3a

PR #42, opened 2018-10-07
Replace bip32utils with our own BIP 32 public derivation

achow101:
bip32utils appears to no longer exist. To replace it, I have implemented my own BIP 32 public derivation function. We only need the public derivation as we just need to derive from xpubs for `getkeypool`.


instagibbs:
This will require someone more familiar with EC primitives to review the EC math.

Will test.

achow101:
> This will require someone more familiar with EC primitives to review the EC math.

`point_add`, `point_mul`, and `point_to_bytes` were taken from the Schnorr BIP.

`deserialize_point` was provided by sipa.

hwilib/commands.py
```diff
@@ -155,8 +155,8 @@ def getkeypool(args, client):
         else:
             path = path_base + '/' + str(i)
 
-        child = parent.ChildKey(i)
-        address = child.Address()
+        child_pk, child_cc = CKDpub(parent_pk, parent_cc, i)
+        address = pubkey_to_address(child_pk, args.testnet)
 
         this_import['pubkeys'] = [{binascii.hexlify(child.PublicKey()).decode() : {master_fpr : path.replace('\'', 'h')}}]
```
instagibbs: "child" no longer eixsts

instagibbs: pretty sure s/child.PublicKey()/child_pk/

instagibbs:
Ah.

tACK, with my noted issue fixed

hwilib/commands.py
achow101: Fixed

instagibbs:
tACK 2c2d1b49ac01f827b522a5ad24728f5754e2b02f

PR #43, opened 2018-10-07
Allow specifying a password in enumerate

achow101:
This lets us get the fingerprints from bitboxes


instagibbs:
utACK, should work for a single bitbox at a time

PR #47, opened 2018-11-05
Use importmulti descriptor support

Sjors:
Works with the latest 
https://github.com/achow101/bitcoin/tree/hww

Also gets rid of bip32 dependency.

Usage:
```sh
./hwi.py --testnet --device-type ledger getkeypool --keypool --bech32 "m/84'/1'/0'" "0/*"  0 10
[{"desc": "wpkh([01234567/84h/1h/0']tpub_for_account/0/*)", "range": [0, 10], "timestamp": "now", "internal": false, "keypool": true}]

./hwi.py --testnet --device-type ledger getkeypool --keypool --internal --bech32 "m/84'/1'/0'" "0/*"  0 10
...
```


hwilib/commands.py
```diff
@@ -210,9 +209,12 @@ def process_commands(args):
     getkeypol_parser = subparsers.add_parser('getkeypool', help='Get JSON array of keys that can be imported to Bitcoin Core with importmulti')
     getkeypol_parser.add_argument('--internal', action='store_true', help='Indicates that the keys are change keys')
     getkeypol_parser.add_argument('--keypool', action='store_true', help='Indicates that the keys are to be imported to the keypool')
-    getkeypol_parser.add_argument('path_base', help='The prefix of the derivation path')
-    getkeypol_parser.add_argument('start', type=int, help='The index to start at. The first key will be <path_base>/<start>')
-    getkeypol_parser.add_argument('end', type=int, help='The index to end at. The last key will be <path_base>/<end>')
+    getkeypol_parser.add_argument('path_base', help='The prefix of the derivation path, e.g. m/84h/0h/0h for account 0 native SegWit')
```
instagibbs: Can we just call it `key_origin` or something since we're making the split?

Sjors: I didn't want to change too much, and the fingerprint isn't in it... We can probably overhaul the command later, e.g. I'd prefer a one-liner that handles both change and receive and just takes two full descriptors.

instagibbs: It's partial key origin info, hence the argument split. Confusing as-is for me.

instagibbs:
light tACK 97e4e369a2b34e7b36506a7dbdd73f16a0758f87

achow101:
Before merging this, I would like to have rebased versions of the mentioned Bitcoin Core PRs so that there is actually working code that people can use. I'll be updating (and probably rewriting) my two PRs (the import to keypool and import key origin data) soon (hopefully). Then I will review this.

Sjors:
I have a work in progress branch here: https://github.com/Sjors/bitcoin/commits/2018/11/watch-only-keypool

It still can't create a valid psbt though, see my comment: https://github.com/bitcoin/bitcoin/pull/14021#issuecomment-436193667

instagibbs:
@Sjors just noting that your WIP branch, when importing keys to keypool, seems to enter keys into the keypool in a random-looking order. First key being index 216, then 307, etc.

Sjors:
@instagibbs I noticed that too. Wasn't sure if I screwed something up during manual rebase.

instagibbs:
It might be related to how descriptors are expanded and entered, versus the separate key entries. Still... already a nicer experience otherwise. Was easily able to do 1000 keys internal and external in a few seconds.

If I get some time today I can take a deeper dive.

achow101:
I've rebased all of my PRs and updated my [hww](https://github.com/achow101/bitcoin/tree/hww) branch which has them all merged.

hwilib/commands.py
```diff
@@ -210,9 +209,12 @@ def process_commands(args):
...
+    getkeypol_parser.add_argument('path_suffix', help='The suffix of the derivation path, e.g. 0/* for receive and 1/* for change')
```
achow101: What is the point of splitting the `path_base` into both `path_base` and `path_suffix`? The `path_base` can be used for both.

instagibbs:
Ok I think I know what is happening: `ImportData` is just a `std::set` structure, which means when you `Expand` your descriptor, they're added in some sort of sorting order, rather than insertion-order.

Probably too cute, but we could make a custom comparator that sorts based on the range?

hwilib/commands.py
```diff
@@ -151,25 +151,24 @@ def getkeypool(args, client):
...
+
+    descriptor_open = 'pkh('
+    descriptor_close = ')'
+    if args.bech32 == True:
+          descriptor_open = 'wpkh('
+    elif args.p2sh_p2wpkh == True:
+          descriptor_open = 'sh(pkh('
+          descriptor_close = '))'
+
+    this_import['desc'] = descriptor_open + '[' + master_fpr[-8:] + '/' + path_base.replace('\'','h').replace('m/', '') + ']' + base_key.replace('\'','h') + '/' + path_suffix + descriptor_close
```
achow101: This is wrong. `master_fpr` has the fingerprint in reversed byte order due to the weirdness with uint160's. With descriptors, they should be put in the correct order as descriptors does not have that issue. You should also change the helper function to just give the fingerprint and not a uint160 so we don't need the `[-8:]` here.

Sjors: Fixed

Sjors:
Using your new `hww` branch, I'm still getting an error when trying to parse the BIP32 paths. Here's my steps:

```
export RECEIVE=`./hwi.py --device-type=ledger --testnet getkeypool --bech32 --keypool "m/84h/1h/0h" "0/*" 0 100`
export CHANGE=`./hwi.py --device-type=ledger --testnet getkeypool --bech32 --keypool --internal "m/84h/1h/0h" "1/*" 0 100`
export DESTINATION=`bitcoin-cli -rpcwallet=A getnewaddress`
bitcoin-cli createwallet "HWW" true
bitcoin-cli -rpcwallet=HWW importmulti '$RECEIVE'
bitcoin-cli -rpcwallet=HWW importmulti '$CHANGE'
bitcoin-cli -rpcwallet=HWW walletcreatefundedpsbt '[]' '[{"$DESTINATION": 0.01}]' 0
bitcoin-cli decodepsbt "the above psbt"
error code: -22
error message:
TX decode failed Size of key was not the expected size for the type BIP32 keypath: unspecified iostream_category error
```

instagibbs:
Nice! BIP-compat as default is a good idea.

Sjors:
With an earlier wallet I'm noticing `walletcreatefundedpsbt ... true` no longer adds bip32 derivation, even though that info is present when you do `getaddressinfo`. Don't we need some sort of change to the PSBT function to use the new `hdmasterkeyid` field?

instagibbs:
IIRC with watchonly type stuff often PSBT stuff would require the private key in the wallet in order to generate various path information. With my branches I had to support pubkey-only.

Sjors:
I've been able to create PSBTs with derivation info in the past without private keys in the wallet.

instagibbs:
Ok then I'm thinking of much older branches

instagibbs:
The underlying provider there `m_provider` is the actual `CWallet`.

hwilib/base58.py
```diff
@@ -79,7 +79,7 @@ def get_xpub_fingerprint_hex(xpub):
 def get_xpub_fingerprint_as_id(xpub):
     data = decode(xpub)
     fingerprint = data[5:9] + b'\x00' * 16
-    return hexlify(fingerprint[::-1]).decode()
+    return hexlify(fingerprint[::-1]).decode()[-8:][::-1]
```
instagibbs: This literally flips the string, not the byte ordering?

instagibbs:
Looks like the fingerprint hex chars, within each byte, are reversed, not the actual bytes themselves?
```
./hwi.py -f ad4c1514 getkeypool 0 1
[{"timestamp": "now", "range": [0, 1], "internal": false, "desc": "pkh([dac45141/44h/0h]xpub6ATogMCQsGJsht8SBGFTNmafpVpprv5eWF1nVKYqgRrNaRCNrMTwL87N6SdTiEC4ZPf6VfcjUxdjxLj3iRMBTkQDEVYwtKiEZtZdutZUFVr/0/0/*)", "keypool": false}]
```

Sjors:
Fixed the fingerprint and removed the fancy python 3.7 syntax.

hwilib/commands.py
```diff
@@ -136,40 +134,79 @@ def signmessage(args, client):
...
+
+      # Purpose
+      if args.bech32 == True:
+        path += "84'/"
+      elif args.p2sh_p2wpkh == True:
+        path += "49'/"
+      else:
+        path += "44'/"
+
+      # Coin type
```
instagibbs: which BIP is this?

Sjors:
I'm still getting the [above error](https://github.com/achow101/HWI/pull/47#issuecomment-436590495). I updated the commands to use the latest syntax. One thing that's perhaps worth noting about the wallet I'm importing is that the PSBT ends up spending from a change address `m/84'/1'/0'/1/0` (and sends change to `m/84'/1'/0'/1/98`, but that's random).

We also still need to deal with the randomised order for the keypool, as that violates the gap limit rule for other wallets.

instagibbs:
Stray thought for future possibly: Would be nice to have it return  both internal and external descriptors in one shot, especially if using a BIP standard derivation path.

hwilib/base58.py
```diff
@@ -79,7 +79,7 @@ def get_xpub_fingerprint_hex(xpub):
 def get_xpub_fingerprint_as_id(xpub):
     data = decode(xpub)
     fingerprint = data[5:9] + b'\x00' * 16
-    return hexlify(fingerprint[::-1]).decode()
+    return hexlify(fingerprint).decode()[:8]
```
achow101: I think it is unnecessary to have `[:8]` here since it will always be 8 characters anyways.

hwilib/commands.py
achow101: Ah, I see. Then in that case, this would be fine.

instagibbs:
```
pkh([14154cad/m]tpubD6NzVbkrYhZ4WcK3ZNuDevuPvG1r2h9svYa6UkzdbtSKNRoyH9wopTnKUFbRPJH23roz2g4rfLSRreVkatdw8wfNsgMd1xdUARiFGHhk2Et/44/0/0/*)
```
the `/m]` part where there's no hardened derivation to the xpub is considered invalid by Core fwiw.

instagibbs:
Make sure you arent using testnet. It gives very vague error for version
byte of xpub.

achow101:
@Sjors I am still unable to replicate your issue. I am able to create a psbt, sign it with `./hwi.py signtx`, finalize it, and send it.

hwilib/base58.py
Sjors: When I remove it, it adds lots of zeros at the end.

Sjors:
Weird. Did you use testnet, bech32 and import a previously used wallet? Was the destination an address from a different wallet?

Spending from a freshly funded address vs. spending from an imported address doesn't make a difference for me.

If I send to an _inside the wallet_ it works fine.

@instagibbs I fixed the `m` with no hardened derivation issue.

achow101:
@Sjors https://github.com/bitcoin/bitcoin/pull/14689 should fix the problem you were having. I've also updated the hww branch with it.

hwilib/base58.py
achow101: Remove `+ b'\x00' * 16` from the line above

Sjors:
Indeed it does. Also note that `generatenewaddress` created a `bech32` address on my system due to the defaults.

hwilib/commands.py
Sjors: Yes, there's a list of coin types in BIP 44: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#registered-coin-types

That includes mainnet as type 0 and testnet as type 1.

hwilib/base58.py
Sjors: Done

achow101:
tACK 44b4496dcd7b383b3cfe24c73b9ed9948053bb4e

instagibbs:
tACK https://github.com/achow101/HWI/pull/47/commits/44b4496dcd7b383b3cfe24c73b9ed9948053bb4e

Did a few different round-trips on ledger, checked all other commands as well as sanity check.

Issue #53, opened 2018-11-09
`combo` support


instagibbs:
In order for us to properly discard keys we've already seen when we have a mix of output types even if the imported descriptor is `wpkh`, for example, we need combo support. While the "ISMINE" logic makes sure that all funds are seen during blockchain scan, related infrastructure such as `CAffectedKeyVisitor` does not, meaning https://github.com/achow101/bitcoin/commit/5958766d4c97dd5339aaf77b828171e2ed8ef738 doesn't really work properly.

instagibbs:
If you run Core currently you already do sadly.

Second, my wallet literally has mixed output types due to change type logic.

instagibbs:
Turns out we need a small fix on the importmulti refactor to make this useful, otherwise it fails to add all the destinations to the watchonly set: https://github.com/bitcoin/bitcoin/pull/14565/files#r232306261

instagibbs:
Don't need this, if we want more in the future let's just support descriptor parsing

PR #54, opened 2018-11-09
add --combo argument to create combo descriptors that respect address…

instagibbs:
… type default derivation paths

Also disallow both `bech32` and `p2sh_p2wpkh` options in `getkeypool`.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -195,7 +195,10 @@ def getkeypool(args, client):
 
     descriptor_open = 'pkh('
     descriptor_close = ')'
-    if args.bech32 == True:
+    if args.combo == True:
+          descriptor_open = 'combo('
+          descriptor_close = ')'
+    elif args.bech32 == True:
           descriptor_open = 'wpkh('
     elif args.p2sh_p2wpkh == True:
           descriptor_open = 'sh(pkh('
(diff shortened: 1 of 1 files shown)


Sjors:
See #53 for discussion.

The "Also disallow both `bech32` and `p2sh_p2wpkh` options in `getkeypool`" bit is useful regardless, maybe make another PR for that?

instagibbs:
Didn't expect supporting a standard descriptor type to be controversial, but I can split it out sure

instagibbs:
took out the warning for dual-address type options

Sjors:
Not _that_ controversial, but this is a small repo :-)

 I'm confused what you mean by "Can be used in conjunction with `--p2sh_p2wpkh` or `--bech32` to use their default derivation paths'". By that you mean it use 44' by default, otherwise 49' or 84'? 

Doesn't Bitcoin Core have it's own derivation path? Maybe use that? Downside is that it only uses hardened, so you'd have to split out a long list of descriptors.

I'm worried that if you use any of 44' / 49' / 84' then a user recovering their mnemonic in any standard-ish wallet will see only partial funds. Does it make sense for `--combo` to split out three seperate descriptors instead, each with the correct derivation path? Or is `combo()` doing the magic we need?

Sjors:
Come to think of it, IIUC this is purely to recover previously imported wallets? So in that case all keys would be under the same 44',  49' or 84' path depending on what you used back then. I'm tempted to suggest that we require `--path` in that case, just to reduce confusion.

instagibbs:
I realized this is just me being stupid thanks to Pieter: https://github.com/instagibbs/bitcoin/commit/63baf703cba4b5dfa2c0307387b1a000d0e1bb3d

ta-dah, no need for `combo`. Closing.

PR #56, opened 2018-11-09
update and add error for mixing address type options

instagibbs:
split out from https://github.com/achow101/HWI/pull/54


achow101:
Instead of having flags for each address type, maybe we should make the flag `--addresstype` and do a `choices` here?

Sjors:
utACK 8f3edf3 as it's a strict improvement, but I like @achow101's suggestion better in the long run.

achow101:
utACK 8f3edf38de42f7bcd6023717c736e446239b9f8e

PR #62, opened 2018-11-25
Fix client creation bugs

### LINKED ISSUE #59: Commands don't work for me anymore.
I have a strange behaviour:
```
./hwi.py -d "IOService:/AppleACPIPlatformExpert/PCI0@0/AppleACPIPCI/XHC1@14/XHC1@14000000/HS02@14200000/Nano S@14200000/Nano S@1/IOUSBHostHIDDevice@14200000,1" -t "ledger" getmasterxpub

{"error": "You must specify a device type or fingerprint for all commands except enumerate", "code": -1}
```
OS: 18.0.0 Darwin Kernel Version 18.0.0: Wed Aug 22 20:13:40 PDT 2018; root:xnu-4903.201.2~1/RELEASE_X86_64 x86_64

commit: b10bfe6680205df55a3670db44e9bad7d07f84d0

**EDIT:** curiously with "--fingerprint" it works.

achow101:
Fixes a bug where specifying device path and type instead of fingerprint would not work. Fixes #59.

Return the connection error message from `get_client()` instead of printing it. Additionally, detect when `get_client()` fails and error on that instead of having something else error when a non-client object is returned.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -22,16 +22,20 @@ INVALID_TX = -5
 NO_PASSWORD = -6
 BAD_ARGUMENT = -7
 
+class NoPasswordError(Exception):
+    def __init__(self,*args,**kwargs):
+        Exception.__init__(self,*args,**kwargs)
+
+class UnknownDeviceError(Exception):
+    def __init__(self,*args,**kwargs):
+        Exception.__init__(self,*args,**kwargs)
+
 # Get the client for the device
 def get_client(device_type, device_path, password=None):
     # Open the device
-    try:
-        device = hid.device()
-        device_path = bytes(device_path.encode())
-        device.open_path(device_path)
-    except Exception as e:
-        print(e)
-        return {'error':'Unable to connect to specified device','code':DEVICE_CONN_ERROR}
+    device = hid.device()
+    device_path = bytes(device_path.encode())
+    device.open_path(device_path)
 
     # Make a client
     if device_type == 'trezor':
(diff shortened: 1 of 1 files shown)


hwilib/commands.py
```diff
@@ -30,8 +30,7 @@ def get_client(device_type, device_path, password=None):
         device_path = bytes(device_path.encode())
         device.open_path(device_path)
     except Exception as e:
-        print(e)
-        return {'error':'Unable to connect to specified device','code':DEVICE_CONN_ERROR}
+        return {'error':'Unable to connect to specified device:' + str(e),'code':DEVICE_CONN_ERROR}
```
instagibbs: not a Pythonese expert, but is there a reason why we can't just throw an error here instead of returning a dict, and handling it as appropriate by the caller?

achow101: Because that would be the logical thing to do...


Done.

instagibbs:
much easier to think about, concept ACK. Will test.

hwilib/commands.py
```diff
@@ -285,10 +289,18 @@ def process_commands(args):
         client = find_device(args)
         if not client:
             return {'error':'Could not find device with specified fingerprint','code':DEVICE_CONN_ERROR}
-    else:
+    elif not args.device_type:
         return {'error':'You must specify a device type or fingerprint for all commands except enumerate','code':NO_DEVICE_PATH}
 
+    try:
```
instagibbs: this whole block should only be called if none of the previous above branches are hit. If the user doesn't give the device path they get: `{"error": "'NoneType' object has no attribute 'encode'", "code"`.

achow101: I've moved it into it's own `elif` block which checks for device type and path.

instagibbs:
tACK

PR #65, opened 2018-11-26
dont use Ledger's getDongle, to use properly specified device path

instagibbs:
Currently specifying the device path doesn't really do anything aside from error if that device path specified doesn't exist at all.

As another side-effect is enumerate shows an error for the keyboard HID instead of falsely reporting a fingerprint for it:

```
[{"path": "0001:XXX:00", "type": "ledger", "serial_number": "0001", "fingerprint": "deadbeef"}, {"error": "Could not open client or get fingerprint information: Exception : Invalid sequence", "path": "0001:XXX:01", "type": "ledger", "serial_number": "0001"}]
```

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -72,8 +72,10 @@ def enumerate(args):
         # Get keepkeys
         elif (d['vendor_id'], d['product_id']) in keepkey_device_ids:
             d_data['type'] = 'keepkey'
-        # Get ledgers
-        elif (d['vendor_id'], d['product_id']) in ledger_device_ids:
+        # Get ledgers, only take interface 0 to avoid HID keyboard
+        elif (d['vendor_id'], d['product_id']) in ledger_device_ids and \
+                ('interface_number' in d and  d['interface_number'] == 0 \
+                or ('usage_page' in d and d['usage_page'] == 0xffa0)):
             d_data['type'] = 'ledger'
         # Get DigitalBitboxes
         elif (d['vendor_id'], d['product_id']) in digitalbitbox_device_ids:
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -36,6 +36,7 @@ def get_client(device_type, device_path, password=None):
     device = hid.device()
     device_path = bytes(device_path.encode())
     device.open_path(device_path)
+    device.set_nonblocking(True)
 
     # Make a client
     if device_type == 'trezor':
--- a/hwilib/ledgeri.py
+++ b/hwilib/ledgeri.py
@@ -18,9 +18,7 @@ class LedgerClient(HardwareWalletClient):
     # hacked in device support using btchip-python
     def __init__(self, device):
         super(LedgerClient, self).__init__(device)
-        # hack to use btchip-python's getDongle pipeline
-        device.close()
-        self.dongle = getDongle(logging.getLogger().getEffectiveLevel() == logging.DEBUG)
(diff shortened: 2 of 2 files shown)


instagibbs:
When commands are put towards the keyboard HID I get:
`btchip.btchipException.BTChipException: Exception : Invalid sequence` raised. Should we capture this error and print something prettier?

instagibbs:
Pushed an additional check to ignore the keyboard HID during enumeration. `btchip-python` does this exact check.

hwilib/commands.py
```diff
@@ -71,8 +72,8 @@ def enumerate(args):
         # Get keepkeys
         elif (d['vendor_id'], d['product_id']) in keepkey_device_ids:
             d_data['type'] = 'keepkey'
-        # Get ledgers
-        elif (d['vendor_id'], d['product_id']) in ledger_device_ids:
+        # Get ledgers, only take interface 0 to avoid HID keyboard
+        elif (d['vendor_id'], d['product_id']) in ledger_device_ids and d['interface_number'] == 0:
```
achow101: btchip-python also checks for `usage_page` and checks that both `interface_number` and `usage_page` exist before getting their values. I think we should do that as well.

hwilib/ledgeri.py
```diff
@@ -18,9 +18,7 @@ class LedgerClient(HardwareWalletClient):
     # hacked in device support using btchip-python
     def __init__(self, device):
         super(LedgerClient, self).__init__(device)
-        # hack to use btchip-python's getDongle pipeline
-        device.close()
-        self.dongle = getDongle(logging.getLogger().getEffectiveLevel() == logging.DEBUG)
+        self.dongle = HIDDongleHIDAPI(device, True, logging.getLogger().getEffectiveLevel() == logging.DEBUG)
```
achow101: Is the second parameter always `True`? Looking at btchip-python, it seems that there are two device id pairs that have it set to `False`. Should we take those into account?

hwilib/commands.py
instagibbs: I'm not sure what `user_page` is for, as I've only seen it as 0.

instagibbs: fixed for interface number

hwilib/ledgeri.py
instagibbs: I believe those are non-nano s, I'll look.

instagibbs: https://github.com/greenaddress/GreenBits/blob/master/app/src/main/res/xml/device_filter.xml

achow101: Right now we will accept non-nano s ledger devices.

hwilib/commands.py
achow101: I found [this comment](https://github.com/digitalbitbox/mcu/blob/master/py/dbb_utils.py#L73) while looking through some Digital BitBox stuff:

```
# hidapi is not consistent across platforms
# usage_page works on Windows/Mac; interface_number works on Linux
```

So I think we still need the `usage_page` check

hwilib/ledgeri.py
instagibbs: They use a completely different API, for many calls. I'd be shocked if anything worked.

hwilib/commands.py
instagibbs: ah, fixed!

achow101:
tACK f45b8e7119ec276c6bafc90df4e2c2e9f18ffd5e

PR #66, opened 2018-11-26
Move device enumeration to device specific drivers and load device drivers without explicit import

### LINKED ISSUE #55: WebUSB support (e.g. Trezor model T)
The Trezor model T uses [webusb](https://wicg.github.io/webusb/) to communicate (see https://github.com/achow101/HWI/issues/38#issuecomment-427032492). I believe Ledger Nano S [supports U2F](https://github.com/LedgerHQ/ledger-wallet-chrome/issues/88#issuecomment-363222976).

I'm not entirely clear on the differences between webusb and U2F.

I'm guessing trying and studying https://github.com/trezor/python-trezor would be a good place to start.

achow101:
This PR moves device enumeration from the `enumerate()` function to `enumerate()` functions within each device driver. These device specific `enumerate()` functions allow us to enumerate those devices with whatever device specific enumeration required, instead of only enumerating with hid. This allows us to interact with devices connected through other methods such as udp (trezor emulator), webusb (trezor t), unix pipes (coldcard emulator), etc. but without requiring us to add that enumeration to the actual `enumerate()` function. Because there can be different types of device interfaces, the creation of the device is now handled by the device client class. The device path will be passed in and the client class uses that to know what type of device interface to create.

Furthermore, device drivers have been moved to `hwilib/devices/` and both `getclient()` and `enumerate()` will call their respective functions and constructors without explicitly importing the device specific drivers. This allows us to easily add a new device driver by simply adding its implementation to the `hwilib/devices/` folder. In order to facilitate this automatic discovery and import of device drivers, the client class names and arguments have been normalized. The class name must be the device name (as will be specified in the `--device-type` parameter), begin with a capital letter, have the rest of the device name in lower case, and followed by `Client`. Classes must also take two positional arguments `path` and `password`, in that order.

This change should enable the use of the Trezor T (so fixes #55) as well as device simulators and emulators so that automated tests can be done.


hwilib/commands.py
```diff
@@ -32,30 +32,24 @@ def __init__(self,*args,**kwargs):
 
 # Get the client for the device
 def get_client(device_type, device_path, password=None):
-    # Open the device
```
instagibbs: don't think these lines should be deleted yet in this commit?

achow101: Moved to "Unify device client constructor to path and password "

hwilib/devices/ledger.py
```diff
@@ -1,26 +1,33 @@
...
 import binascii
 import logging
 
+LEDGER_VENDOR_ID = 0x2c97
+LEDGER_DEVICE_ID = 0x0001
+
 # This class extends the HardwareWalletClient for Ledger Nano S specific things
 class LedgerClient(HardwareWalletClient):
 
     # device is an HID device that has already been opened.
```
instagibbs: you can delete these stale lines, especially now the initialization is done locally.

hwilib/devices/coldcard.py
```diff
@@ -106,3 +115,34 @@ def wipe_device(self):
...
+            master_xpub = client.get_pubkey_at_path('m/0h')['xpub']
+            d_data['fingerprint'] = get_xpub_fingerprint_hex(master_xpub)
+            client.close()
+        except Exception as e:
+            d_data['error'] = "Could not open client or get fingerprint information: " + str(e)
+
+        results.append(d_data)
+    # Check if the simulator is there
+    try:
+        client = ColdcardClient(CC_SIMULATOR_SOCK)
```
instagibbs: Can a tutorial on setting up the simulator be linked?

achow101: Some documentation for it is here: https://github.com/Coldcard/firmware/blob/master/unix/README.md

Although I haven't quite figured it out for myself either.

I added a comment with a link to the docs.

hwilib/commands.py
```diff
@@ -22,82 +24,35 @@
...
-        except Exception as e:
-            d_data['error'] = "Could not open client or get fingerprint information: " + str(e)
-            pass
+    # Gets the module names of all the files in devices/
+    files = glob.glob(dirname(__file__)+"/devices/*.py")
+    modules = [ basename(f)[:-3] for f in files if isfile(f) and not f.endswith('__init__.py')]
 
-        result.append(d_data)
+    for module in modules:
+        imported_dev = importlib.import_module('.devices.' + module, __package__)
```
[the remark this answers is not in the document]
achow101: Done. the errors should be suppressed now.

hwilib/devices/ledger.py
achow101: Removed

instagibbs:
Ok Ledger stuff at least works now.

hwilib/commands.py
```diff
@@ -53,41 +53,12 @@ def get_client(device_type, device_path, password=None):
 # Get a list of all available hardware wallets
 def enumerate(args):
```
Sjors: It would be nice if `enumerate` didn't ignore `--device-type=` as it's a bit slow.

instagibbs: I like the suggestion but obviously should wait until this is merged

hwilib/devices/trezor.py
```diff
@@ -195,3 +193,22 @@ def wipe_device(self):
     # Close the device
     def close(self):
         self.client.close()
+
```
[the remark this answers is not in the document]

PR #70, opened 2018-11-29
Increase default logging level to warning

achow101:
`trezorlib` outputs stuff with logging level info which results in fairly verbose output. That output is not particularly useful, so increase the logging level to warning to suppress that output.


instagibbs:
utACK though the ledger is less verbose, makes sense to only support warn/debug.

PR #73, opened 2018-12-03
WIP: descriptor request using "getkeys"

Sjors:
Todo:
* implement and discuss descriptor requests in Bitcoin Core
* rebase on #117

Implements `getkeys` as proposed [here](https://github.com/Sjors/bitcoin/blob/2018/11/rpc-signer/doc/external-signer.md#getkeys-optional).

_Update 2019/02/02_: this used to include #117, which I've moved to a new PR

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -196,6 +196,20 @@ def displayaddress(client, path=None, desc=None, sh_wpkh=False, wpkh=False):
             return {'error':'Descriptor missing origin info: ' + desc,'code':BAD_ARGUMENT}
         client.display_address(descriptor.m_path, descriptor.sh_wpkh, descriptor.wpkh)
 
+def getkeys(client, desc=None):
+    descriptor = Descriptor.parse(desc, client.is_testnet)
+
+    if descriptor is None:
+        return {'error':'Unable to parse descriptor: ' + desc,'code':BAD_ARGUMENT}
+    if descriptor.m_path is None:
+        return {'error':'Descriptor missing origin info: ' + desc,'code':BAD_ARGUMENT}
+
+    # Get the key at the base
+    # TODO: if the last path component is hardened, obtain a list of keys
+    descriptor.base_key = client.get_pubkey_at_path(descriptor.m_path_base)['xpub']
+
+    return [descriptor.serialize()]
+
 def setup_device(client, label='', backup_passphrase=''):
     try:
         return client.setup_device(label, backup_passphrase)
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -9,6 +9,7 @@ from .serializations import PSBT, Base64ToHex, HexToBase64, hash160
 from .base58 import xpub_to_address, xpub_to_pub_hex, get_xpub_fingerprint_as_id, get_xpub_fingerprint_hex
 from os.path import dirname, basename, isfile
 from .hwwclient import NoPasswordError, UnavailableActionError, DeviceAlreadyInitError
+from .descriptor import Descriptor
 
 # Error codes
 NO_DEVICE_PATH = -1
@@ -182,10 +183,18 @@ def getkeypool(client, path, start, end, internal=False, keypool=False, account=
     import_data.append(this_import)
     return import_data
 
-def displayaddress(client, path, sh_wpkh=False, wpkh=False):
-    if sh_wpkh == True and wpkh == True:
-        return {'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT}
-    return client.display_address(path, sh_wpkh, wpkh)
+def displayaddress(client, path=None, desc=None, sh_wpkh=False, wpkh=False):
+    if path is not None:
+        if sh_wpkh == True and wpkh == True:
+            return {'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT}
+        client.display_address(path, sh_wpkh, wpkh)
+    elif desc is not None:
+        descriptor = Descriptor.parse(desc, client.is_testnet)
+        if descriptor is None:
+            return {'error':'Unable to parse descriptor: ' + desc,'code':BAD_ARGUMENT}
+        if descriptor.m_path is None:
+            return {'error':'Descriptor missing origin info: ' + desc,'code':BAD_ARGUMENT}
+        client.display_address(descriptor.m_path, descriptor.sh_wpkh, descriptor.wpkh)
 
 def setup_device(client, label='', backup_passphrase=''):
     try:
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -1,0 +1,89 @@
+import re
+
+class Descriptor:
+    def __init__(self, origin_fingerprint, origin_path, base_key, path_suffix, testnet, sh_wpkh, wpkh, is_request):
+        self.origin_fingerprint = origin_fingerprint
+        self.origin_path = origin_path
+        self.path_suffix = path_suffix
+        self.base_key = base_key
+        self.testnet = testnet
+        self.sh_wpkh = sh_wpkh
+        self.wpkh = wpkh
+        self.m_path = None
+        self.is_request = is_request
+
+        if origin_path:
+            self.m_path = "m" + origin_path + (path_suffix or "")
+
+            if is_request:
+                self.m_path_base = "m" + origin_path
+
+    @classmethod
+    def parse(cls, desc, testnet = False):
+        sh_wpkh = None
+        wpkh = None
+        origin_fingerprint = None
+        origin_path = None
+        base_key_and_path_match = None
+        base_key = None
+        is_request = False
+
+        if desc.startswith("sh(wpkh("):
+            sh_wpkh = True
+        elif desc.startswith("wpkh("):
+            wpkh = True
+
+        origin_match = re.search(r"\[(.*)\]", desc)
+        if origin_match:
+            origin = origin_match.group(1)
+            match = re.search(r"^([0-9a-fA-F]{8})(\/.*)", origin)
+            if  match:
+              origin_fingerprint = match.group(1)
+              origin_path = match.group(2)
+              # Replace h with '
+              origin_path = origin_path.replace('h', '\'')
+
+            base_key_and_path_match = re.search(r"\[.*\](\w+)([\/\)][\d'\/\*]*)", desc)
+        else:
+            base_key_and_path_match = re.search(r"\((\w+)([\/\)][\d'\/\*]*)", desc)
+
+        if base_key_and_path_match:
+            base_key = base_key_and_path_match.group(1)
+            path_suffix = base_key_and_path_match.group(2)
+            if path_suffix == ")":
+                path_suffix = None
+        else:
+            if origin_match == None:
+                return None
+            else:
+                # Check if this is a descriptor request, which does not contain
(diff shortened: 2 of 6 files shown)


achow101:
Needs rebase as #66 has been merged.

Sjors:
@achow101 rebased, but I'm not happy with `signtransaction` just yet. Right now it requires a `fingerprint` argument. Ideally this should be inferred from the PSBT, but I have no idea how. Also it currently hardcodes `Testnet`, which I would also like to infer from the PSBT.

Sjors:
I removed the code that inferred Testnet from the BIP32 derivation, instead relying on `--testnet` being set.

hwilib/descriptor.py
```diff
@@ -0,0 +1,60 @@
...
+        #     if testnet and match.group(2) != "1":
+        #         throw...
+
+        path = re.search(r"(m\/[\dh\/\*]*)", desc.replace(fingerprint, "m")).group(1)
+
+        return cls(fingerprint, path, testnet, sh_wpkh, wpkh)
+
+
+    def serialize(self):
+      descriptor_open = 'pkh('
```
achow101: nit: spacing in this function

hwilib/descriptor.py
```diff
@@ -0,0 +1,60 @@
...
+            break
+          self.path_suffix = '/' + component + self.path_suffix
+        self.path_base = path.rsplit(self.path_suffix)[0]
+
+        self.base_key = None
+
+    @classmethod
+    def parse(cls, desc, testnet = False):
+        sh_wpkh = None
+        wpkh = None
```
achow101: These two are only booleans, so should be initialized to `False`, not `None`?

hwilib/commands.py
```diff
@@ -226,9 +231,11 @@ def process_commands(args):
     getkeypool_parser.set_defaults(func=getkeypool)
 
     displayaddr_parser = subparsers.add_parser('displayaddress', help='Display an address')
-    displayaddr_parser.add_argument('path', help='The BIP 32 derivation path of the key embedded in the address')
-    displayaddr_parser.add_argument('--sh_wpkh', action='store_true', help='Display the p2sh-nested segwit address associated with this key path')
-    displayaddr_parser.add_argument('--wpkh', action='store_true', help='Display the bech32 version of the address associated with this key path')
+    displayaddr_parser.add_argument('--desc', help='Pseudo-descriptor for which to return a descriptor with keys, e.g. wpkh(00000000/84h/1h/0h/0/*). Alternatively, use --path with --sh_wpkh / --wpkh')
+    displayaddr_parser.add_argument('--path', help='Derivation path, default follows BIP43 convention, e.g. m/84h/0h/0h/1/*')
```
achow101: You can use a [mutually exclusive group](https://docs.python.org/3/library/argparse.html#mutual-exclusion) here that is also required so that argparse can handle errors in argument format for us.

hwilib/commands.py
```diff
@@ -178,9 +179,13 @@ def getkeypool(args, client):
...
-    if args.sh_wpkh == True and args.wpkh == True:
-        return json.dumps({'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT})
-    return client.display_address(args.path, args.sh_wpkh, args.wpkh)
+    if args.desc != None:
+      key = Descriptor.parse(args.desc, args.testnet)
+      return client.display_address(key.path, key.sh_wpkh, key.wpkh)
+    elif args.path != None:
+      return client.display_address(args.path, args.sh_wpkh, args.wpkh)
+    else:
+      return {'error':'Either --descriptor or --path must be set','code':BAD_ARGUMENT}
```
achow101: The option is `--desc`, not `--descriptor`.

Using a mutually exclusive group removes the need for this error.

hwilib/commands.py
```diff
@@ -178,9 +179,13 @@ def getkeypool(args, client):
     return import_data
 
 def displayaddress(args, client):
-    if args.sh_wpkh == True and args.wpkh == True:
-        return json.dumps({'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT})
-    return client.display_address(args.path, args.sh_wpkh, args.wpkh)
+    if args.desc != None:
+      key = Descriptor.parse(args.desc, args.testnet)
```
achow101: nit: spacing in these if blocks.

hwilib/commands.py
```diff
@@ -245,6 +252,10 @@ def process_commands(args):
     if command == 'enumerate':
         return args.func(args)
 
+    if hasattr(args, "desc") and args.desc:
+      descriptor = Descriptor.parse(args.desc, args.testnet)
```
achow101: nit: spacing

hwilib/commands.py
```diff
@@ -217,6 +217,9 @@ def process_commands(args):
 
     signtransaction_parser = subparsers.add_parser('signtransaction', help='Sign a PSBT')
     signtransaction_parser.add_argument('--psbt', help='The Partially Signed Bitcoin Transaction to sign')
+    signtransaction_parser.add_argument('--testnet', help='Use testnet prefixes', action='store_true')
+    signtransaction_parser.add_argument('--fingerprint', help='Specify the device to connect to using the first 4 bytes of the hash160 of the master public key. It will connect to the first device that matches this fingerprint.')
```
achow101: Why? `--fingerprint` is a global option.

hwilib/commands.py
```diff
@@ -269,6 +269,9 @@ def process_commands(args):
       descriptor = Descriptor.parse(args.desc, args.testnet)
       args.fingerprint = descriptor.fingerprint
 
+    if hasattr(args, "psbt") and args.psbt:
+      args.fingerprint = args.fingerprint
```
achow101: ... Why? What's the point of this?

hwilib/commands.py
```diff
@@ -245,6 +252,10 @@ def process_commands(args):
     if command == 'enumerate':
         return args.func(args)
 
+    if hasattr(args, "desc") and args.desc:
```
achow101: I don't think this if-block is useful. Anything that has a `desc` option should be handling it by itself inside of it's own function.

hwilib/commands.py
Sjors: What makes something global?

hwilib/commands.py
Sjors: I don't like this either, but it was throwing errors if I didn't set fingerprint early enough.

Sjors:
Regarding global variables, I'm not sure what the right approach is.

I guess if we translate this to gRPC we could use `POST /:fingerprint/signtransaction` with `{psbt: ....}` as the payload. However some methods don't need the fingerprint argument because it's in the descriptor. Unless we just require this redundantly.

hwilib/commands.py
Sjors: Leftover from an earlier version, will remove.

instagibbs:
Can we mark this as WIP if the Core branch is unmerged?

hwilib/commands.py
Sjors: Fixed typo, but if I leave out the check the command just returns `null`:
`./hwi.py --testnet --fingerprint 00000000 displayaddress`

hwilib/commands.py
Sjors: I removed the commit that adds this redundancy.

hwilib/commands.py
Sjors: I replaced this with a consistency check if both `--desc` and `--fingerprint` is set.

Sjors:
I narrowed the scope a bit, so hopefully it doesn't have to stay in WIP limbo:
* no longer supporting redundant ability to put global params at the end
 * no longer renaming `signtransaction` to `signtx`

The only backwards incompatible change is the `psbt` is now a named argument. If necessary I can change that so the position argument is optional but still possible.

Will deal with Travis failures shortly.

hwilib/commands.py
```diff
@@ -282,9 +290,12 @@ def process_commands(args):
     getkeypool_parser.set_defaults(func=getkeypool)
 
     displayaddr_parser = subparsers.add_parser('displayaddress', help='Display an address')
-    displayaddr_parser.add_argument('path', help='The BIP 32 derivation path of the key embedded in the address')
-    displayaddr_parser.add_argument('--sh_wpkh', action='store_true', help='Display the p2sh-nested segwit address associated with this key path')
-    displayaddr_parser.add_argument('--wpkh', action='store_true', help='Display the bech32 version of the address associated with this key path')
+    group = displayaddr_parser.add_mutually_exclusive_group()
+    group.add_argument('--desc', help='Pseudo-descriptor for which to return a descriptor with keys, e.g. wpkh(00000000/84h/1h/0h/0/*). Alternatively, use --path with --sh_wpkh / --wpkh')
```
instagibbs: what is "psuedo" about the option?

Sjors: I'm calling it a pseudo-descriptor, because there's no public key in it. The goal of a pseudo-descriptor like `wpkh(00000000/84h/1h/0h/0/*)` is to ask for an actual descriptor ``wpkh(00000000/84h/1h/0h/tpub..../0/*)` with a public key inserted in the right place. That's not covered in @sipa's document: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md

instagibbs: On further thought, I think these are ok. Might want to call them something like "descriptor requests" or something, and say exactly what info must reside in the requests, and what's missing.

achow101:
I don't follow why `signtx` needs `psbt` to be a named argument.

Sjors:
@achow101 I made everything a named argument, because it's easier to translate to an RPC.

Sjors:
I'm still trying to figure out how this mapping would work. I see it's still possible to use a special purpose `value` field in the JSON payload to deal with a positional argument: https://grpc.io/blog/coreos

So I'll change `psbt` back to positional. That's not possible for `displayaddress --path / --descriptor`  though, unless we try to guess based on the format of the positional argument.

instagibbs:
just leaving this here since I went to the stale branch: https://github.com/bitcoin/bitcoin/pull/14912

Sjors:
Thanks, I closed that PR and updated the description above.

The branch wasn't stale by the way, but the PR discussion and description was out of date.

Sjors:
* renamed "pseudo descriptor" to "descriptor request" (which now requires square brackets around the origin part, so it's exactly like a descriptor with origin info, but without the public key)
* cleaned op the descriptor.py and added tests
(pushing the corresponding changes to bitcoind in a moment)

Sjors:
I extracted the descriptor class and `displayaddress` changes into a new PR #117. 

I'm no longer sure if descriptor requests are the right approach. Ultimately we're only asking the device for a key at a certain derivation path. That key is the same regardless of what descriptor it's used in. So it makes more sense for Bitcoin Core to just directly ask for the key using `getxpub`, and then it can construct the descriptor. It won't need `getkeypool` in any case.

It's still important for Bitcoin Core to know which kinds of things the device can sign, especially if we start adding more complicated stuff like multisig. Perhaps `enumerate` is the right place for that. That would still require some standard way of expressing these capabilities.

Something similar to descriptors, but perhaps with wildcards, e.g.: `["pkh("44'/1'/$'/{0,1}/*"), sh(wpkh("49'/1'/$'/{0,1}/*")), wpkh("84'/1'/$'/{0,1}/*")]` would be returned by a device that supports legacy, wrapped segwit and segwit native, when called with `--testnet`.

PR #75, opened 2018-12-13
Fix trezor implementation to work with python-trezor 0.11.0

achow101:
python-trezor 0.11.0 changed a bunch of their api stuff, so this updates the trezor implementation to work with 0.11.0. The minimum trezor library version is also set in setup.py.


achow101:
Merging this since travis built with an earlier version of HWI indicating that the implementation was fixed.

PR #76, opened 2018-12-17
Change range from array to object

achow101:
The API for descriptor ranges changed in the latest hww from an array to an object. Update this to match that change.


achow101:
Closing this, will make a larger fix things PR that includes this commit.

PR #77, opened 2018-12-17
travis: Hard reset when hww changes

achow101:
Changes to hww are going to be force pushed, so we need to hard reset instead of pull.


achow101:
Travis indicates that the branch was hard reset and the updated branch is being built, so merging.

PR #78, opened 2018-12-17
Fixes to work with updated hww and trezor library

achow101:
The hww branch changed the api for ranges. It is now an object instead of an array.

The trezor library no longer specifies a default UI that it uses for getting passwords. Specify the UI for this so that users can enter a password.

The trezor library keys prevtxs using byte objects for the hashes, not hex strings.


achow101:
Will merge once travis passes.

PR #79, opened 2018-12-19
Change tests to use unittest and some bug fixes as a result of this

achow101:
Changes the tests to use Python's unittest framework. This cuts down on duplication and allows for tests to be run in parallel and all tests to be run when one fails.

Also fixes some bugs that were uncovered in the process of doing this.


achow101:
Merging as tests pass on travis

PR #80, opened 2018-12-21
Set Python version for PyEnv

Sjors:
Matches the version in `.travis.yml` and prevents devs (who use PyEnv) from accidentally using too modern Python syntax.


achow101:
utACK a508077da17a896fffb42f9763255a87a63619cf

PR #81, opened 2018-12-26
Handle Exceptions correctly

achow101:
Handles NotImplementedErrors with proper error messages and codes instead of with the normal python tracebacks.

Also fixes and adds tests for two broken error handling return statements.


instagibbs:
utACK

PR #84, opened 2018-12-30
Implement displayaddress for Coldcard and Digital BitBox and make it return the address displayed

achow101:
Implements `displayaddress` for the Coldcard and Digital BitBox. The BitBox doesn't have a screen, so the implementation is just a message saying the `displayaddress` cannot be used with it.

Also changes displayaddress to return the address that was being displayed rather than `null`,

Builds on top of #82 as this needs the documentation update and the `UnavailableActionError` introduced there.


peter-conalgo:
Looks good. My only feedback is on use of `self.device.check_mitm()` which you're doing for the backup command (boring) and not in the show-address command (where it could matter).

achow101:
I've added `self.device.check_mitm()` to this and the other commands that don't have it.

PR #87, opened 2019-01-02
Implement signmessage for Coldcard to complete its implementation

achow101:
Implements the `signmessage` command for the Coldcard. This completes the Coldcard implementation and the documentation is updated as such.

Requires #84 and #82 which add the other commands for the Coldcard.


achow101:
Right now user aborts (and other device errors) aren't handled super well. I think I'll handle that issue en masse with everything else that can abort sometime later.

Fixed the other two issues.

PR #88, opened 2019-01-02
add 'watchonly' field and set to true + docs

fivepiece:
Adds the "watchonly" field to `importmulti` output so the warning doesn't show (we expect to always import xpubs, correct?)

```
[
  {
    "success": true,
    "warnings": [
      "Some private keys are missing, outputs will be considered watchonly. If this is intentional, specify the watchonly flag."
    ]
  }
]
```

Also fixup the `"range"` notation in the docs because using `"range": [0, 1000]` is incorrect and causes an error :

```
[
  {
    "success": false,
    "error": {
      "code": -8,
      "message": "End of range for descriptor must be specified"
    }
  }
]
```

### DIFF
--- a/docs/bitcoin-core-usage.md
+++ b/docs/bitcoin-core-usage.md
@@ -49,7 +49,7 @@ We will be fetching keys at the BIP 84 default.
 
 ```
 $ ./hwi.py -f 8038ecd9 getkeypool --wpkh --keypool 0 1000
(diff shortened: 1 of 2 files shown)


instagibbs:
utACK https://github.com/achow101/HWI/pull/88/commits/e02e70416bf38c6af63dab03d5c87102d22b8d91

achow101:
utACK e02e70416bf38c6af63dab03d5c87102d22b8d91

PR #90, opened 2019-01-05
Implement signmessage, displayaddress, and signtx for keepkey and mark its implementation as complete

achow101:
Implements the signmessage, displayaddress, and signtx commands for the keepkey. This also completes its implementation and the documentation is updated as such.

Note that because the Keepkey is a clone of the Trezor, these implementations are copies of the Trezor ones modified slightly to work with the Keepkey library.


achow101:
Added some notes in the documentation about firmware transaction signing limitations. These are the same as for the Trezor.

achow101:
Closing as #102 is built on top of this but review for it is easier because it has tests too.

Issue #91, opened 2019-01-06
TODO for 1.0 release


achow101:
At some point I would like to start doing proper versioning of HWI and release it as both a library and executable script on PyPi. The following need to be done before I feel comfortable doing this:

- [x] Separate `process_commands` from `commands.py` and have the functions in `commands.py` take actual arguments
- [x] Handle all possible errors so that no backtraces are printed
- [x] Remove any interactiivty and command line prompts (except for `--debug`). The only thing that should be output is the result. (This may not be possible for Trezor and Keepkey, see #83)
- [x] Have all devices have implementations for all commands. No `NotImplementedError`s
- [x] Tests for all commands, especially transaction signing (incl. multisig)
- [x] Manual test for ledger
- [x] ~~Manual~~ Automated test for bitbox
- [x] Automated test for keepkey
- [x] Move to bitcoin-core org
- [x] Deterministic binary and archive builds
- [x] `--version` option
- [x] Whatever else is listed in the 1.0 milestone

instagibbs:
Just food for thought: It might also make sense to move this repo into some organization, and establish contributor guidelines/review process going forward.

achow101:
Since the main goal of this project is to get hardware wallet support into Core, we could see if this can be moved under the bitcoin-core org.

instagibbs:
https://github.com/bitcoin-core/HWI/pull/121 resolves "* Deterministic binary and archive builds" I presume

instagibbs:
I think "* Tests for all commands, especially transaction signing (incl. multisig)" has been covered. Multisig tests have been merged.

instagibbs:
is there a tracking issue for `--version`?

Sjors:
I'd also like #117 in that release as it would allow for library based testing of https://github.com/bitcoin/bitcoin/pull/14912. Although probably I'll need message signing commands too, to double check the import.

instagibbs:
With the -cli limit on transaction sizes(noted in https://github.com/bitcoin-core/HWI/pull/125) I think we should add https://github.com/bitcoin-core/HWI/issues/46 to the list so a -cli user can side-step this limitation.

achow101:
Everything tagged has been merged. I'll tag and publish the release tomorrow.

PR #93, opened 2019-01-06
Refactor to actually be useful as a library

achow101:
Refactors `command.py` so that it just has the functions for each command and the arguments are the actual arguments for the command instead of `args` and `client`. `process_commands` has been moved to `cli.py` along with the handlers that take `args` and `client` which then call the functions in `commands.py` to do the command.

`setup.py` has been updated to also install a `hwi` executable script and the version number is bumped to 0.0.5.

This makes HWI actually usable as a library while also having the command line functionality.

Completes part 1 of #91

### DIFF
--- a/hwi.py
+++ b/hwi.py
@@ -2,11 +2,6 @@
 
 # Hardware wallet interaction script
 
-from hwilib.commands import process_commands
+from hwilib.cli import main
 
-import sys
-import json
-
-if __name__ == '__main__':
-    result = process_commands(sys.argv[1:])
-    print(json.dumps(result))
+main()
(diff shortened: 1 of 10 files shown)


achow101:
Had to rebase

achow101:
Fixed the error

Sjors:
Concept ACK. But let's use named arguments, especially for `getkeypool(client, args.path, args.start, args.end, args.internal, args.keypool, args.account, args.sh_wpkh, args.wpkh)`. That makes it easier for me in #73 to pass in a descriptor, without having to explicitly set `args.path`, `args.sh_wpkh`, etc to null.

Sjors:
The latter. It's probably fine to keep some arguments like `client` positional.

achow101:
@Sjors Done.

achow101:
Seems like a spurious failure. I restarted the travis build and it seems fine now

Sjors:
tACK 8e4b3fc (based on testing #73, which I just rebased)

PR #94, opened 2019-01-07
Send the full amount of funds in TestSignTx to test all input types

instagibbs:
Also beef up the signing test a bit to try combinations of 1 to 3 inputs to catch corner cases in signing logic.

I am writing a Ledger test and this case was not failing previously because it can handle legacy or witness, but not both.


achow101:
The first commit can be dropped as it isn't really doing anything.

instagibbs:
squashed

achow101:
ACK 106c23128ba6e4ed98a71177a11522ae16cc9550

PR #96, opened 2019-01-08
Bitbox test

achow101:
Like the ledger test, but for bitbox.

Also lets you specify the password for the device. This is currently only used for the bitbox.


achow101:
Rebased onto master and enabled multisig and mixed input tests for digitalbitbox

instagibbs:
utACK though I cannot test https://github.com/achow101/HWI/pull/96/commits/c29a135e97f8a3526ceb8432154840a846979d3a

PR #99, opened 2019-01-09
Test signtx with unknown inputs by wiping keypath and processing

instagibbs:
Probably should start thinking about breaking out tests a bit more soon.


instagibbs:
Coldcard test not liking it...

instagibbs:
rebased, refactored, and I think fixed the issue with Coldcard(I think I wasn't sweeping the wallet between runs, leading to UTXO buildup)

instagibbs:
I was mistaken:
Trezor is complaining about the multisig script not having its own pubkey there, and Coldcard complaining it doesn't have the keypath.

instagibbs:
Not exactly sure what's best to do here. Make the most generic test that Coldcard can pass?

instagibbs:
@peter-conalgo was told you would know more. What's considered an appropriate PSBT input by the ColdCard?

peter-conalgo:
The Coldcard won't sign any input for which we don't have the corresponding keypath. Are you expecting us to sign it, or just do nothing and pass through an input?

instagibbs:
Was expecting the latter. "Sign what you can" is the nicest to work with, unless there's a security reason not to do it?

instagibbs:
In other words, what's the trigger for the keypath check? The PSBT Input existing? (I'd have to look at BIP174 again to see what's legal)

peter-conalgo:
Without a keypath, there is no way to know what derivation to use. We try not to make assumptions about the HD key paths that wallets use, so it's not possible to go from pubkey in unsigned txn to 'yes could I sign this'.

If at least one input was sign-able by us, then I suppose we shouldn't treat lack of information on  other inputs as an error.

But I'm unclear what you're trying to do. Please describe the PSBT you are giving it.

instagibbs:
@peter-conalgo sorry if the PR wasn't explicit enough let me explain:

We want a test suite to make sure each device, if it can, signs inputs it can, and skips over ones it cannot.

The quickest way I could think to support this in a test was to give the keypath for specific inputs a lobotomy. Ledger driver sees no keypath, and simply doesn't try to sign for that input. With Coldcard, you're handing an entire PSBT, and it throws an error rather than sign the rest of the inputs.

Rather than guess and check on my end, what kind of input is the Coldcard expecting when it wants to pass over an input? That the master key fingerprint differs?

peter-conalgo:
Yes. Just give us a garbage master key fingerprint. However, I think you might get (a different) error if you give us a PSBT where we have nothing to do. Probably best to give us one input to sign, and one to ignore.

instagibbs:
@peter-conalgo yep you're right: `{'error': 'Remote Error: None of the keys involved in this transaction belong to this Coldcard (expect 0x4369050f).', 'code': -7}`.

instagibbs:
Fixed and squashed.

test/test_device.py
```diff
@@ -202,6 +203,56 @@ def setUp(self):
...
+            first_psbt.deserialize(psbt['psbt'])
+            second_psbt = PSBT()
+            second_psbt.deserialize(psbt['psbt'])
+
+
+            # Blank master fingerprint to make hww fail to sign
+            # Single input PSBTs will be fully signed by first signer
+            for psbt_input in first_psbt.inputs[1:]:
+                for pubkey, path in psbt_input.hd_keypaths.items():
+                    psbt_input.hd_keypaths[pubkey] = (0,0,0,0)+path[4:]
```
achow101: The master fingerprint is not a tuple of bytes. It's an integer that is in `path[0]`. So this is wrong and not the correct way to blank out the master fingerprint. It should really be `psbt_input.hd_keypaths[pubkey] = [0]+path[1:]`

instagibbs: the keypath itself is a tuple of integers, so it's actually:

`psbt_input.hd_keypaths[pubkey] = (0,)+path[1:]`

to allow concatenation.

instagibbs: fixed

PR #102, opened 2019-01-11
Finalize implementation of keepkey and add automated tests for it

achow101:
Implements the signmessage, displayaddress, and signtx commands for the keepkey. This also completes its implementation and the documentation is updated as such.

Note that because the Keepkey is a clone of the Trezor, these implementations are copies of the Trezor ones modified slightly to work with the Keepkey library.

This also adds an automated test for the KeepKey using the KeepKey simulator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ addons:
             - cython3
             - ccache
 install:
-    - pip install pipenv pysdl2 python-bitcoinrpc
+    - pip install pipenv pysdl2 python-bitcoinrpc protobuf
     # From trezor-mcu to get the correct protobuf version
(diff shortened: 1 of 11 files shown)


instagibbs:
is this WIP?

achow101:
> is this WIP?

Kind of. It's waiting on upstream changes. However I have finished everything else.

achow101:
Upstream has fixed their issues (besides their versioning and PyPi problems) so this is ready for review now. Instead of splitting implementation and tests, I have decided to just make them one large pr as having tests make it easier to check that the implementation is correct.

keepkeyjon:
I filed a few issues for things on our end that could be better:
https://github.com/keepkey/keepkey-firmware/issues/78
https://github.com/keepkey/python-keepkey/issues/40
https://github.com/keepkey/python-keepkey/issues/39

instagibbs:
utACK

PR #103, opened 2019-01-11
Ensure that a bogus signature will always be created for ignored inputs on Trezor

### LINKED ISSUE #100: Trezor cannot partially sign a transaction with a multisig input it cant sign
`{'error': 'DataError: signing.c:849:Pubkey not found in multisig script', 'code': -7}` even when you don't want to sign the input.

achow101:
Fixes the issue where multisig inputs not belonging to the trezorwould cause an error. Also re-enables that test for the trezor

Fixes #100

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -131,6 +131,8 @@ class TrezorClient(HardwareWalletClient):
 
                 def ignore_input():
                     txinputtype.address_n = [0x80000000]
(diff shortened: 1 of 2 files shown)


instagibbs:
utACK https://github.com/achow101/HWI/pull/103/commits/c54bc333810de679a1a515030978c0bb3cd223b3

Nice to see that test case reintroduced

PR #104, opened 2019-01-12
Update digital bitbox encryption scheme for firmware v5 and bug fixes

achow101:
Firmvware v5.0.0 changed how messages are encrypted and authenticated, so update the encryption methodolgy to handle that.

The changes come from https://github.com/digitalbitbox/mcu/blob/master/py/dbb_utils.py with some minor changes to accommodate older firmware versions.

A bug in DER serialization which is only used by the bitbox is also fixed here.


achow101:
I've added a fix for a bug in the the DER serialization that I found.

PR #106, opened 2019-01-14
More verbose test output indicating which tests for which devices pass tests

achow101:
Increases the verbosity level of the test runner to 2 which outputs the `str()` representation of each test case. This includes the test name by default. This has been modified to also print out the device types for the tests in `test_devices.py`. This will allow us to be able to see what tests each device is passing so debugging is easier.

### DIFF
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -54,5 +54,5 @@ if args.digitalbitbox:
         print('Cannot run Digital Bitbox test without --password set')
 if not args.no_keepkey:
     suite.addTest(keepkey_test_suite(args.keepkey, rpc, userpass))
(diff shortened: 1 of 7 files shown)


instagibbs:
just for completeness could you over-load the device-specific tests too such as `KeepkeyTestCase`, `TrezorTestCase`?

achow101:
Overloaded those too.

instagibbs:
new pushes are breaking something, logs don't make it clear

instagibbs:
utACK https://github.com/achow101/HWI/pull/106/commits/b404a5050d533877eda66be51c63caef5cbcedff

PR #107, opened 2019-01-17
Ledger firmware doesn't support mixed input types during signing pass

instagibbs:
You can theoretically sign but requires that all inputs are given as non-segwit UTXO, so not compliant with BIP174.


achow101:
ACK 8b9d0685b7f7e1795c20f50d9815ff7cc87bdab9

PR #113, opened 2019-01-25
Use updated upstream things

achow101:
Keepkey has updated their PyPi to have their latest library version, so switch things to using that instead of using a specific commit from their repo.

The Digital Bitbox simulator was merged into the upstream repo so use that instead of my own branch for it.

Also added documentation for building the Keepkey emulator.


achow101:
Merging as automated tests pass.

PR #114, opened 2019-01-30
Change links to bitcoin-core

achow101:
The repo migrated to bitcoin-core, so changing things that referred to the old location to point to the new one.


instagibbs:
ACK

PR #115, opened 2019-01-31
Always close clients if they exist

achow101:


instagibbs:
tests fail for some poorly defined reason

achow101:
I fixed that error.

@instagibbs Is this what you had in mind for "I suggest we sprinkle `client.close()` when `not None` at the end of each `enumerate` function." (from https://github.com/bitcoin-core/HWI/pull/109#issuecomment-459101885)

instagibbs:
yep, LGTM

PR #116, opened 2019-02-01
Handle all exceptions so that none are shown to the user

achow101:
This PR adds `except` statements around the main command function to catch any and all exceptions and output the correct error code and message. Additional exceptions and error codes are added for new types of errors. Any exception thrown by a library will be caught and converted into the correct type of exception so that the correct message and code can be returned. This change makes it so that users using the command line tool should never see a backtrace.

This removes from the main `commands.py` file handling of exceptions. Instead those functions (which are intended for use as a library in other software) will throw the exceptions and the caller can handle them.


instagibbs:
utACK, seems to work with Ledger.

PR #117, opened 2019-02-02
Descriptor support for displayaddress

Sjors:
Adds a `Descriptor` class, which currently handles some of the descriptor functionality.

Changes `displayaddress` to allow a descriptor argument. This means `--path` and `--desc` are now named arguments.

When using a descriptor  `displayaddress` performs additional checks to make sure the fingerprint and xpub match. To enable the fingerprint check, I'm storing the fingerprint on `HardwareWalletClient` instances.

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -22,7 +22,7 @@ def backup_device_handler(args, client):
     return backup_device(client, label=args.label, backup_passphrase=args.backup_passphrase)
 
 def displayaddress_handler(args, client):
-    return displayaddress(client, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh)
+    return displayaddress(client, desc=args.desc, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh)
 
 def enumerate_handler(args):
     return enumerate(password=args.password)
@@ -102,7 +102,9 @@ def process_commands(args):
     getkeypool_parser.set_defaults(func=getkeypool_handler)
 
     displayaddr_parser = subparsers.add_parser('displayaddress', help='Display an address')
-    displayaddr_parser.add_argument('path', help='The BIP 32 derivation path of the key embedded in the address')
+    group = displayaddr_parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--desc', help='Output Descriptor. E.g. wpkh([00000000/84h/0h/0h]xpub.../0/0), where 00000000 must match --fingerprint and xpub can be obtained with getxpub. See doc/descriptors.md in Bitcoin Core')
+    group.add_argument('--path', help='The BIP 32 derivation path of the key embedded in the address, default follows BIP43 convention, e.g. m/84h/0h/0h/1/*')
     displayaddr_parser.add_argument('--sh_wpkh', action='store_true', help='Display the p2sh-nested segwit address associated with this key path')
     displayaddr_parser.add_argument('--wpkh', action='store_true', help='Display the bech32 version of the address associated with this key path')
     displayaddr_parser.set_defaults(func=displayaddress_handler)
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -6,9 +6,10 @@ import glob
 import importlib
 
 from .serializations import PSBT, Base64ToHex, HexToBase64, hash160
-from .base58 import get_xpub_fingerprint_as_id, get_xpub_fingerprint_hex
+from .base58 import get_xpub_fingerprint_as_id, get_xpub_fingerprint_hex, xpub_to_pub_hex
 from os.path import dirname, basename, isfile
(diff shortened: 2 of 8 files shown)


hwilib/cli.py
```diff
@@ -95,7 +95,9 @@ def process_commands(args):
     getkeypool_parser.set_defaults(func=getkeypool_handler)
 
     displayaddr_parser = subparsers.add_parser('displayaddress', help='Display an address')
-    displayaddr_parser.add_argument('path', help='The BIP 32 derivation path of the key embedded in the address')
+    group = displayaddr_parser.add_mutually_exclusive_group()
```
achow101: Can you make this required? E.g. `group = displayaddr_parser.add_mutually_exclusive_group(required=True)`

achow101:
Needs rebase

Sjors:
Rebased

test/test_device.py
```diff
@@ -380,15 +380,16 @@ def tearDown(self):
...
         self.assertEqual(result['code'], -7)
 
     def test_display_address(self):
-        process_commands(self.dev_args + ['displayaddress', 'm/44h/1h/0h/0/0'])
-        process_commands(self.dev_args + ['displayaddress', '--sh_wpkh', 'm/49h/1h/0h/0/0'])
-        process_commands(self.dev_args + ['displayaddress', '--wpkh', 'm/84h/1h/0h/0/0'])
+        process_commands(self.dev_args + ['displayaddress', '--path', 'm/44h/1h/0h/0/0'])
+        process_commands(self.dev_args + ['displayaddress', '--sh_wpkh', '--path', 'm/49h/1h/0h/0/0'])
+        process_commands(self.dev_args + ['displayaddress', '--wpkh', '--path', 'm/84h/1h/0h/0/0'])
+        process_commands(self.dev_args + ['displayaddress', '--desc', 'wpkh(m/84h/1h/0h/0/0)'])
```
achow101: I don't think that's a valid descriptor.

hwilib/descriptor.py
```diff
@@ -0,0 +1,76 @@
...
+        if base_key_and_path_match:
+            base_key = base_key_and_path_match.group(1)
+            path_suffix = base_key_and_path_match.group(2)
+            if path_suffix == ")":
+                path_suffix = None
+        else:
+            if origin_match == None:
+                return None
+
+        return cls(origin_fingerprint, origin_path, base_key, path_suffix, testnet, sh_wpkh, wpkh)
```
achow101: I got an error which points here: `UnboundLocalError: local variable 'path_suffix' referenced before assignment`

`path_suffix` won't be set if the else branch above is taken.

Sjors:
Rebased. I added a check to prevent `--sh_wpkh` and `--wpkh` from being used in combination with `--desc`. Fixed `path_suffix`. Added a bunch more tests.

hwilib/commands.py
```diff
@@ -154,10 +156,25 @@ def getkeypool(client, path, start, end, internal=False, keypool=False, account=
...
+        if descriptor is None:
+            return {'error':'Unable to parse descriptor: ' + desc,'code':BAD_ARGUMENT}
+        if descriptor.m_path is None:
+            return {'error':'Descriptor missing origin info: ' + desc,'code':BAD_ARGUMENT}
+        if descriptor.origin_fingerprint != client.fingerprint:
+            return {'error':'Descriptor fingerprint does not match device: ' + desc,'code':BAD_ARGUMENT}
+        xpub = client.get_pubkey_at_path(descriptor.m_path_base)['xpub']
+        if descriptor.base_key != xpub and descriptor.base_key != xpub_to_pub_hex(xpub):
+            return {'error':'Key in descriptor does not match device: ' + desc,'code':BAD_ARGUMENT}
+        client.display_address(descriptor.m_path, descriptor.sh_wpkh, descriptor.wpkh)
```
achow101: This needs to be a `return` statement.

Sjors: I think we need to add a linter to Travis :-)

hwilib/commands.py
```diff
@@ -154,10 +156,25 @@ def getkeypool(client, path, start, end, internal=False, keypool=False, account=
...
 
-def displayaddress(client, path, sh_wpkh=False, wpkh=False):
-    if sh_wpkh == True and wpkh == True:
-        return {'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT}
-    return client.display_address(path, sh_wpkh, wpkh)
+def displayaddress(client, path=None, desc=None, sh_wpkh=False, wpkh=False):
+    if path is not None:
+        if sh_wpkh == True and wpkh == True:
+            return {'error':'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.','code':BAD_ARGUMENT}
+        client.display_address(path, sh_wpkh, wpkh)
```
achow101: `return` here too

Sjors: Fixed

achow101:
ACK a71658cf6db6b59dac88366ab0c2a7f4c90d2b1b

PR #118, opened 2019-02-04
Add more tests and related fixes

achow101:
Adds tests for nearly all commands for all devices and related fixes. Only `restore` for `trezor` and `keepkey` are untested as I could not get them to work. All other commands for all devices have tests and related bug fixes.

### DIFF
--- a/hwilib/base58.py
+++ b/hwilib/base58.py
@@ -87,18 +87,6 @@ def to_address(b, version):
     data += checksum
     return encode(data)
 
-def pubkey_to_address(pubkey, testnet=False):
-    pkh = hash160(pubkey)
-    if testnet:
-        return to_address(pkh, b'\x6f')
-    else:
-        return to_address(pkh, b'\x00')
-
-def xpub_to_address(xpub, testnet=False):
-    data = decode(xpub)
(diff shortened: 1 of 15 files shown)


hwilib/base58.py
```diff
@@ -87,23 +87,6 @@ def to_address(b, version):
...
-        return to_address(pkh, b'\x6f')
-    else:
-        return to_address(pkh, b'\x00')
-
-def xpub_to_address(xpub, testnet=False):
-    data = decode(xpub)
-    pubkey = data[-37:-4]
-    return pubkey_to_address(pubkey, testnet)
-
-def xpub_to_pub_hex(xpub):
```
Sjors: I still need `xpub_to_pub_hex` in #117, because the inferred descriptors from Core use the pubkey in hex.

Sjors:
I'm all for more tests. Light review of d097646 only found one problem, but I'm not very familiar with the test suite.

hwilib/base58.py
achow101: I've added it back in

achow101:
Merging as tests pass

PR #120, opened 2019-02-10
Replace device libraries with stripped down versions

achow101:
This PR replaces all of the device libraries (`trezor`, `keepkey`, `ckcc-protocol,` and `btchip-python`) with stripped down versions of them directly in the source tree. These stripped down versions have several unused functions removed, including support for outdated hardware (e.g. Ledger HW.1) and altcoins. Some additional modifications have been made to them in order to make our implementations cleaner and easier to understand.

One major part of this change is replacing the Keepkey implementation with the Trezor implementation entirely. I found during testing that the Keepkey and Trezor use largely the same protocol. Some parts were different but those did not affect our usage of the Keepkey and the Trezor, so those differences were removed and reconciled. This allows us to have only one library which serves both the Trezor and Keepkey and have just one implementation that works for both Trezor and Keepkey.

Overall, by replacing the vendor provided libraries with our own versions of them, we are able to reduce the total number of dependencies required, and reduce some possible implementation errors (e.g. it would be a simple mistake to let Trezor or Keepkey do their default behavior of phoning home to fetch previous transactions). We can also make changes to the lower level implementations as needed instead of waiting on changes and fixes to be merged into upstream libraries.

This PR is based on #118 so that everything that was changed could be tested.


Sjors:
Concept ACK. I am a bit concerned about long term maintainability. I'd have to review the code get a better sense of how many moving parts we're dealing with. Ideally though we'd upstream these lighter libraries back to the manufactures (and then pin them by hash, like Bitcoin Core `/depends`). 

I'll review Ledger and Trezor stuff some time after Bitcoin Core 0.18 feature freeze.

achow101:
> I am a bit concerned about long term maintainability

I don't really think that will be a problem. The changes that we are concerned about are limited to a specific set of files. After looking through the commit history of python-trezor and python-keepkey, it seems like many of the changes happening to these libraries have been for supporting altcoins or changes to their tests, neither of which are things we need here. And `btchip-python` and `ckcc-protocol` haven't really been changed in several months, so I don't expect those to change very much or frequently. Any changes that are made are things we should be able to easily pull in here.

> Ideally though we'd upstream these lighter libraries back to the manufactures

I don't think that is possible. The manufacturers have obligations to support older hardware, altcoins, and other functionality that we don't use or care about. These lighter libraries are just versions of the original ones with all of that stuff deleted.

instagibbs:
concept ACK. Updating these stripped down libraries should be pretty easy even in the case of them being updated upstream. Some tests are failing.

achow101:
Test failure caused by bad hww branch. Should be fixed now.

instagibbs:
With https://github.com/bitcoin-core/HWI/pull/122 Ledger tests still pass for me, without the ledger python library installed.

instagibbs:
light tACK https://github.com/bitcoin-core/HWI/pull/120/commits/681a52d3da474cbdb2974c8dfc559214bdf748dc for ledger

PR #124, opened 2019-02-15
Add a switch for interactivity and make setup and restore interactive only commands

achow101:
Currently using setup and restore on the Trezor and Keepkey (which are basically the only devices that support setup and restore) are broken. They require user interaction. Instead of removing them entirely, move those commands behind a switch, `--interactive`.

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -9,7 +9,8 @@ from .errors import (
     DEVICE_CONN_ERROR,
     NO_PASSWORD,
     UNKNWON_DEVICE_TYPE,
-    UNKNOWN_ERROR
+    UNKNOWN_ERROR,
+    UNAVAILABLE_ACTION
 )
 from . import __version__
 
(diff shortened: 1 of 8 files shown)


Sjors:
Maybe make interactive the default?

achow101:
> Maybe make interactive the default?

No.

instagibbs:
why not?

achow101:
Keeping commands non-interactive means that callers of this will know exactly what to expect out of each and every command. The only reason that an interactive mode is being added is to facilitate functions that for some devices must be interactive (i.e. Trezor and Keepkey). Additionally the interaction for those commands will differ between devices (e.g. Trezor requires interaction for `setup` and `restore` but the Digital Bitbox doesn't). The interactive flag is really more like an indicator that the interface for those commands will vary. The difference in what is required for the interactivity of those commands is also not easily handled by the software calling those commands and requires specific knowledge of what to provide for each device.

instagibbs:
needs rebase

achow101:
Rebased

hwilib/cli.py
```diff
@@ -69,6 +74,7 @@ def process_commands(cli_args):
     parser.add_argument('--fingerprint', '-f', help='Specify the device to connect to using the first 4 bytes of the hash160 of the master public key. It will connect to the first device that matches this fingerprint.')
     parser.add_argument('--version', action='version', version='%(prog)s {}'.format(__version__))
     parser.add_argument('--stdin', help='Enter commands and arguments via stdin', action='store_true')
+    parser.add_argument('--interactive', '-i', help='Use some commands interactively', action='store_true')
```
instagibbs: maybe add "Required for some commands"

instagibbs: or "Currently required for all device configuration commands"

achow101: Done

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/124/commits/4453555c21b0a118c6f509964ef0e4710293456a

PR #126, opened 2019-02-15
Add option to enter commands over stdin

achow101:
This PR adds a `--stdin` option which allows arguments to be entered over stdin in a way similar to bitcoin-cli's `--stdin` option.

Built on #125 and to test the `stdin` interface

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,8 @@ jobs:
             script: cd test; poetry run ./run_tests.py --interface=library
         -   name: With command line interface
             script: cd test; poetry run ./run_tests.py --interface=cli
+        -   name: With stdin interface
+            script: cd test; ./run_tests.py --interface=stdin
         -   name: With linux binary distribution command line interface
             services: docker
             before_script:
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -58,7 +58,7 @@ def prompt_pin_handler(args, client):
 def send_pin_handler(args, client):
     return send_pin(client, pin=args.pin)
 
-def process_commands(args):
+def process_commands(cli_args):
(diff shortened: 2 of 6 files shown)


Sjors:
Can you give some usage examples? Do you have to enter all arguments this way, or can do something like: `cat psbt > hwi.py --fingerprint=... signtransaction`?

achow101:
You can do it mixed. You just need to have the arguments entered in the order as if they were entered on the command line.

instagibbs:
needs rebase

achow101:
Rebased

instagibbs:
needs rebase again

achow101:
Rebased

instagibbs:
So this is only supposed to work on commands and the comman's arguments, right? Not path, type, etc.

instagibbs:
light tACK https://github.com/bitcoin-core/HWI/pull/126/commits/49bc7fa5dab02668c8a8b33cc4b6932acf688110

achow101:
> So this is only supposed to work on commands and the comman's arguments, right? Not path, type, etc.

It should work with all arguments, including path, type, etc.

instagibbs:
ah nevermind, was entering the other options in incorrectly.

re-tACK https://github.com/bitcoin-core/HWI/commit/49bc7fa5dab02668c8a8b33cc4b6932acf688110

Sjors:
Post merge ACK. We should document how to use this...

E.g. this works:
```sh
echo "signtx PSBT" | ./hwi.py --fingerprint 00000000 --stdin
```

But this throws an error:
```sh
echo "PSBT" | ./hwi.py --fingerprint 00000000 --stdin signtx
```

I updated https://github.com/bitcoin/bitcoin/pull/14912 to use `--stdin` for the PSBT, since that could be quite long.

instagibbs:
@Sjors what error are you getting?

Sjors:
Some argument parse error. But I just tried again and now it's happy, so maybe I did something else wrong.

instagibbs:
Seems like a common thread here :)

PR #128, opened 2019-02-15
Add {read, write}psbt helper functions to handle binary PSBT files

instagibbs:
This would help with typical workflows from things like Glenn's PSBT wizard for Core where the primary method of passing the information around is via binary file.

I can add basic tests if concept ACK'd.

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -2,7 +2,7 @@
 
 from .commands import backup_device, displayaddress, enumerate, find_device, \
     get_client, getmasterxpub, getxpub, getkeypool, prompt_pin, restore_device, send_pin, setup_device, \
-    signmessage, signtx, wipe_device
+    signmessage, signtx, wipe_device, read_psbt, write_psbt
 from .errors import (
     HWWError,
     NO_DEVICE_PATH,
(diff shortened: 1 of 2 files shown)


achow101:
I think it would make more sense to just have signtx be able to take a path to a binary psbt instead of having converter functions.

instagibbs:
Maybe. I tried writing what you suggested and didn't like the UX complexity
blowup. I can try it again to see if I find a nicer way.

achow101:
It should be possible to make an argparse mutually exclusive group which has an optional argument and positional argument be mutually exclusive. So it should be possible to require either `--psbt-file FILE` or the current `psbt` as arguments to signtx.

instagibbs:
We may also want to support the case of going base64 to binary file and
vice versa.

Sjors:
Being able to pass (binary, base64 and JSON) PSBT through stdin and stdout seems useful in general. Reading from a file: not so sure, since the user can pipe it into stdin themselves, but I'm not against make things easier.

instagibbs:
@Sjors right, if the user just had to supply a `binary` flag for both input and output separately, I think stdin plus this is sufficient. Depends on how flexible stdin option is.

instagibbs:
closing in favor of just using common utilities and documenting them: https://github.com/bitcoin-core/HWI/pull/142

PR #130, opened 2019-02-17
Implement descriptor checksums

### LINKED ISSUE #129: Descriptor checksums
https://github.com/bitcoin/bitcoin/pull/15368 adds checksums for descriptors and requires them for using `importmulti`. We need to make sure descriptors that we generate have checksums. It would also be nice for us to check that descriptors being input have valid checksums.

achow101:
Adds checksums to the Descriptor class. Checksums are optional, but if provided, must be valid. `getkeypool` is also changed to use the descriptor class to return checksums. Tests are updated so that all descriptors used have checksums.

Closes #129

### DIFF
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -1,5 +1,54 @@
 import re
 
+# From: https://github.com/bitcoin/bitcoin/blob/master/src/script/descriptor.cpp
+
+def PolyMod(c, val):
+    c0 = c >> 35
+    c = ((c & 0x7ffffffff) << 5) ^ val
+    if (c0 & 1):
+        c ^= 0xf5dee51989
+    if (c0 & 2):
+        c ^= 0xa9fdca3312
+    if (c0 & 4):
+        c ^= 0x1bab10e32d
+    if (c0 & 8):
+        c ^= 0x3706b1677a
+    if (c0 & 16):
+        c ^= 0x644d626ffd
+    return c
+
+def DescriptorChecksum(desc):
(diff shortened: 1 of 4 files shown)


Sjors:
Concept ACK.

`displayaddress` also needs to support (optional?) checksums, since Bitcoin Core will automatically add them

hwilib/descriptor.py
```diff
@@ -1,5 +1,54 @@
 import re
 
+# From: https://github.com/bitcoin/bitcoin/blob/master/src/script/descriptor.cpp
```
instagibbs: why not just copy `test/functional/test_framework/descriptors.py` which I think has more review than transliteration of the C++?

instagibbs:
@Sjors we should make sure we have a test covering that case as well

hwilib/descriptor.py
achow101: Because I didn't know that file existed. I don't think it really matters either way. At least the C++ has code comments explaining how it works.

instagibbs: meh ok as long as we're generally relying on the C++ code for generation/validation we should be fine.

achow101:
> `displayaddress` also needs to support (optional?) checksums, since Bitcoin Core will automatically add them

This change does that. Checkums are currently optional for `displayaddress`.

Sjors:
For anyone else having difficulty fetching this branch after the repo move, this works too:

```sh
wget https://patch-diff.githubusercontent.com/raw/bitcoin-core/HWI/pull/130.patch && git apply 130.patch
```

hwilib/descriptor.py
instagibbs: Worst case let's just swap it out in a future PR.

instagibbs:
tACK https://github.com/bitcoin-core/HWI/pull/130/commits/2d622046ceff9752d58faeb4d1ab8a2273b6b802

with the same dissent as @Sjors . We can swap it out next.

PR #139, opened 2019-03-08
Add --version

achow101:
Adds version info and a `--version` option.


Sjors:
ACK, but consider also showing version above the usage instructions.

achow101:
> consider also showing version above the usage instructions.

Done

PR #142, opened 2019-03-15
Document how to handle binary format files from command line with bas…

instagibbs:
…h example

Since these are quite standard utilities, we may want to take the unix philosophy and let other utils take care of these common conversions.


achow101:
ACK f096c184785114658c7e18dc40dba697bff798e7

PR #147, opened 2019-04-15
Add udev rules for all supported devices

achow101:
It's useful to have all of the udev rules needed for all devices in one location instead of having to go to every vendor's webpage to find these.

Original sources:
`20-hw1.rules` (Ledger): https://github.com/LedgerHQ/udev-rules/blob/master/20-hw1.rules
`51-coinkite.rules` (Coldcard): https://github.com/Coldcard/ckcc-protocol/blob/master/51-coinkite.rules
`51-hid-digitalbitbox.rules`, `52-hid-digitalbitbox.rules` (Digital Bitbox): https://shiftcrypto.ch/start_linux
`51-trezor.rules` (Trezor): https://github.com/trezor/trezor-common/blob/master/udev/51-trezor.rules
`51-usb-keepkey.rules` (Keepkey): https://github.com/keepkey/udev-rules/blob/master/51-usb-keepkey.rules


nopara73:
Concept ACK. Coldcard tested. Stealing these btw.

instagibbs:
very nice, utACK https://github.com/bitcoin-core/HWI/pull/147/commits/1222d9a4922e496ca4334e70a92b71ffa1b8c38c

nopara73:
Can you add a CLI option that must be called with sudo? `sudo hwi installudevs`

nopara73:
Also a detect missing udevs CLI option would be also handy. I think that doesn't need sudo, right? Anyone can read the udev rules. (At least when I tested it didn't need it.) Here's a working C# code you may want to use as reference.

```cs
	public static class UdevRules
	{
		public static readonly string[] RuleFileNames =
		{
			"20-hw1.rules",
			"51-coinkite.rules",
			"51-hid-digitalbitbox.rules",
			"51-trezor.rules",
			"51-usb-keepkey.rules",
			"52-hid-digitalbitbox.rules"
		};

		public const string UdevRuleFolderPath = "/etc/udev/rules.d/";

		public static IEnumerable<string> FindMissingUdevs()
		{
			if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
			{
				yield break; // If no Linux then no udevs are missing.
			}

			var dir = new DirectoryInfo(UdevRuleFolderPath);
			var existingRules = new List<string>();
			foreach (var rule in dir.EnumerateFiles())
			{
				existingRules.Add(rule.Name);
			}

			foreach (var rule in RuleFileNames)
			{
				if (existingRules.All(x => x != rule))
				{
					yield return rule;
				}
			}
		}
	}
```

achow101:
> Can you add a CLI option that must be called with sudo? `sudo hwi installudevs`

I think that's something that we can work on in another PR.

Sjors:
tACK for Trezor T and Ledger Nano S on Ubuntu 18.04 using the v0.1.0 binary.

`enumerate` returns an empty array when using source & Poetry, but I think that's unrelated.

Instruction should also mention you need to add the user to the user to the `plugdev` group.

It like @nopara73's detect missing udevs CLI suggestion, but indeed can wait.

achow101:
> Instruction should also mention you need to add the user to the user to the `plugdev` group.

Added that to the docs.

PR #152, opened 2019-04-18
Keep Trezor sessions open unless state becomes inconsistent

### LINKED ISSUE #151: Can't retain passphrase state on Trezor T
This is actually two closely related issues. Both are on a Trezor T with default config and passphrase enabled.

Unlike Trezor One, TT has the ability to enter passphrase on device. So HWI's method of passing `--password` argument doesn't work. Instead, every time you perform a password-protected action, the device shows a pop-up asking for the passphrase.

Issues:
1. When you run any action (e.g., `hwi displayaddress`), you are asked for the passphrase twice.
This is because you don't retain passphrase state between actions.
you are supposed to do something like
```python
# in enumerate():
saved_state = client.client.state
# in TrezorClient.__init__:
self.client = Trezor(..., state=saved_state)
```
2. You can't retain passphrase state across runs. So if some software used hwi to do several actions, they'd be swamped with passphrase pop-ups.
Not sure what the solution is here. If the CLI is stateless, then presumably you'd need a `get-passphrase-state` command and then allow supplying that on repeated runs, sort of what you do with `prompt-pin`/`send-pin`

achow101:
When the client is opened, we would send an `Initialize` message. However because a client may end up being opened many times in the process of executing one command, the saved state (in particular cached passphrases) would be lost each time the client is opened due to `Initialize` being sent. In order to have it keep that state, we don't want to send `Initialize` every time. Instead we will send `GetFeatures` which does the other part of what `Initialize` does (get the features for the device). `GetFeatures` failing would indicate that the state is inconsistent (which can be caused by command abort, exceptions, etc.), so in that case, `Initialize` is sent to clear the inconsistent state.

I tested this using the Trezor T emulator and the password prompt does not appear multiple times per command. Additionally the password prompt does not appear for every command.

Fixes #151

### DIFF
--- a/hwilib/devices/trezorlib/client.py
+++ b/hwilib/devices/trezorlib/client.py
@@ -178,7 +178,10 @@ class TrezorClient:
 
     @tools.session
     def init_device(self):
-        resp = self.call_raw(messages.Initialize(state=self.state))
+        resp = self.call_raw(messages.GetFeatures())
+        # If GetFeatures fails, try initializing and clearing inconsistent state on the device
+        if isinstance(resp, messages.Failure):
+            resp = self.call_raw(messages.Initialize())
         if not isinstance(resp, messages.Features):
             raise exceptions.TrezorException("Unexpected initial response")
         else:


achow101:
Can someone with a Trezor T check that this works? Mine's still in the mail.

The expected behavior is that you only get a passphrase prompt once when enumerating, and then not again for subsequent commands. For those subsequent commands, if you interrupt HWI (e.g. with Ctrl+C) while the Trezor is doing something, the next command should require you to enter your passphrase.

achow101:
I've been able to test this with a trezor T and it does work as expected.

PR #153, opened 2019-04-20
Include Windows UCRT in Windows build

### LINKED ISSUE #150: Error loading Python DLL on Windows 7
A Wasabi wallet user who is helping to test the HW integration got this error:

Screenshot from 2019-04-17 02-24-15

It looks like a library dependency problem. Is the binary cli tool expected to work on Win7?

achow101:
Older versions of Windows needs to have the Windows Universal C++ Runtime packaged with the binary to work. This PR adds to the `build_wine.sh` build script to download the DLLs for the UCRT and package them with the Windows build.

Fixes #150


nopara73:
tACK on win7. (Tested as I tested the binary you provided here: https://github.com/bitcoin-core/HWI/issues/150#issuecomment-485052324)

Also tACK that it doesn't ruin win10.

PR #154, opened 2019-04-20
Make versioned compressed packages for releases

achow101:
The current build scripts produce binary files named `hwi` and `hwi.exe` which is both unversioned and conflicting for the Linux and MacOS binaries. For the 1.0.0 binaries, I had compressed these and named them myself before uploading. This PR changes the build scripts to compress and name the packages too. Additionally the names will include the version number which I forgot to do for 1.0.0.


achow101:
Merging as travis indicates the correct files are created.

PR #156, opened 2019-04-21
Add Trezor Model T

nopara73:
# Notes

- I marked with `?` the functions I did not test myself.
- I can make a PR in a way that I add Trezor Model T to the last row so the diff would be easier to see if you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,29 +64,29 @@ The below table lists what devices and features are supported for each device.
 
 Please also see [docs](docs/) for additional information about each device.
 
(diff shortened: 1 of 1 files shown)


achow101:
ACK 95bd88e41d3b293043223ff93962e50d575e4871

achow101:
I'll merge this for now and update it later once I my Trezor T arrives and I can test the rest.

Sjors:
I tested a number of commands on the model T a while ago: https://github.com/bitcoin-core/HWI/pull/82#issuecomment-451396292

However given the recent refactoring it doesn't hurt to retest.

PR #157, opened 2019-04-23
Indicate in enumerate whether a device needs a passphrase or a pin

achow101:
This adds two fields to `numerate`: `needs_passphrase_sent` and `needs_pin_sent` which indicate whether a command with a `-p` option is needed or `promptpin` and `sendpin` need to be done. These fields will be true in the instances that a passphrase needs to be sent and have not been cached by the device. To avoid a chicken and egg problem with fingerprints on Trezors, if the device needs a passphrase but one was not specified in `enumerate`, the fingerprint won't be retrieved in order to not contaminate the passphrase cache on device.

Built on #152 because it has some interactions with it that needed to be addressed.

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -225,6 +225,7 @@ def enumerate(password=''):
         path = d['path'].decode()
         d_data['type'] = 'coldcard'
         d_data['path'] = path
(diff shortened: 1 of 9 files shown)


achow101:
Closing and reopening to retrigger travis.

PR #159, opened 2019-04-29
Add packages to setup.py

jonasnick:
I'm packaging HWI for nix and therefore I need a normal install method using setup.py which doesn't require poetry. I don't know anything about python setup tooling, but just adding a list of packages allows me to both install it via `python setup.py install` as well as creating a nix package (WIP: https://github.com/fort-nix/nix-bitcoin/blob/3650198107e82352ffad0b46415b2e37d9bdfea7/pkgs/hwi/default.nix ).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup_kwargs = {
     'author_email': 'andrew@achow101.com',
     'url': 'https://github.com/bitcoin-core/HWI',
     'py_modules': modules,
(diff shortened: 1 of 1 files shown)


achow101:
Was `python setup.py install` not working before?

`setup.py` is generated automatically by poetry. So instead of modifying `setup.py` directly, `pyproject.toml` needs to be modified such that poetry generates `setup.py` with the packages listed.

jonasnick:
Without this PR `python setup.py build`, just creates single build artifact with `hwi.py`.

PR #160, opened 2019-04-29
Fix keypool refill instructions in docs

jonasnick:
The current instructions both give the same results, so I presume the first should be non-internal.


achow101:
ACK 483b589eebcb71f5abe2dd5dca95ebb15ed2dcc3

PR #164, opened 2019-04-29
pin poetry version to 0.12.12 as 0.12.14 is broken

achow101:
The latest version of poetry (0.12.14) is broken which is causing travis to fail. Pin the version to 0.12.13 for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ addons:
             - cython3
             - ccache
 install:
-    - pip install pipenv pysdl2 python-bitcoinrpc protobuf poetry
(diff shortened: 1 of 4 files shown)


achow101:
This will be merged when travis passes.

instagibbs:
like, poetry itself or some versioning difference with breaking change?

achow101:
> like, poetry itself or some versioning difference with breaking change?

poetry 0.12.14 itself is broken.

achow101:
Downgrading further to 0.12.12 as apparently 0.12.13 has a bug where the script isn't added to the path in the poetry virtualenv.

PR #165, opened 2019-04-29
Add packages to pyproject.toml to allow generating a working setup.py

jonasnick:
Fix of #159. Now setup.py was generated with `contrib/generate_setup.sh` after adding `hwi.py` and `hwilib` explicitly to `pyproject.toml`. This allows installing with `setup.py`.


achow101:
ACK ae9bca3ae386ec62c415a5b6fa5fa3b97069f7b9

PR #169, opened 2019-05-13
Add installudevrules command for linux

lontivero:
This PR is for discussing the concept and the implementation. The idea is to make it easier for Linux users to install the udev rules in their systems by running:

```
$ hwi installudevrules
```

### DIFF
--- a/hwilib/udevinstaller.py
+++ b/hwilib/udevinstaller.py
@@ -1,0 +1,60 @@
+import sys
+from subprocess import check_call, CalledProcessError, DEVNULL
+from .errors import NEED_TO_BE_ROOT
+from shutil import copy
+from os import path, environ, listdir, getlogin, geteuid
+
+class UDevInstaller(object):
+    @staticmethod
+    def install(source, location):
+        try:
+            udev_installer = UDevInstaller()
+            udev_installer.copy_udev_rule_files(source, location)
+            udev_installer.trigger()
+            udev_installer.reload_rules()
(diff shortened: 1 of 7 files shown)


achow101:
Would it be possible to add a test for this? Maybe add an argument to the command that specifies the location so for tests install the rules to that location and check the files are correct?

hwilib/udevinstaller.py
```diff
@@ -0,0 +1,60 @@
...
+            udev_installer = UDevInstaller()
+            udev_installer.copy_udev_rule_files(location)
+            udev_installer.trigger()
+            udev_installer.reload_rules()
+            udev_installer.add_user_plugdev_group()
+        except CalledProcessError as e:
+            if geteuid() != 0:
+                return {'error':'Need to be root.','code':NEED_TO_BE_ROOT}
+            raise
+        return {}
```
achow101: This should return `{"success": True}` instead of an empty obj.

lontivero:
@achow101 I had to duplicate code for error handling. Do you want I create a error handling function like this: https://github.com/bitcoin-core/HWI/pull/181#issuecomment-495306348?

achow101:
> I had to duplicate code for error handling. Do you want I create a error handling function

I think this is fine for now. I plan on making an error handling pr soon which can take care of this.

achow101:
Please squash

achow101:
ACK cada7f5ce0cf0854d7fc526b366107d5f0d868aa

PR #175, opened 2019-05-17
Support Nano Ledger X

Sjors:
Fixes #174.

It now detects both model S and model X, which is just a different device number. No need to update any drivers just yet, and there's no need to split `devices/ledger.py` for now.

I didn't test multisignature and advanced script types. Although I suspect it will work identically to the Nano S, I marked them with `?` in the overview.

I only tested via USB, not BlueTooth.

I noticed a bug where if the device falls asleep - it suffers from narcolepsy compared to the model S - and wakes up again, it won't shop anymore until you quit the Bitcoin (testnet) app and enter the app again. `enumerate` will think it's still asleep.

Being able to see a bech32 address on screen without having to scroll is awesome though.

PR #181, opened 2019-05-21
Provide error codes in enumerate output

achow101:
Adds error codes to `enumerate`'s output as suggested here: https://github.com/bitcoin-core/HWI/pull/175#discussion_r286126395

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -1,7 +1,7 @@
 # Coldcard interaction script
 
 from ..hwwclient import HardwareWalletClient
-from ..errors import ActionCanceledError, BadArgumentError, DeviceBusyError, UnavailableActionError, DeviceFailureError
+from ..errors import ActionCanceledError, BadArgumentError, DeviceBusyError, DeviceFailureError, HWWError, UnavailableActionError, UNKNOWN_ERROR
 from .ckcc.client import ColdcardDevice, COINKITE_VID, CKCC_PID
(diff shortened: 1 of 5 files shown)


nopara73:
I'm not a Python dev, but LGTM and straightforward.

jb55:
looks like a bunch of code duplication, could there a be an error handler util? ie:

```py
def handle_errors(res, fn):
    try:
        fn()
    except HWWError as e:
        res['error'] = "Could not open client or get fingerprint information: " + e.get_msg()
        res['code'] = e.get_code()
    except Exception as e:
        res['error'] = "Could not open client or get fingerprint information: " + str(e)
        res['code'] = UNKNOWN_ERROR
    return res

if __name__ == "__main__":
    data = {}

    def process():
        raise Exception("oh no")

    res = handle_errors(data, process)
    print(res)
```

achow101:
> looks like a bunch of code duplication, could there a be an error handler util?

Probably, but I think we can do that later. It would require rearranging quite a bit of the enumeration code which would make this more difficult to review.

jb55:
@achow101 you mind if I take a stab at it?

achow101:
Go ahead

jb55:
apparently it can be done quite elegantly with contextmanager + with blocks. TIL. here's a first sketch:

https://github.com/bitcoin-core/HWI/compare/26b8808c3b9573fd759cee2e04fed6fadb10005e~2..26b8808c3b9573fd759cee2e04fed6fadb10005e

will PR soon

PR #184, opened 2019-05-28
Clean up error handling in hwilib

jb55:


achow101:
Concept ACK

Can you also do this in cli.py?

jb55:
I'm not sure if these are covered by tests, I'm doing these refactors without a way of testing them.

jb55:
> Can you also do this in cli.py?

@achow101 done

hwilib/cli.py
```diff
@@ -204,38 +201,19 @@ def process_commands(cli_args):
...
-        result = {'error': e.get_msg(), 'code': e.get_code()}
-    except Exception as e:
-        if args.debug:
-            import traceback
-            traceback.print_exc()
-        result = {'error': str(e), 'code': UNKNOWN_ERROR}
-
-    # Close the device
-    try:
         client.close()
```
achow101: This still needs to be in its own `with` block. Otherwise if there's an error, the client won't be closed which can cause issues with using the device later.

jb55: ah right, good call

hwilib/cli.py
```diff
@@ -204,38 +201,21 @@ def process_commands(cli_args):
...
-    except Exception as e:
-        if args.debug:
-            import traceback
-            traceback.print_exc()
-        result = {'error': str(e), 'code': UNKNOWN_ERROR}
-
-    # Close the device
-    try:
+
+    with handle_errors(result=result, debug=args.debug):
```
jb55: one thing to keep in mind is that if there is an exception during client.close(), it will overwrite the error result above. not sure if this is much of a problem though.

achow101: That's not a problem. It's what it originally did.

achow101:
ACK f430e25ce2d764e8ff7cef39dc1ff76481788f30

PR #186, opened 2019-05-28
Fix bug in keepkey and trezor's enumerate

jb55:
Continued from #184. The commit to review is 1d25bab.

`not 'trezor'` would always evaluate to False, which wouldn't be in the list. So this should never continue here?

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -1,7 +1,7 @@
 # Coldcard interaction script
 
 from ..hwwclient import HardwareWalletClient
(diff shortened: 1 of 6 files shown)


achow101:
I think both ways (master and this PR) are correct. My understanding of how `not` works is that it will evaluate the `in` part first then the `not`, not the `not` then `in`, i.e. it is `not ('trezor' in ...)` rather than `(not 'trezor) in ...)`.

jb55:
hmm ok feel free to close, for some reason flake8 warned about it. might have been more of a style thing.

PR #188, opened 2019-06-05
Coldcard v2.1.0 and above does multisig

peter-conalgo:
If you give us the PSBT, perhaps with XPUB values in global section, we can sign it.

However, to support "display_address" we'll need a redeem script and paths for each co-signer. To achieve that, descriptor parsing needs major additions, and some API changes would be needed which would affect other devices.


achow101:
Does HWI need changes to support multisig on Coldcard? My understanding was that https://github.com/bitcoin/bips/pull/784 was needed for multisig on Coldcard.

peter-conalgo:
No changes needed in HWI to support multisig signing on Coldcard. However, some changes would be needed for the "show address" feature, when showing P2SH/P2WSH addresses.

This PR is just to update the feature matrix in the README.

achow101:
Needs rebase

peter-conalgo:
Rebased.

achow101:
ACK 3d86374905dc32fb55e891727dbd04083082c575

Issue #190, opened 2019-06-18
keepkey: enumerate is not actually covered by tests


jb55:
I noticed a bug in the devices/keepkey: `DeviceNotReadyError` is not imported, yet is referenced. I was wondering why the tests still pass, only to realize it seems to be testing the trezor emulator instead?

jb55:
see: https://github.com/jb55/HWI/commit/37ead9e85cc3729466d2d6e8f94ae4467a2e26ea

tests still pass: https://travis-ci.org/jb55/HWI/builds/547392947

the only other place that would produce a `Keepkey is locked` error message is here:

https://github.com/bitcoin-core/HWI/blob/4c9af14a5fc57feb3ed8e106e7e04796002a5e0d/hwilib/devices/trezor.py#L117

which is why I believe we're testing TrezorClient instead of keepkey.enumerate here?

achow101:
What happens if you just remove the keepkey build from `setup_environment.sh`? If it was running the trezor emulator, tests will still pass. Otherwise they should hang and timeout or something like that.

jb55:
The keepkey tests hang when I remove it from setup_environment, but that
still doesn't help explain why the functional tests see the "Keepkey is
locked" error even when I remove it. Perhaps there is some mocking going
on that is messing something up? I haven't had a chance to look into it
further.

instagibbs:
is this fixed?

jb55:
> is this fixed?

I don't think so. This commit current passes the tests:

https://github.com/jb55/HWI/commit/91e867584bff09cdc041c9b5604fd4e73d0f63f9
https://travis-ci.org/jb55/HWI/builds/606121826

Here is the keepkey test that is supposed to be covering this, and would
fail if it was since it's expecting the "Keepkey is locked" message.

    # Make sure we get a needs pin message
    result = self.do_command(self.dev_args + ['getxpub', 'm/0h'])
    self.assertEqual(result['code'], -12)
    self.assertEqual(result['error'], 'Keepkey is locked. Unlock by using \'promptpin\' and then \'sendpin\'.')

The only way it could be passing is if it was testing the trezor code instead (see line 111 in trezor.py:)

    def _check_unlocked(self):
        self.client.init_device()
        if self.client.features.pin_protection and not self.client.features.pin_cached:
            raise DeviceNotReadyError('{} is locked. Unlock by using \'promptpin\' and then \'sendpin\'.'.format(self.type))

achow101:
[...]

PR #191, opened 2019-06-18
keepkey: fix missing error import

jb55:
DeviceNotReadyError was missing from the import list. lint caught this
as a missing name error.

This was missed because the keepkey tests seem to be testing TrezorClient instead?

See #190

Signed-off-by: William Casarin <jb55@jb55.com>


achow101:
ACK 7dacdfe24c5dce3f3ac5f4f95e5ff647f6e1b744

PR #202, opened 2019-06-28
Always initialize Trezor One and Keepkey and clear any previous session

### LINKED ISSUE #199: trezor: password is ignored when already unlocked
When you unlock a trezor, all future commands will continue to use the unlocked passphrase, even when you pass different passwords.

### To reproduce

```
$ hwi promptpin
$ hwi sendpin
$ hwi --stdinpass enumerate
[{"needs_passphrase_sent": false, "needs_pin_sent": false, "type": "trezor", "path": "webusb:001:11", "fingerprint": "446284ab"}]

$ hwi --stdinpass enumerate # with different password
[{"needs_passphrase_sent": false, "needs_pin_sent": false, "type": "trezor", "path": "webusb:001:11", "fingerprint": "446284ab"}]
```

### Expected result

I would expect to git a different fingerprint if the password is changed, or at least to not re-prompt

achow101:
#152 Kept Trezor sessions open in between commands which allowed the passphrase to remain cached. This works well for the Trezor T where the user is prompted to enter their password the first time that anything from the device is queried (e.g. during enumerate) as the user has consciously entered the correct password. However on the Trezor One, the password is entered via the `-p` option and it is unintuitive that the password must be entered the first time the device is queried which results in the incorrect password being used.

To fix this, we instead revert to the original behavior of always initializing new sessions, but only for the Trezor One. This will clear the password cache every time a command is run and thus will always use the password provided on the command line for that command.

Also do the same for the Keepkey because it is basically identical to the Trezor One.

Fixes #199

### DIFF
--- a/hwilib/devices/keepkey.py
+++ b/hwilib/devices/keepkey.py
@@ -28,7 +28,7 @@ def enumerate(password=''):
             if not 'keepkey' in client.client.features.vendor:
                 continue
             d_data['needs_pin_sent'] = client.client.features.pin_protection and not client.client.features.pin_cached
-            d_data['needs_passphrase_sent'] = client.client.features.passphrase_protection and not client.client.features.passphrase_cached
+            d_data['needs_passphrase_sent'] = client.client.features.passphrase_protection # always need the passphrase sent for Keepkey if it has passphrase protection enabled
             if d_data['needs_pin_sent']:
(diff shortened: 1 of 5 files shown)


jb55:
awesome, will test this soon

jb55:
tested locally, seems to be more reasonable now. the fingerprint appears to be updating when the password changes.

jb55:
re-ACK, looks like it still works on my end after the latest rebase

PR #203, opened 2019-06-29
Add type hints for base58.py

fanquake:
TLDR: This add type hints to base58.py, as well as some basic tests and a Travis job that runs [`mypy`](http://mypy-lang.org). 

This adds some simple base58 tests (copied from the Bitcoin Core repo), [type hints](https://docs.python.org/3.6/library/typing.html) for base58.py and a Travis job that runs `mypy` on base58.py. I started with base58.py to keep everything self contained.

`mypy` was added as a dev dependency using `poetry add mypy --dev`.

Given that the project is using > Python 3.6, we can also take advantage [variable type annotations](https://www.python.org/dev/peps/pep-0526/).

Just opening as a draft for some concept level discussion. If there is interested then more anntotations can be added, and the Travis `mypy` scope increased. If we are happy to add type annotations here, then I'll PR the same changes to any upstream libraries, i.e [`bech32.py`](https://github.com/sipa/bech32/blob/master/ref/python/segwit_addr.py).

Another good reference for type hints is: https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,11 @@ jobs:
             install:
                 - pip install flake8
             script: flake8
+        -   name: Type annotation checking
+            stage: lint
+            install:
+                - pip install mypy
+            script: mypy hwilib/base58.py
         -   name: Run non-device tests only
             stage: test
             script: cd test; poetry run ./run_tests.py
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ hwi.egg-info/
 test/emulator.img
 test/work
 pip-wheel-metadata
+.mypy_cache/
--- a/hwilib/base58.py
+++ b/hwilib/base58.py
@@ -28,11 +28,7 @@ def encode(b):
     res = ''.join(res[::-1])
 
     # Encode leading zeros as base58 zeros
-    import sys
-    czero = b'\x00'
-    if sys.version > '3':
-        # In Python3 indexing a bytes returns numbers, not characters.
-        czero = 0
+    czero = 0
     pad = 0
     for c in b:
         if c == czero:
--- a/hwilib/base58.py
+++ b/hwilib/base58.py
@@ -11,25 +11,25 @@
 from .serializations import hash256
 import struct
 from binascii import hexlify, unhexlify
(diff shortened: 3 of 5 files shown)


jb55:
Approach ACK!

Can't go wrong with more type annotations.

IlyasRidhuan:
Concept Ack

Sjors:
@fanquake can you force push a trivial change when you're released from Travis prison? `git commit --amend --date "$(date)"` should do the trick.

fanquake:
@Sjors I've force pushed a change.

achow101:
The project is actually listed as allowing Python 3.5.6, so I believe this would break that. However I wouldn't be opposed to bumping the minimum to 3.6.

Also needs rebase.

fanquake:
Will rebase once #246 and #247 are in.

practicalswift:
Concept ACK: Python without type hints is like driving without a safety belt :-)

(Like C++ without the sanitisers or static analysis!)

achow101:
Tried using Python 3.5.6, our current minimum, and got:

```
  File "./run_tests.py", line 7, in <module>
    from test_base58 import TestBase58
  File "/home/andy/bitcoin/HWI6/test/test_base58.py", line 12
    TEST_VECTORS: List[Tuple[str, str]] = [
                ^
SyntaxError: invalid syntax
```

.travis.yml
```diff
@@ -54,6 +54,8 @@ jobs:
         -   name: lint
             stage: lint
             script: flake8
+        -   name: Type annotation checking
+            script: poetry run mypy hwilib/base58.py
```
achow101: Could you add an `install` phase that just installs `mypy` and whatever else is needed to run the check?

fanquake: Updated to use an `install` phase, and dropped the `mypy` dev dependency.

fanquake:
> Tried using Python 3.5.6, our current minimum, and got:

If you're still happy to bump the minimum Python version to 3.6.x, I can do that here. Otherwise I can propose it in a separate PR?

fanquake:
Rebased for #261 & #262.

laanwj:
ACK b739f38435f3792ba00f45edf824db56e169ede1

achow101:
ACK b739f38435f3792ba00f45edf824db56e169ede1

PR #205, opened 2019-07-10
Clean up keepkey/trezor emulator image after each test

hugohn:
**Problem**:
On my Mac, the Keepkey test suite succeeds the 1st time it is run, but starts hanging from the 2nd time onward:
`./run_tests.py --no_trezor_t --no_trezor --no_coldcard --no_bitbox`
`keepkey: test_enumerate (test_device.TestDeviceConnect) ... [HANG HERE]`

The reason is emulator.img, which is generated by the Keepkey emulator (kkemu), is not cleaned up afterwards which leaves the emulator in a bad state. Ideally, tests should leave their environment unchanged after they are run.

**Solution**:
Remove emulator.img from /bin after each Keepkey run.

### DIFF
--- a/test/test_keepkey.py
+++ b/test/test_keepkey.py
@@ -67,6 +67,11 @@ class KeepkeyEmulator(DeviceEmulator):
         self.emulator_proc.kill()
         self.emulator_proc.wait()
 
+        # Clean up emulator image
+        emulator_img = os.path.dirname(self.emulator_path) + "/emulator.img"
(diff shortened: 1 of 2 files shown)


achow101:
The trezor probably has this issue too, so I think we should also do this for trezor.

hugohn:
@achow101 agree. updated!

achow101:
ACK 76714edab08e5107209b7a189ec5d7b8619f7449 Tested locally and the images are removed

PR #211, opened 2019-07-12
 Enable trezor and keepkey change detection for standard single key ouputs

### LINKED ISSUE #170: Trezor displays change address for confirmation
[One of our users reported](https://www.reddit.com/r/WasabiWallet/comments/bntdkh/new_wasabi_user_hardware_wallet/) that the way Trezor displays tx confirmation change address, too. What can be done about it? Wait for Trezor to fix it in the driver or you're planning to fix it here?

> Sending coins- the Trezor asks me to confirm the address as usual however the first address presented on it is NOT the destination address. I assumed, correctly, that this must be the address for the change, and as it was for a tiny amount I took the chance. The second address presented on the Trezor was the destination address as expected. There could do with being an option to see the “change address” prior to it appearing on the Trezor, it’s really unnerving to be authorising a transaction to an address I have not seen anywhere! And it makes the whole confirmation on the Trezor pointless as I’m not actually checking it against any known (to me) address.

achow101:
Enable change detection for p2pkh, p2sh-p2wpk, and p2wpkh on the Trezor and Keepkey

Fixes #170

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -10,7 +10,7 @@ from .trezorlib.ui import echo, PassphraseUI, mnemonic_words, PIN_CURRENT, PIN_N
 from .trezorlib import protobuf, tools, btc, device
 from .trezorlib import messages as proto
 from ..base58 import get_xpub_fingerprint, decode, to_address, xpub_main_2_test, get_xpub_fingerprint_hex
-from ..serializations import ser_uint256, uint256_from_str
+from ..serializations import CTxOut, ser_uint256, uint256_from_str
 from .. import bech32
 from usb1 import USBErrorNoDevice
 from types import MethodType
@@ -243,7 +243,7 @@ class TrezorClient(HardwareWalletClient):
 
             # prepare outputs
             outputs = []
-            for out in tx.tx.vout:
+            for i, out in py_enumerate(tx.tx.vout):
                 txoutput = proto.TxOutputType()
                 txoutput.amount = out.nValue
                 txoutput.script_type = proto.OutputScriptType.PAYTOADDRESS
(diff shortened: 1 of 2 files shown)


instagibbs:
Is there a way of scraping the simulator screen output to test the change detection? Would be nice to automatically test the change address doesn't appear on the screen.

achow101:
> Is there a way of scraping the simulator screen output to test the change detection? Would be nice to automatically test the change address doesn't appear on the screen.

It is possible to get the entire raw buffer of the display, but I don't think that is particularly useful.

hwilib/devices/trezor.py
```diff
@@ -258,6 +258,26 @@ def ignore_input():
                     else:
                         raise BadArgumentError("Output is not an address")
 
+                # Add the derivation path for change, but only if there is exactly one derivation path
+                psbt_out = tx.outputs[i]
+                if len(psbt_out.hd_keypaths) == 1:
```
instagibbs: just noting the ledger detection is more eager, trying any number of hd keypaths that appear to match. We could tighten that logic up in a follow-on PR.

hwilib/devices/trezor.py
```diff
@@ -258,6 +258,26 @@ def ignore_input():
...
+                    _, keypath = next(iter(psbt_out.hd_keypaths.items()))
+                    if keypath[0] == master_fp:
+                        wit, ver, prog = out.is_witness()
+                        if out.is_p2pkh():
+                            txoutput.address_n = keypath[1:]
+                            txoutput.address = None
+                        elif wit:
+                            txoutput.script_type = proto.OutputScriptType.PAYTOWITNESS
+                            txoutput.address_n = keypath[1:]
+                            txoutput.address = None
```
instagibbs: without being able to test this locally, does this mean that *any* output with known derivation path will be considered change and elided from the screen?

I do a slightly hackier check for ledger, checking that 2nd to last index is 1(hdsplit): https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/ledger.py#L158 but in my real-world use it behaves much better for self-sending scenarios.

achow101: > without being able to test this locally, does this mean that _any_ output with known derivation path will be considered change and elided from the screen?

It would appear it does. If the transaction is fully send to self, then no amounts are shown on screen. It does still check that the address derived matches the address given and if it does not match, it will be shown.

instagibbs: :man_shrugging: ok. Maybe document that, keeping logic the same.

achow101: Added a note in `docs/trezor.md`

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/211/commits/3dd66bd09413c8d556c8ed5413428de3a5e6feca

PR #218, opened 2019-07-16
Simplify setup.py

promag:


achow101:
NACK. This file is automatically generated to keep it consistent with pyproject.toml.

promag:
Oh sorry, I wasn't aware of that.

PR #220, opened 2019-07-17
Replace dead "bytes_to_hex_str" with "binascii.hexlify"

### LINKED ISSUE #219: `bytes_to_hex_str` not imported in `serializations.py`
`bytes_to_hex_str` is used [here](https://github.com/bitcoin-core/HWI/blob/master/hwilib/serializations.py#L239) and [here](https://github.com/bitcoin-core/HWI/blob/master/hwilib/serializations.py#L290) but doesn't exist as far as I can tell, causing a `NameError` if either `__repr__` is ever exercised

justinmoon:
Fixes #219

### DIFF
--- a/hwilib/serializations.py
+++ b/hwilib/serializations.py
@@ -236,7 +236,7 @@ class CTxIn(object):
 
     def __repr__(self):
         return "CTxIn(prevout=%s scriptSig=%s nSequence=%i)" \
-            % (repr(self.prevout), bytes_to_hex_str(self.scriptSig),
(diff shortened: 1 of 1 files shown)


instagibbs:
`.hex()` not usable instead? minimum python version is enough for it.

justinmoon:
I used `binascii` because it was used elsewhere in the file. Updated to use `.hex()`!

achow101:
Please squash

achow101:
Code review ACK bccf63288ba485e2a331dc4170b0e14233ecf53d

PR #228, opened 2019-07-24
Add device model to enumerate and accept it in --device-type

### LINKED ISSUE #223: [Feature] Introduce "--device-model"
It"d be the same as "device-type" but more specific. Eg. trezor -> trezort + trezorone, leger -> ledgernanos, ledgernanox.

This'd be especially useful in some cases, like trezor setup. With Trezor T I can just tell the device to setup and it does everything by itself, however with Trezor One, I have to fire up a hwi console app, because setup requires a fair bit of interaction.

However I cannot differentiate between them based on what hwi enumerate tells me, so I'll have to ask the user what device he has before giving out the command.

achow101:
In the enumerate output, add a `model` field which has the full vendor and model name (as reported by the device) of the device all lowercase and separated by underscores. So a Trezor One would have the model name `trezor_1` and a Trezor T would have `trezor_t`. For some devices which don't have distinct model names (such as Coldcard), the `model` will the be same as `type`. Additionally, model names for simulators will also have `_simulator` at the end of it.

These model names can then be used in the `--device-type` parameter. For now, using the device model in `--device-type` just strips the actual model name (stuff after the first underscore) and does the same thing as just giving the original device type. This could change later if other device models change their APIs.

The `type` will still be in the `enumerate` output and will still work for `--device-type`.

Current model names that can be produced:
* `coldard`
* `coldard_simulator`
* `digitalbitbox_01`
* `digitalbitbox_01_simulator`
* `keepkey`
* `keepkey_simulator`
* `ledger_nano_s`
* `trezor_1`
* `trezor_1_simulator`
* `trezor_t`
* `trezor_t_simulator`

Closes #223


hwilib/commands.py
```diff
@@ -44,7 +45,7 @@ def enumerate(password=''):
 def find_device(device_path, password='', device_type=None, fingerprint=None):
     devices = enumerate(password)
     for d in devices:
-        if device_type is not None and d['type'] != device_type:
+        if device_type is not None and d['type'] != device_type and d['model'] != device_type:
```
instagibbs: do we want to attempt opening it if `device_type` is `None`?

instagibbs: I suspect this is supposed to be an `or`

achow101: No, this is correct.

When `device_type` is `None`, we are trying to find the device via fingerprint. So if it is `None`, then we still continue and just try every device. If `device_type` is not `None`, then the user has specified a type and we need to filter on that. So we will abort if the specified type does not match both the type and the model given by `enumerate`. If one of them match, then we should not skip checking this device but rather continue.

achow101:
I've added a test

instagibbs:
utACK changes https://github.com/bitcoin-core/HWI/pull/228/commits/a3957e9dee395aa5c888e08c866d4feb84334594

PR #231, opened 2019-08-19
Point pip user to python deps in pyproject.toml

instagibbs:
It was an incomplete list, so confusing at best for users who want to manually install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,16 +27,8 @@ Once HWI's source has been downloaded with git clone, it and its dependencies ca
 poetry install
 ```
 
-Pip can also be used to install all of the dependencies (in virtualenv or system):
+Pip can also be used to install all of the dependencies (in virtualenv or system) required for operation and development. See `pyproject.toml` for all dependencies. Dependencies under `[tool.poetry.dependecies]` are user dependencies, and `[tool.poetry.dev-dependencies]` for development based dependencies.
 
-```
-pip3 install hidapi # HID API needed in general
-pip3 install ecdsa
-pip3 install pyaes
-pip3 install typing_extensions
-pip3 install mnemonic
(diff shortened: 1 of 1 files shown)


instagibbs:
FWIW I was missing a few for Trezor, and had to have it print out the error to figure out which packages were necessary.

achow101:
I think it makes more sense to point people to `pyproject.toml` rather than `poetry.lock`. `poetry.lock` has every single package, including all of the dependencies of each of the packages we rely on. If you just install those listed in `pyproject.toml`, all of those dependencies will be installed too.

instagibbs:
oh that's better, will update

instagibbs:
updated

README.md
```diff
@@ -27,16 +27,8 @@ Once HWI's source has been downloaded with git clone, it and its dependencies ca
 poetry install
 ```
 
-Pip can also be used to install all of the dependencies (in virtualenv or system):
+Pip can also be used to install all of the dependencies (in virtualenv or system) required for operation and development. See `pyproject.toml` for all dependencies.
```
achow101: It's probably helpful to also say that the dependencies for normal operation are those under `[tool.poetry.dependencies]` except for `pywin32-ctypes` which is Windows only (and of course Python itself which can't be installed with pip). To run the tests and build the binaries, you will need to install those under `[tool.poetry.dev-dependencies]` too.

instagibbs: trying not to make this too verbose, hopefully that windows one will throw a legible error.

instagibbs: made it a bit more verbose.

achow101:
ACK 0f09706f3155e70ea6257a7de46dcea1882f3b4d

PR #238, opened 2019-08-21
Improve setup_environment.sh instructions

Sjors:


test/README.md
```diff
@@ -18,8 +18,8 @@ It also tests usage with `bitcoind`.
 It uses the [Coldcard simulator](https://github.com/Coldcard/firmware/tree/master/unix#coldcard-desktop-simulator).
 It also tests usage with `bitcoind`.
 
-`setup_environment.sh` will build the Trezor emulator, the Coldcard simulator, the Keepkey emulator, the Digital Bitbox simulator, and `bitcoind`.
-if run in the `test/` directory, these will be built in `work/test/trezor-mcu`, `work/test/firmware`, `work/test/keepkey-firmware`, `work/test/mcu`, and `work/test/bitcoin` respectively.
+There is a script to build the Trezor emulator, the Coldcard simulator, the Keepkey emulator, the Digital Bitbox simulator, and `bitcoind`. On Ubuntu / Debian first install: `sudo apt-get install libsdl2-dev`. On all systems: `pip install pipenv`
```
achow101: We already list the specific dependencies below. I think it would be better to just point people down there instead of trying to list them here.

test/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Running the tests
 
-This folder contains test cases for HWI. To run these tests, `hwilib` will need to be installed to your python system. You can install it by doing `pip install -e .[tests]` in the root directory.
+This folder contains test cases for HWI. To run these tests, `hwilib` will need to be installed to your python system. You can install it by doing `pip install -e .[tests]` in the root directory. In addition install `pip install hidapi python-bitcoinrpc`.
```
achow101: Instead of listing specific dependencies, I think we should take the approach that was done in the main readme Just tell people to go look in `pyproject.toml` and have them install everything under `[tool.poetry.dependecies]` and `[tool.poetry.dev-dependencies]`.

Also `[tests]` is no longer an install target, so we can drop that. So the command for installing hwilib will be `pip install -e .`.

achow101:
Closing due to lack of activity.

Sjors:
Oops, I completely forgot about this PR. I don't even remember making it, so let's call it up for grabs.

PR #239, opened 2019-08-23
Include udev rules with PyPI distribution

### LINKED ISSUE #234: "installudevrules" fails in virtual environment
From a vanilla Python 3 virtual environment:

```
$ hwi installudevrules
{"error": "installudevrules failed: [Errno 2] No such file or directory: 'udev'", "code": -13}
```

achow101:
The PyPI did not have the udev rules included so users who installed via `pip` would not be able to use `installudevrules`. This PR fixes that.

The udev rules are also moved from the `udev/` to `hwilib/udev` to facilitate this. Additionally the udev rules installer now looks for the udev rules relative to `hwilib/udevinstaller.py` which allows for it to more consistently find the udev rules. `installudevrules`'s hidden `--source` argument has also been removed for now.

Lastly, since this was something that should have been caught by tests (we have udev installer tests), I've setup two new travis builds that will install and test both the wheel and source distributions.

Closes #234


achow101:
cc @justinmoon @lontivero

PR #241, opened 2019-08-29
Easier on the eyes device support table

tkijewski:
HWI: Bitcoin Hardware Wallet Interaction scripts 2019-08-29 11-15-52


instagibbs:
visually scanned the diff, looks right. utACK https://github.com/bitcoin-core/HWI/pull/241/commits/0f981623c7c604ee6010dbccbb7be3b8ee9a8f11

achow101:
Are these checkmarks and stuff rendered with other markdown renderers? The readme is also shown on PyPI, and it would also be useful to be able to read the file as plaintext too.

tkijewski:
> Are these checkmarks and stuff rendered with other markdown renderers? The readme is also shown on PyPI, and it would also be useful to be able to read the file as plaintext too.

I am not sure - but just to be safe I switched to unicode "ballot box" `&#9745;` &#9745; and `&#10007;` &#10007;

HWI: Bitcoin Hardware Wallet Interaction scripts 2019-08-29 16-38-27


I didn't think about plaintext readability, in which case this might not be a good idea.

instagibbs:
what's the merge commit for?

tkijewski:
I updated this to include the Ledger Nano X column. 

I reverted back to the original check mark `:white_check_mark:` for now. It is github flavored markdown, so it will work in github/PyPI. Not so much other markdown renderers. The unicode ones will work everywhere: `&#9745;` ☑ and `&#10007;` ✗

There is a separate plaintext description for PyPI that remains unaffected by this PR at this time: https://github.com/bitcoin-core/HWI/blob/5309f294b02f4349dd55a5739ada875172cf440e/setup.py#L36 

@achow101 After thinking through the tradeoffs, I don't think this PR is worth it...function over fashion. table does look cool though.

PR #242, opened 2019-08-29
Replace python-bitcoinrpc with Core's authproxy

achow101:
It seems python-bitcoinrpc is causing issues with the ledger tests due to the RPC connection timing out. The fork of it used in Bitcoin Core's test framework has code to handle this, so instead of depending on python-bitcoinrpc, we should just pull in and use Core's fork of it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ addons:
             - cython3
             - ccache
 install:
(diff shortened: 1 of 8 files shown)


instagibbs:
seems to fix nano_whatever tests. tACK

Sjors:
All tests pass for Nano S with 6a582cf. I didn't try updating dependencies.

instagibbs:
reACK, only dep changes

PR #137, opened 2019-03-08
Add getdescriptors command

Sjors:
Returns all relevant descriptors for a given account, e.g.:

```sh
./hwi.py --fingerprint 00000000 getdescriptors --account 0
```

```json
{
  "receive": [
    "pkh([00000000/44h/0h/0h]xpub6C.../0/*)#fn95jwmg",
    "sh(wpkh([00000000/49h/0h/0h]xpub6B..../0/*))#j4r9hntt",
    "wpkh([00000000/84h/0h/0h]xpub6C.../0/*)#qw72dxa9"
  ],
  "internal": [
    "pkh([00000000/44h/0h/0h]xpub6C.../1/*)#c8q40mts",
    "sh(wpkh([00000000/49h/0h/0h]xpub6B..../1/*))#85dn0v75",
    "wpkh([00000000/84h/0h/0h]xpub6C..../1/*)#36mtsnda"
  ]
}
```

PR #246, opened 2019-09-05
Add flake8 linter and do style fixes

### LINKED ISSUE #192: repo: add pyflakes lint checks on pull requests
flake8 would have caught bugs such as https://github.com/bitcoin-core/HWI/pull/191

It would be cool if we have lint checks on pull requests, perhaps just the error level, warning level is a bit nazi.

### LINKED ISSUE #221: InvalidBase58Error doesn't exist
[`InvalidBase58Error` in `hwilib/base58.py`](https://github.com/bitcoin-core/HWI/blob/a3610b23ae37fc46e0ade0ba579a6233825bb7c4/hwilib/base58.py#L52) doesn't exist.

achow101:
Adds support for using the flake8 linter and a travis job that runs it to check for code style and other errors. Also adds a script for using the autopep8 formatter and applies the suggested changes. Both autopep8 and flake8 are added as developer dependencies in pyproject.toml.

Most of the fixes can be done by running an autopep8 command and are indicated as such in the commit message using the scripted-diff format even though scripted diffs aren't actually being checked by travis.

Closes #221
Closes #192

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+exclude = *.pyc,__pycache__,hwilib/devices/btchip/,hwilib/devices/ckcc/,hwilib/devices/trezorlib/,test/work/
+ignore = E261,E302,E305,E501,E722,W5
+per-file-ignores = setup.py:E122
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,15 +51,23 @@ install:
     - poetry install
 jobs:
     include:
+        -   name: lint
+            stage: lint
+            script: flake8
         -   name: Run non-device tests only
+            stage: test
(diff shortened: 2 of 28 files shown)


achow101:
Rebased following merges of #137 and #247

practicalswift:
ACK d07e102e95849d5a619acb4f96598e132453aecc -- diff looks correct

In addition to `flake8`/`autopep8`: what about also using `black` to reduce the degrees of freedom when it comes to DIY formatting?

achow101:
> In addition to `flake8`/`autopep8`: what about also using `black` to reduce the degrees of freedom when it comes to DIY formatting?

I don't care about formatting *that* much and IMO black's style is too restrictive.

Issue #136, opened 2019-03-04
getkeypool should `--keypool` by default?


instagibbs:

jb55:
I believe this has key reuse implications? What core probably needs is a way to semi-intelligently import keys to the keypool or not based on if they have any transaction associated with them, if I'm understanding the purpose of keypool correctly...

jb55:
some more context from @achow101 since I didn't realize the subtleties of keypool imports:

```
#bitcoin-core-dev
21:56 <jb55> achow101: will core re-use keypool-imported addresses if they've been used already? if not then there's no issue.
21:57 <achow101> jb55: only if you rescan after importing
21:57 <achow101> rescanning will remove used things from the keypool
```

it seems currently core will only remove used keys from the keypool on the next rescan, not on the initial one.

Sjors:
There is no rescan when you import keys using the instructions in this repo. You have to call rescan yourself. At that point used keys are removed from the keypool.

jb55:
if that's the case then :+1: for keypool default. will verify...

instagibbs:
this continues to mess me up, we should really do this

PR #249, opened 2019-09-06
Make keypool the default, switch option to --nokeypool

instagibbs:
resolves https://github.com/bitcoin-core/HWI/issues/136

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -125,7 +125,9 @@ def process_commands(cli_args):
     signmsg_parser.set_defaults(func=signmessage_handler)
 
     getkeypool_parser = subparsers.add_parser('getkeypool', help='Get JSON array of keys that can be imported to Bitcoin Core with importmulti')
-    getkeypool_parser.add_argument('--keypool', action='store_true', help='Indicates that the keys are to be imported to the keypool')
+    kparg_group = getkeypool_parser.add_mutually_exclusive_group()
+    kparg_group.add_argument('--keypool', action='store_true', dest='keypool', help='Indicates that the keys are to be imported to the keypool', default=True)
+    kparg_group.add_argument('--nokeypool', action='store_false', dest='keypool', help='Indicates that the keys are not to be imported to the keypool', default=False)
     getkeypool_parser.add_argument('--internal', action='store_true', help='Indicates that the keys are change keys')
     getkeypool_parser.add_argument('--sh_wpkh', action='store_true', help='Generate p2sh-nested segwit addresses (default path: m/49h/0h/0h/[0,1]/*)')
     getkeypool_parser.add_argument('--wpkh', action='store_true', help='Generate bech32 addresses (default path: m/84h/0h/0h/[0,1]/*)')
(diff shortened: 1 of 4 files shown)


instagibbs:
rebased

achow101:
This is a breaking change to the API. Could it instead have `--keypool` remain, but have also have `--keypool=false`?

instagibbs:
doesn't look like that's possible, maybe add nokeypool as the new way to turn it off, but still accept both(mutually exclusive)?

achow101:
> doesn't look like that's possible, maybe add nokeypool as the new way to turn it off, but still accept both(mutually exclusive)?

That's fine

instagibbs:
updated. I couldn't figure out fancy mutually exclusive args so I did it manually with a couple extra lines.

achow101:
> I couldn't figure out fancy mutually exclusive args so I did it manually with a couple extra lines.

It's not that hard. We do this in tests, so just copy from that: https://github.com/bitcoin-core/HWI/blob/master/test/run_tests.py#L19

hwilib/cli.py
```diff
@@ -9,7 +9,8 @@
     HELP_TEXT,
     MISSING_ARGUMENTS,
     NO_DEVICE_TYPE,
-    UNAVAILABLE_ACTION
+    UNAVAILABLE_ACTION,
+    BAD_ARGUMENT,
```
achow101: `BAD_ARGUMENT` isn't needed (linter fails here)

hwilib/cli.py
```diff
@@ -125,7 +128,9 @@ def process_commands(cli_args):
     signmsg_parser.set_defaults(func=signmessage_handler)
 
     getkeypool_parser = subparsers.add_parser('getkeypool', help='Get JSON array of keys that can be imported to Bitcoin Core with importmulti')
-    getkeypool_parser.add_argument('--keypool', action='store_true', help='Indicates that the keys are to be imported to the keypool')
+    kparg_group = getkeypool_parser.add_mutually_exclusive_group()
+    kparg_group.add_argument('--keypool', action='store_const', const=True, help='Indicates that the keys are to be imported to the keypool (default: true)')
+    kparg_group.add_argument('--nokeypool', action='store_const', const=True, help='Indicates that the keys are not to be imported to the keypool (default: false)')
```
achow101: It would be cleaner to have both of these use `dest='keypool'` and `action='store_true'` (and remove the `const=...`). `--keypool` will need a `default=True` also. Then later we can still use `args.keypool` instead of having to check between both `args.keypool` and `args.nokeypool`.

Also, please remove the `(default: ..)` parts of the help text. We should show the defaults for every option, and argparse has a way to do this without specifying the default in every help time via the use of the `default=...` parameter.

instagibbs: done

instagibbs: done(though I had to add some more args to make it work, guess and check)

achow101:
Tested ACK 156b1ce6ecf293970e84fedf27331718a11161ea

PR #253, opened 2019-09-09
Fix btchip error module name

instagibbs:
Make ledger use possible.


achow101:
ACK f043d6dab4f8da3347122ab85604e4e7850d85bc

Tests for ledger pass now. We really need some kind of CI for ledgers...

PR #255, opened 2019-09-13
Print out hash of message being signed for verification

instagibbs:
Really this is only useful for documenting exactly how the sighash is computed(for verification of what you're signing on a known good host?). I'm actually leaning concept NACK on this, but wanted to open this up to see if I'm alone on this.

### DIFF
--- a/hwilib/devices/ledger.py
+++ b/hwilib/devices/ledger.py
@@ -12,7 +12,7 @@ import hid
 import struct
 from .. import base58
 from ..base58 import get_xpub_fingerprint_hex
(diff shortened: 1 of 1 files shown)


achow101:
I would prefer to not always show this to the user. But it may be useful in debugging, so add it as debug output (hidden behind `--debug` switch)?

Should also do this for bitbox.

instagibbs:
hmmm not sure it's worth it tbh. I can't imagine a scenario where it's useful for debugging(the debugger would just reinvent it like I just did).

PR #258, opened 2019-09-20
Allow CLI to limit enumerate search to specific device type

PierreRochard:
This is helpful when debugging a specific device module import/dependency issue.

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -26,7 +26,7 @@ def displayaddress_handler(args, client):
     return displayaddress(client, desc=args.desc, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh)
 
 def enumerate_handler(args):
-    return enumerate(password=args.password)
+    return enumerate(password=args.password, device_type=args.device_type)
 
 def getmasterxpub_handler(args, client):
     return getmasterxpub(client)
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -3,6 +3,7 @@
 # Hardware wallet interaction script
 
 import importlib
+import logging
 import platform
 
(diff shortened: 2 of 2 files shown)


instagibbs:
argh travis is down, we've been flagged for "abuse". Probably because "mining"

achow101:
Closing and reopening to trigger travis

hwilib/commands.py
```diff
@@ -30,14 +31,22 @@ def get_client(device_type, device_path, password=''):
...
+        modules = all_devs
+    else:
+        modules = [device_type]
+
+    for module in modules:
         try:
             imported_dev = importlib.import_module('.devices.' + module, __package__)
             result.extend(imported_dev.enumerate(password))
-        except ImportError:
+        except ImportError as e:
```
achow101: This block is an unrelated change.

hwilib/cli.py
```diff
@@ -26,7 +26,7 @@ def displayaddress_handler(args, client):
     return displayaddress(client, desc=args.desc, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh)
 
 def enumerate_handler(args):
-    return enumerate(password=args.password)
+    return enumerate(password=args.password, device_type=args.device_type)
```
achow101: I'm not sure that this is safe if device type is not specified.

Also, the device type may not necessarily be just the type, it could also be the model name in which the changes to `enumerate()` would not work.

instagibbs:
you failed to trigger travis apparently. @PierreRochard could you clean up PR and push again to see if it gets triggered?

achow101:
Closing due to lack of activity. marking as up for grabs

Issue #248, opened 2019-09-06
coldcard multisig testing coverage


instagibbs:
we need it

@justinmoon you said there's a bug currently with native segwit outputs? may want to wait until that is fixed.

justinmoon:
@peter-conalgo just released a new ColdCard firmware that he says fixes it. Have not tested yet, but will today.

instagibbs:
@justinmoon I forget, is this confirmed working? Thinking about writing support again here.

achow101:
Multisig on coldcard works so long as the xpubs are registered with the device first. However an important thing to note is that they require BIP 67, so a `sortedmulti` descriptor would need to be used.

Their other methods of registering xpubs requires supporting the PSBT xpub field which we do not yet support.

PR #260, opened 2019-09-23
[WIP] Enable multisig testing for coldcard in CI

instagibbs:
Probably doesn't Just Work but will give me some hint on what more there is to do.

would resolve https://github.com/bitcoin-core/HWI/issues/248

### DIFF
--- a/test/test_device.py
+++ b/test/test_device.py
@@ -419,7 +419,7 @@ class TestSignTx(DeviceTestCase):
     # Test wrapper to avoid mixed-inputs signing for Ledger
     def test_signtx(self):
         supports_mixed = {'coldcard', 'trezor_1', 'digitalbitbox', 'keepkey'}
-        supports_multisig = {'ledger', 'trezor_1', 'digitalbitbox', 'keepkey'}
+        supports_multisig = {'ledger', 'trezor_1', 'digitalbitbox', 'keepkey', 'coldcard'}
         if self.full_type not in supports_mixed:
             self._test_signtx("legacy", self.full_type in supports_multisig)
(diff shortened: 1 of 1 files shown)


instagibbs:
ok doesn't work as I expected. Need to figure out the new image situation.

stepansnigirev:
Without change detection it should work if coldcard has the corresponding multisig wallet imported.
Example of the multisig wallet:
```
Name: MyMultisigWallet
Policy: 2 of 2
Derivation: m/48'/1'/0'/2'
Format: P2WSH
7C16EF20: tpubDFKT3s2snbsbu9ABgNwkne3pe2ygiXP3hWpoFhrprJP2ySsSVKqCZq8tuCBXXon4RiXAVpC1QnrjdA8y6Mk8WxcianZByPEm4T2zy2fWgN7
F79EC910: tpubDEe6NkFUArzPR4AZ5q5kKwaWLZjQhPjD3xqQbE51kQ6xb9cANrUyeauYtHrwTjX8z32xBHzVKYAiAoykBwt5LX1vb7j163RuMJn9qrQ8Hnc
```
Saving it as `.txt` file and importing it from the SD card would allow you to sign multisig transactions on the ColdCard (it should be possible to do over USB as well, but I don't know the exact command).
If you also want to detect change addresses you also need to provide xpubs in the global scope of the PSBT (according to the bip174 - `0x01<xpub>`: `<fingerprint><hardened_derivation>`).
The problem here is that HWI is stateless so if co-signers are not connected we don't know their xpubs.

instagibbs:
no time to carry this, sorry

PR #262, opened 2019-09-24
Bump minimum python version to 3.6

achow101:
Needed for #203

This is not expected to impact existing projects using HWI. AFAIK there are only two projects that use HWI, [Wasabi Wallet](https://github.com/zkSNACKs/WalletWasabi) and [Junction](https://github.com/justinmoon/junction). Wasabi use the binary distribution so it is not effected. Junction also requires 3.6+.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -191,8 +191,8 @@ version = "3.7.4"
 typing = ">=3.7.4"
 
 [metadata]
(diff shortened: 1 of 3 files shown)


achow101:
Note that `^` is used here instead of `>=` so that the python version remains within python 3. This is so that poetry's dependency resolver will allow us to include a library needed for Bitbox 2 support which has a similar python version requirement and would not be added if `>=` were used.

PR #264, opened 2019-09-27
psbt serialization fix

stepansnigirev:
unknowns should be in global scope


achow101:
Hmm. I thought we had a test for this... if not, could you add a test?

stepansnigirev:
Added a test with unknowns. These unknowns are actually not-yet-implemented xpub fields.

achow101:
ACK aff5f57d49afc0637ee521fcd1012a8f48a1d91d

Reviewed code and checked that the test fails on master but not with this PR.

nopara73:
@stepansnigirev would you mind to give me a short description on what this PR was about? I'm mainly interested if this affects Wasabi users or not.

stepansnigirev:
This is mostly related to ColdCard multisignature signing.
If psbt has extra fields in global scope (i.e. cosigner's xpubs) HWI would serialize it incorrectly.
As you are only using single key inputs and don't put any unknown fields to the global scope it doesn't affect your users.

instagibbs:
@stepansnigirev oh you might have some ideas for https://github.com/bitcoin-core/HWI/pull/260 then if you know how it works. Haven't had the time to dive in to figure out how to fix it.

PR #267, opened 2019-10-08
Better keepkey enumeration

### LINKED ISSUE #230: KeepKey detected by `enumerate` when Trezor plugged in
```
$ hwi enumerate
[{'code': -13,
  'error': 'Could not open client or get fingerprint information: '
           'LIBUSB_ERROR_BUSY [-6]',
  'path': 'webusb:001:1',
  'type': 'trezor'},
 {'code': -14,
  'error': 'Could not open client or get fingerprint information: '
           'get_pubkey_at_path canceled',
  'path': '0003:0016:00',
  'type': 'ledger'},
 {'code': -13,
  'error': 'Could not open client or get fingerprint information: '
           'LIBUSB_ERROR_BUSY [-6]',
  'path': 'webusb:001:1',
  'type': 'keepkey'}]
```

I have a Trezor One and a Ledger Nano-S plugged in. Just prompted a pin for the Trezor just before this from python and that `TrezorClient` instance was still active. No KeepKey has ever been plugged into my machine.

achow101:
Since Trezor and Keepkey both use the Trezor library for everything, sometimes the stated device type gets mixed up in enumeration when something goes wrong. To try to avoid this, do some additional filtering on the USB vendor id if it is available. Otherwise, fallback to the original filtering. This should work for all physical devices, and the fallback is only for the simulators.

Fixes #230


jb55:
perhaps this fixes #190?

achow101:
I don't know if this fixes #190, but it definitely should fix #230

jb55:
I tried to test if it fixed #190 but travis blocked HWI on my account, so I'm waiting for them to unblock that =/

instagibbs:
@jb55 did you get unblocked?

jb55:
> @jb55 did you get unblocked?

oh yeah and then I forgot about this. will try again soon

PR #270, opened 2019-10-18
Reduce Trezor UDP timeout to 1 second

### LINKED ISSUE #269: `hwi enumerate` taking 20 seconds on Windows
Every once in a while `hwi enumerate` will take 20 seconds on my Windows 10 machine.

In such cases, [this timeout](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezorlib/transport/udp.py#L88) gets hit for both trezor and keepkey, adding up to 20 second duration.

@nopara73 @lontivero Do you guys ever see this?

Thoughts on lowering the timeout? (cc @prusnak)

achow101:
Fixes #269


achow101:
Travis passes which is all that really matters with the UDP interface since it's only for the emulators. So merging this.

PR #274, opened 2019-10-28
dbb: Handle string error codes too

achow101:
Some bitbox firmware versions may return the error code as a string rather than an integer, handle those as well.

Should fix travis.


achow101:
Passes travis, merging.

PR #275, opened 2019-10-29
Better cleanup of emulators after tests complete

achow101:
Use SIGTERM to stop emulators/simulators in tests instead of relying on them to shut themselves down via normal means. This is safe since they're data is ephemeral anyways.

This should fix the travis timeouts as it was waiting for emulators to shut down. By using SIGTERM, they should all terminate.

### DIFF
--- a/test/test_coldcard.py
+++ b/test/test_coldcard.py
@@ -3,34 +3,36 @@
 import argparse
 import atexit
 import os
+import signal
 import subprocess
 import time
 import unittest
 
 from hwilib.cli import process_commands
-from hwilib.devices.ckcc.protocol import CCProtocolPacker
-from hwilib.devices.ckcc.client import ColdcardDevice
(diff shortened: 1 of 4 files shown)


instagibbs:
concept ACK

linter fun

```
./test/test_coldcard.py:11:1: F401 'hwilib.devices.ckcc.protocol.CCProtocolPacker' imported but unused
./test/test_coldcard.py:12:1: F401 'hwilib.devices.ckcc.client.ColdcardDevice' imported but unused
```

achow101:
> linter fun

Bah. Fixed.

Also changed Coldcard to terminate the process group (because apparently it spawns child processes).

instagibbs:
ACK https://github.com/bitcoin-core/HWI/pull/275/commits/6513842447ff9e8c1c55ef5a33e021ef70f9da9c

PR #278, opened 2019-11-13
Use mnemonics with getxpub tests

achow101:
Since Trezor T doesn't support setting up with an xprv, and Trezor 1 was recently changed to no longer support this, we need to setup the devices with mnemonics so that getxpub can be tested on them. So the test vectors in bip32_vectors.json are changed to be ones that also have a mnemonic and then the Trezor getxpub test will load with the mnemonic instead of the xprv. Additionally, the Trezor T will be tested now as mnemonics are available.

The new test vectors use mnemonics from https://github.com/trezor/python-mnemonic/blob/master/vectors.json and derivation paths from the current bip32_vectors.json. They were generated using pycoin. Note that original mnemonic test vectors use a BIP 39 passphrase while ours don't.


instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/278/commits/0f1c5b40174d815a4879f20bb1fd6637a8ef61ce

PR #279, opened 2019-11-15
[WIP] Multisig address display

justinmoon:
This PR modifies the existing `displayaddress` command to also display multisig addresses for the 2 devices that I know support it -- Trezor and ColdCard. I don't own a KeepKey so haven't tried that, yet. To my knolwledge, Ledger and BitBox firmwares don't support this feature.

Looking for feedback on the API:
- The ColdCard demands a list of derivation paths for every signer in a redeem script. To support this, I allow `path` to be a comma-separated list of paths -- existing behavior is preserved and only in the ColdCard multisig case is it treated as a comma-separated list. But this might confuse people ...
- I left the `--sh_wpkh` and `--wpkh` flags, which would clearly need to be renamed if they can also represent P2WSH.
-  Added a `redeem_script` parameter to `displayaddress` command and to the `display_address` methods of `TrezorClient` and `ColdcardClient`. I use the presence of this variable to indicate whether we're dealing with a multisig address or not.
- I did not attempt to modify the `desc` parameter to support multisig since it feels a little odd for HWI to be generating redeem scripts (what the HWW libraries need to display a multisig address). Should I add this?

Test coverage is decent:
- 2/2 Native segwit, wrapped segwit, and legacy addresses are checked for Trezor
- 2/4 legacy address checked for ColdCard. As described in [this document](https://github.com/Coldcard/firmware/blob/061c3e5f79cc83f1a1e5af5ca1315d7814bff5f8/unix/README.md#other-startup-flags), the ColdCard simulator has `-m`,  `--wrap` or `--p2wsh` flags to add legacy, wrapped, or native segwit multisig wallets on the ColdCard. The problem, however, is that each wallet shares the same set of bip32 fingerprints -- so you can only have one of these wallets active at a time (cc @peter-conalgo is this accurate?). This is why I'm only testing legacy addresses and not wrapped / native segwit. Perhaps my unittest could restart the simulator once for each flag? Another option would be for HWI unittests to generate and upload our own multisig wallets to the simulator.

### DIFF
--- a/hwilib/devices/ckcc/protocol.py
+++ b/hwilib/devices/ckcc/protocol.py
@@ -110,6 +110,28 @@ class CCProtocolPacker:
         return pack('<4sI', b'show', addr_fmt) + subpath.encode('ascii')
 
     @staticmethod
+    def show_p2sh_address(M, xfp_paths, witdeem_script, addr_fmt=AF_P2SH):
+        # For multisig (aka) P2SH cases, you will need all the info required to build
+        # the redeem script, and the Coldcard must already have been enrolled
+        # into the wallet.
+        # - redeem script must be provided
+        # - full subkey paths for each involved key is required in a list of lists of ints, where
+        #   is a XFP and derivation path, like in BIP174
+        # - the order of xfp_paths must match the order of pubkeys in
+        #   redeem script (after BIP67 sort). This allows for dup xfp values.
+        assert addr_fmt & AFC_SCRIPT
+        assert 30 <= len(witdeem_script) <= 520
+
+        rv = pack('<4sIBBH', b'p2sh', addr_fmt, M, len(xfp_paths), len(witdeem_script))
+        rv += witdeem_script
+
+        for xfp_path in xfp_paths:
+            ln = len(xfp_path)
+            rv += pack('<B%dI' % ln, ln, *xfp_path)
+
+        return rv
+
+    @staticmethod
     def sim_keypress(key):
         # Simulator ONLY: pretend a key is pressed
         return b'XKEY' + key
--- a/hwilib/devices/ckcc/utils.py
+++ b/hwilib/devices/ckcc/utils.py
@@ -85,4 +85,24 @@ def get_pubkey_string(b):
         y = p - y
     return x.to_bytes(32, byteorder="big") + y.to_bytes(32, byteorder="big")
 
+def str_to_int_path(xfp, path):
+    # convert text  m/34'/33/44 into BIP174 binary compat format
+    # - include hex for fingerprint (m) as first arg
+
+    rv = [struct.unpack('<I', binascii.a2b_hex(xfp))[0]]
+    for i in path.split('/'):
+        if i == 'm': continue
+        if not i: continue      # trailing or duplicated slashes
+
+        if i[-1] in "'phHP":
+            assert len(i) >= 2, i
+            here = int(i[:-1]) | 0x80000000
+        else:
+            here = int(i)
+            assert 0 <= here < 0x80000000, here
+
+        rv.append(here)
+
+    return rv
+
 # EOF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -1,11 +1,11 @@
 # Coldcard interaction script
 
-from binascii import b2a_hex
 from ..hwwclient import HardwareWalletClient
 from ..errors import ActionCanceledError, BadArgumentError, DeviceBusyError, DeviceFailureError, UnavailableActionError, common_err_msgs, handle_errors
 from .ckcc.client import ColdcardDevice, COINKITE_VID, CKCC_PID
 from .ckcc.protocol import CCProtocolPacker, CCBusyError, CCProtoError, CCUserRefused
-from .ckcc.constants import MAX_BLK_LEN, AF_P2WPKH, AF_CLASSIC, AF_P2WPKH_P2SH
+from .ckcc.constants import MAX_BLK_LEN, AF_P2WPKH, AF_CLASSIC, AF_P2WPKH_P2SH, AF_P2WSH, AF_P2SH, AF_P2WSH_P2SH
+from .ckcc.utils import str_to_int_path
 from ..base58 import xpub_main_2_test
 from hashlib import sha256
 
@@ -15,7 +15,9 @@ import io
 import sys
 import time
 import struct
-from binascii import hexlify
+from binascii import hexlify, a2b_hex, b2a_hex
+
+py_enumerate = enumerate  # Need to use the enumerate built-in but there's another function already named that
 
 CC_SIMULATOR_SOCK = '/tmp/ckcc-simulator.sock'
 # Using the simulator: https://github.com/Coldcard/firmware/blob/master/unix/README.md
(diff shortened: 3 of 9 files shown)


hwilib/devices/ckcc/protocol.py
```diff
@@ -109,6 +109,28 @@ def show_address(subpath, addr_fmt=AF_CLASSIC):
         # shows on screen, no feedback from user expected
         return pack('<4sI', b'show', addr_fmt) + subpath.encode('ascii')
 
+    @staticmethod
+    def show_p2sh_address(M, xfp_paths, witdeem_script, addr_fmt=AF_P2SH):
```
justinmoon: Copied from [here](https://github.com/Coldcard/ckcc-protocol/blob/master/ckcc/protocol.py#L134)

hwilib/devices/ckcc/utils.py
```diff
@@ -85,4 +85,24 @@ def get_pubkey_string(b):
         y = p - y
     return x.to_bytes(32, byteorder="big") + y.to_bytes(32, byteorder="big")
 
+def str_to_int_path(xfp, path):
```
justinmoon: Copied from [here](https://github.com/Coldcard/ckcc-protocol/blob/master/ckcc/cli.py#L583)

hwilib/devices/coldcard.py
```diff
@@ -15,7 +15,9 @@
 import sys
 import time
 import struct
-from binascii import hexlify
+from binascii import hexlify, a2b_hex, b2a_hex
+
+py_enumerate = enumerate  # Need to use the enumerate built-in but there's another function already named that
```
justinmoon: This was done elsewhere in the project so I did the same hack

hwilib/devices/coldcard.py
```diff
@@ -147,18 +149,43 @@ def sign_message(self, message, keypath):
...
-        keypath = keypath.replace('h', '\'')
-        keypath = keypath.replace('H', '\'')
-
-        if p2sh_p2wpkh:
-            format = AF_P2WPKH_P2SH
-        elif bech32:
-            format = AF_P2WPKH
+    def display_address(self, keypath, p2sh_p2wpkh, bech32, redeem_script=''):
+        # redeem_script means p2sh/multisig
+        if redeem_script:
```
justinmoon: Big if-statement based on `redeem_script` is ugly ...

New code below adapted from [here](https://github.com/Coldcard/ckcc-protocol/blob/master/ckcc/cli.py#L610)

hwilib/hwwclient.py
```diff
@@ -33,6 +33,11 @@ def sign_message(self, message, keypath):
         raise NotImplementedError('The HardwareWalletClient base class does not '
                                   'implement this method')
 
+    # Display address on device display
```
justinmoon: This abstract method was missing so figured I'd add it

test/test_coldcard.py
```diff
@@ -13,7 +13,7 @@
 
 def coldcard_test_suite(simulator, rpc, userpass, interface):
     # Start the Coldcard simulator
-    coldcard_proc = subprocess.Popen(['python3', os.path.basename(simulator)], cwd=os.path.dirname(simulator), stdout=subprocess.DEVNULL, preexec_fn=os.setsid)
+    coldcard_proc = subprocess.Popen(['python3', os.path.basename(simulator), '-m'], cwd=os.path.dirname(simulator), stdout=subprocess.DEVNULL, preexec_fn=os.setsid)
```
justinmoon: Added the `-m` flag discussed in my PR description to pre-seed the simulator with a 2/4 legacy multisig wallet

test/test_device.py
```diff
@@ -87,7 +87,7 @@ def do_command(self, args):
         for arg in args:
             cli_args.append(shlex.quote(arg))
         if self.interface == 'cli':
-            proc = subprocess.Popen(['hwi ' + ' '.join(cli_args)], stdout=subprocess.PIPE, shell=True)
+            proc = subprocess.Popen(['../hwi.py ' + ' '.join(cli_args)], stdout=subprocess.PIPE, shell=True)
```
justinmoon: `pip install -e .[tests]` wasn't working on my system and this was the only way I could run the unittests. Should I remove this?

test/test_device.py
```diff
@@ -535,6 +535,35 @@ def test_display_address_descriptor(self):
...
+            self.assertEqual(wrapped_address, result['address'])
+
+            # legacy
+            legacy_address = '3EYGX7a12tazeFZU2TkBauidkr2T3vx1Rw'
+            result = self.do_command(self.dev_args + ['displayaddress', '--path', derivation_path, '--redeem_script', redeem_script])
+            self.assertEqual(legacy_address, result['address'])
+        elif 'coldcard' in self.full_type:
+            redeem_script = '522102292134f1de237a68c206d3d45f2f412e5d22d45c264472329359404f3112baa221023f36d1f7e51102947f9b56928beacf9f9a6a1f63823b41c2a6aed9f63f61e1da21027fc1b3d175943945a5b0a84c14ef5e4ffa78a3b738a5b0c37b076c63c1fdb04d2102e6640847a5acd85f5252bfd4e800adb95718fdeb734cceb33a80d42095618e5554ae'
+
+            # legacy
```
justinmoon: ColdCard only tests legacy case because I haven't managed to upload wrapped or native segwit wallets to the simulator

test/test_device.py
```diff
@@ -535,6 +535,35 @@ def test_display_address_descriptor(self):
         self.assertIn('code', result)
         self.assertEqual(result['code'], -7)
 
+    def test_display_address_multisig(self):
+        if 'trezor' in self.full_type:
```
justinmoon: If we could upload, say, m/5 multisig wallet to the coldcard containing xpubs from each simulator then I wouldn't need this if-statement.

test/test_device.py
Sjors: Maybe make that a seperate commit?

hwilib/hwwclient.py
Sjors: Shouldn't this also have a `redeem_script` argument? The other wallets that support `displayaddress` should then throw an error if that argument is present.

instagibbs:
concept ACK

instagibbs:
you should add the capability to that capability table in the readme, filling out the appropriate values for each device.

test/test_device.py
achow101: Did you try using poetry to install the it and use its virtualenv?

hwilib/hwwclient.py
achow101: Agreed that this should also have `redeem_script` and every other `HWWClient` should as well, just throwing an error if they see it.

achow101:
Concept ACK

Could you break this up into smaller commits? In general, updating the vendor provided libraries (like ckcc-protocol) should be their own commit. Changes to the interface should be their own commit.

instagibbs:
Could this get resurrected? This is a very nice feature to have for devices that support it.

justinmoon:
Yes @instagibbs. I'll try to fix up this PR in next week.

Sjors:
I just tried with a 2-of-3 multisig on a physical Trezor T, native SegWit multisig. I passed the witness script into the `--redeem_script` argument, set the `--wpkh` flag and used `--path "m/48'/1'/0'/2'/0/0"`. The command echos the correct address, but nothing appears on the display. 

End of the log looks like this:

```
INFO:hwilib.devices.trezorlib.client:creating client instance for device: udp:127.0.0.1:21324
DEBUG:hwilib.devices.trezorlib.transport.protocol:sending message: GetFeatures
DEBUG:hwilib.devices.trezorlib.transport.protocol:received message: Features
DEBUG:hwilib.devices.trezorlib.transport.protocol:sending message: GetAddress
DEBUG:hwilib.devices.trezorlib.transport.protocol:received message: ButtonRequest
DEBUG:hwilib.devices.trezorlib.transport.protocol:sending message: ButtonAck
DEBUG:hwilib.devices.trezorlib.transport.protocol:sending message: DebugLinkDecision
DEBUG:hwilib.devices.trezorlib.transport.protocol:received message: Address
{"address": "tb1q...."}
```

justinmoon:
Update: Life got in the way, but I'm still planning to do this.

@Sjors Thanks for the review BTW

PR #280, opened 2019-11-19
Replace Ledger tests with automated tests and run them on Travis

### LINKED ISSUE #277: Integrate a Ledger simulator into CI
May be the ticket: https://github.com/LedgerHQ/speculos

achow101:
This PR adds the necessary build and test scripts to make use of Ledger's [Speculos](https://github.com/LedgerHQ/speculos) emulator. Speculos emulates Ledger Nano S's and will allow us to test that the commands we send to Ledger devices work.

Following the same pattern as our other device tests, Speculos will be cloned to `test/work`, built from source, and then run by the `test_ledger.py` script. The functionality to use a physical device in this test has been removed and the command line args to `test_ledger.py` and `run_tests.py` have been updated to reflect this. They follow the same pattern as other device tests.

As the emulator does not automatically accept and click through user interface prompts, we needed some way to do the user interaction. This required two pieces: 1) a way to read the emulated display, and 2) another thread to send input. To get the display output, I've added a patch that will output the text displayed to a unix socket. This way we can read the display and know which keypresses to send. Then, another thread is started which listens on that unix socket and sends the appropriate keypress responses  the the emulator's keypress server. This way we can automatically respond to all the stuff displayed on screen.

The Travis config has also been updated. Ubuntu Bionic was needed for the emulator to run, so Travis has been updated to use Bionic instead of Xenial. All of the necessary dependencies will be installed.

Lastly, `test_big_tx` on the emulator takes ~15 minutes. This is long enough to cause Travis to timeout. To get around this, we will install the [`travis-wait-enhanced`](https://github.com/crazy-max/travis-wait-enhanced) tool. This will print out an information line every minute so that we don't hit the ["No output was received" timeout](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) but still print the test results as the tests run (Travis's recommended `travis_wait` function unfortunately does not print out the test output until the command finishes, making it really hard to know what is going on). Our Travis runs are starting to take quite a long time now.

Closes #277

### DIFF
--- a/test/setup_environment.sh
+++ b/test/setup_environment.sh
@@ -162,6 +162,38 @@ pipenv run make -j$(nproc) kkemu
 find . -name "emulator.img" -exec rm {} \;
 cd ..
 
+# Clone ledger simulator Speculos if it doesn't exist, or update it if it does
+speculos_setup_needed=false
+if [ ! -d "speculos" ]; then
+    git clone --recursive https://github.com/LedgerHQ/speculos.git
+    cd speculos
+    speculos_setup_needed=true
+else
+    cd speculos
+    git fetch
+
+    # Determine if we need to pull. From https://stackoverflow.com/a/3278427
+    UPSTREAM=${1:-'@{u}'}
+    LOCAL=$(git rev-parse @)
+    REMOTE=$(git rev-parse "$UPSTREAM")
+    BASE=$(git merge-base @ "$UPSTREAM")
+
+    if [ $LOCAL = $REMOTE ]; then
+        echo "Up-to-date"
+    elif [ $LOCAL = $BASE ]; then
+        git pull
+        speculos_setup_needed=true
+    fi
+fi
+# Apply patch to get screen info
+git am ../../data/speculos-screen-text.patch
+
+# Build the simulator. This is cached, but it is also fast
+mkdir -p build
+cmake -Bbuild -H.
+make -C build/ emu launcher
+cd ..
+
 # Clone bitcoind if it doesn't exist, or update it if it does
 bitcoind_setup_needed=false
 if [ ! -d "bitcoin" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 os: linux
-dist: xenial
+dist: bionic
 python:
     - '3.6.8'
 cache:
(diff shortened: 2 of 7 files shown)


instagibbs:
`ledger: test_backup (test_ledger.ledger_test_suite.<locals>.TestLedgerDisabledCommands) ... [Errno 111] Connection refused`

in a bunch of the test timeouts

concept ACK! Would this allow me to still run manual testing on my own device, since f.e., this doesn't emulate Nano X?

test/setup_environment.sh
```diff
@@ -162,6 +162,38 @@ pipenv run make -j$(nproc) kkemu
 find . -name "emulator.img" -exec rm {} \;
 cd ..
 
+# Clone ledger simulator Speculos if it doesn't exist, or update it if it does
+speculos_setup_needed=false
+if [ ! -d "speculos" ]; then
+    git clone --recursive https://github.com/LedgerHQ/speculos.git
```
instagibbs: we should choose a specific hash to target a commit

achow101: We don't do that for the other simulators...

achow101:
For some reason the transaction signing tests time out on travis. This doesn't happen when I run it locally. Maybe for now we should skip those tests?

instagibbs:
I'm seeing lots of failures, not simple timeouts?

achow101:
The failures were travis-wait-extended timing out, not travis itself timing out.

instagibbs:
Having issues running this(and actually the other emulators) locally. Investigating. 

concept ACK to merging solution provided they can be run locally reliably.

instagibbs:
tests hanging locally:
`ledger: test_backup (test_ledger.ledger_test_suite.<locals>.TestLedgerDisabledCommands) ...`

investigating along with https://github.com/bitcoin-core/HWI/issues/288

instagibbs:
I was having an import error for `DongleServer`. For some reason re-arranging the definition to before `HIDDongleHIDAPI` causes it to not fail to import, and passes tests.

instagibbs:
it's now working without any modification... I don't understand this. but tACK I guess

instagibbs:
tACK e085f99add575f429763875e1c1af2c28312e393

PR #287, opened 2019-12-11
Disable sending the passphrase on Trezor T

### LINKED ISSUE #286: Trezor T: How to know that the user want to enter passphrase on host?
There is no difference in the way Trezor T behave if the user click "enter passphrase on device" or "enter passphrase on host" when prompt by the device.

So software side, how can I know if I should ask the users to write the passphrase on the host?

achow101:
The Trezor T allows users to enter passphrases on the host in the same manner that passphrases are entered for the Trezor One. We shouldn't allow this. So, when the passphrase callback happens, if it's a Trezor T, we raise an exception.

Fixes #286

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -110,6 +110,8 @@ class TrezorClient(HardwareWalletClient):
 
     def _check_unlocked(self):
         self.client.init_device()
+        if self.client.features.model == 'T':
+            self.client.ui.disallow_passphrase()
         if self.client.features.pin_protection and not self.client.features.pin_cached:
             raise DeviceNotReadyError('{} is locked. Unlock by using \'promptpin\' and then \'sendpin\'.'.format(self.type))
 
--- a/hwilib/devices/trezorlib/client.py
+++ b/hwilib/devices/trezorlib/client.py
@@ -135,7 +135,7 @@ class TrezorClient:
         else:
             try:
                 passphrase = self.ui.get_passphrase()
-            except exceptions.Cancelled:
+            except:
                 self.call_raw(messages.Cancel())
                 raise
 
--- a/hwilib/devices/trezorlib/debuglink.py
+++ b/hwilib/devices/trezorlib/debuglink.py
@@ -159,6 +159,7 @@ class DebugUI:
         self.pin = None
         self.passphrase = "sphinx of black quartz, judge my wov"
         self.input_flow = None
+        self.return_passphrase = True
 
     def button_request(self, code):
         if self.input_flow is None:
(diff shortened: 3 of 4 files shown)


NicolasDorier:
Concept ACK for not allowing enter on host.

However, I may miss it, but I don't see how this code fix #286 . If I call getxpub and the user click on "enter on host", it does not seem to me that getxpub will throw an exception with this code.

achow101:
It should return an error if you test it.

When `getxpub` is called, the function within each `HWWClient` that is executed is `get_pubkey_at_path`. In the Trezor [implementation](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezor.py#L119), this calls `btc.get_public_node`](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezorlib/btc.py#L22). This sends `GetPublicKey` message to the device using [`client.call`](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezorlib/client.py#L162). `client.call` has a `while True` loop for handling the responses. There are some responses that require special actions, such as Pin responses and Passphrase responses. If Passphrase on Host is selected, then a `PassphraseRequest` message is the response which will send us into [`_callback_passphrase`](https://github.com/bitcoin-core/HWI/blob/b20ef4297b01c7fa247f551e5b0109dcd9017d93/hwilib/devices/trezorlib/client.py#L132). This then calls `ui.get_passphrase()` which is where this PR has made changes.

When `ui.get_passphrase()` raises and exception, `_callback_passphrase` catches it so that it can send the `Cancel` message. It then raises the same exception. This exception gets passed back up the call stack until it gets to the [`@trezor_exception`](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezor.py#L118) decorator. [This decorator](https://github.com/bitcoin-core/HWI/blob/master/hwilib/devices/trezor.py#L58) will handle the exception and raise the correct exception that makes the correct error message and code.

implementing it this way ensures that anything that requires a passphrase will have an exception thrown for the Trezor T if the Passphrase on Host option is selected.

instagibbs:
I'll test the Trezor 1 passphrase flow and report back(I'm unfamiliar with the expected flow in master so I'll be pedantic).

ormick8:
I'm the real G.O.A.T

NicolasDorier:
@instagibbs this PR is for Trezor T.

@achow101 sorry I will test this soon.

instagibbs:
@NicolasDorier need to test it doesn't break trezor 1 :)

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/287/commits/08d54cc8bffa3a1d0ca2c191bf3712db9618ea37

instagibbs:
tACK https://github.com/bitcoin-core/HWI/pull/287/commits/08d54cc8bffa3a1d0ca2c191bf3712db9618ea37, password on -cli still works for Trezor 1

NicolasDorier:
Because I don't use linux, I need to dockerize all your build scripts. 
This will take me a while, so if nothing break I think you can merge this and if it breaks when I can finally test, I will just complain.

instagibbs:
maybe merge this prior to a release RC?

achow101:
@NicolasDorier Here is a binary you can test with.
[hwi-1.0.3-windows-amd64.zip](https://github.com/bitcoin-core/HWI/files/3979206/hwi-1.0.3-windows-amd64.zip)

NicolasDorier:
tACK works as advertized.

PR #289, opened 2019-12-18
Complete multisig support

### LINKED ISSUE #248: coldcard multisig testing coverage

achow101:
I've tested multisig with all of the devices supported by HWI. This PR fixes the implementations in devices where multisig did not work as well as adds tests for multisig as applicable.

Only two devices had issues with multisig: Trezor T and Coldcard. The Trezor T has some issues because of it's inability to have mixed segwit and non-segwit inputs in a transaction. This poses an issue for us because we skip inputs by declaring the input as being a segwit input and then discarding the signature for it. In this PR, this issue is now documented as a caveat. However this did cause some problems with multisigs since we have to do multiple passes of the transaction, and later rounds would originally mark fully signed inputs as not ours, which, when having legacy multisigs, runs into the mixed segwit issue. This is resolved by not marking those as segwit inputs but still ignoring them.

Additionally, there is now a supports external inputs whitelist in the test to avoid issues with that.

To use multisig with the Coldcard requires enrolling the multisig wallet beforehand. This requirement is now documented. For our tests, we enroll the multisig wallet at startup using the method provided by the simulator's command line arguments. We apply a patch to the simulator to use the multisig wallets that we use in the tests instead of the default provided by the simulator. Additionally, like the Trezor, the Coldcard needs to do multiple passes of signing the transaction if there is a multisig in which the Coldcard is a participant multiple times. So the `sign_tx()` function is updated to handle the multiple passes. Lastly, the Coldcard uses and enforces BIP 67. So we need to use the `sortedmulti` descriptor to handle that.

Both the Coldcard and Trezor T are now enabled for multisig tests.

Closes #248

### DIFF
--- a/docs/trezor.md
+++ b/docs/trezor.md
@@ -20,6 +20,7 @@ Due to the limitations of the Trezor, some transactions cannot be signed by a Tr
 - Multisig inputs are limited to at most n-of-15 multisigs. This is a firmware limitation.
 * Transactions with arbitrary input scripts (scriptPubKey, redeemScript, or witnessScript) and arbitrary output scripts cannot be signed. This is a firmware limitation.
 * Send-to-self transactions will result in no prompt for outputs as all outputs will be detected as change.
+- For **Trezor T**, a transaction cannot contain both segwit and non-segwit inputs
 
 ## Note on `backup`
 
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -6,7 +6,8 @@ from ..errors import ActionCanceledError, BadArgumentError, DeviceBusyError, Dev
 from .ckcc.client import ColdcardDevice, COINKITE_VID, CKCC_PID
 from .ckcc.protocol import CCProtocolPacker, CCBusyError, CCProtoError, CCUserRefused
 from .ckcc.constants import MAX_BLK_LEN, AF_P2WPKH, AF_CLASSIC, AF_P2WPKH_P2SH
-from ..base58 import xpub_main_2_test
+from ..base58 import get_xpub_fingerprint, xpub_main_2_test
+from ..serializations import PSBT
 from hashlib import sha256
 
 import base64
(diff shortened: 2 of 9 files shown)


hwilib/devices/coldcard.py
```diff
@@ -68,53 +69,72 @@ def _get_fingerprint_hex(self):
...
-        left = sz
-        chk = sha256()
-        for pos in range(0, sz, MAX_BLK_LEN):
-            here = fd.read(min(MAX_BLK_LEN, left))
-            if not here:
+        # Get this devices master key fingerprint
+        xpub = self.device.send_recv(CCProtocolPacker.get_xpub('m/0\''), timeout=None)
+        master_fp = get_xpub_fingerprint(xpub)
+
+        # For multisigs, we may need to do multiple passes if we appear in an input multiple times
```
instagibbs: oof, what does this look like for the user? multiple independent signing passes?

instagibbs: also, this needs a test

hwilib/devices/trezor.py
```diff
@@ -220,7 +224,7 @@ def ignore_input():
                 if our_keys > passes:
                     passes = our_keys
 
-                if not found:
+                if not found and not found_in_sigs:
```
instagibbs: can you add some more comments/explanation in the code for this logic? I don't think this is obvious.

docs/trezor.md
```diff
@@ -20,6 +20,7 @@ Due to the limitations of the Trezor, some transactions cannot be signed by a Tr
 - Multisig inputs are limited to at most n-of-15 multisigs. This is a firmware limitation.
 * Transactions with arbitrary input scripts (scriptPubKey, redeemScript, or witnessScript) and arbitrary output scripts cannot be signed. This is a firmware limitation.
 * Send-to-self transactions will result in no prompt for outputs as all outputs will be detected as change.
+- For **Trezor T**, a transaction cannot contain both segwit and non-segwit inputs
```
instagibbs: what happens if it does?

test/test_device.py
```diff
@@ -361,9 +361,9 @@ def _test_signtx(self, input_type, multisig):
                    pkh_info['desc'][4:-10]]
 
         # Get the descriptors with their checksums
-        sh_multi_desc = self.wrpc.getdescriptorinfo('sh(multi(2,' + pubkeys[0] + ',' + pubkeys[1] + ',' + pubkeys[2] + '))')['descriptor']
-        sh_wsh_multi_desc = self.wrpc.getdescriptorinfo('sh(wsh(multi(2,' + pubkeys[0] + ',' + pubkeys[1] + ',' + pubkeys[2] + ')))')['descriptor']
-        wsh_multi_desc = self.wrpc.getdescriptorinfo('wsh(multi(2,' + pubkeys[2] + ',' + pubkeys[1] + ',' + pubkeys[0] + '))')['descriptor']
+        sh_multi_desc = self.wrpc.getdescriptorinfo('sh(sortedmulti(2,' + pubkeys[0] + ',' + pubkeys[1] + ',' + pubkeys[2] + '))')['descriptor']
```
instagibbs: note that this functionality is only in Core master branch for now, so support is a little shaky until 0.20. This should probably go in notes somewhere that this is required for use with Core.

hwilib/devices/coldcard.py
achow101: It is multiple independent signings. There already are tests for this, our multisig tests use 3 keys from the same device. This was needed to pass our test suite.

docs/trezor.md
achow101: You get an error that says `Some signatures are missing`

hwilib/devices/coldcard.py
instagibbs: ah i see github was confusing me with commit ordering, looks good

hwilib/devices/trezor.py
achow101: Added some comments.

test/test_device.py
achow101: Added a note in the coldcard signtx caveats.

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/289/commits/fc0b01f70a03c63009d140d372a6dfef74b389be

achow101:
Is anyone able to replicate the travis failures? I can't get those tests to fail locally.

achow101:
Finally figured out what was causing the travis error.

instagibbs:
re-utACK https://github.com/bitcoin-core/HWI/pull/289/commits/c9107b03c9da1b2a391c9ac674400247071c6caf

PR #290, opened 2019-12-18
tests, coldcard: cleanup backup files

achow101:
The Coldcard backup test produces a backup file which just sits in the test directory. Remove it after the test so it stops polluting the directory.


instagibbs:
ledger signmessage test timed out, kicked

PR #291, opened 2019-12-19
Implement a simple Qt GUI

### LINKED ISSUE #263: [feature] A gui interface
In light of https://github.com/bitcoin/bitcoin/pull/16944, it would probably be useful to have a GUI interface that people can use.

achow101:
Closes #263

Adds a simple Qt GUI which reflects the CLI. Qt for Python (PySides2) is being used. It has been added as a dependency.

The UI is designed using QtCrerator and makes use of `.ui` files. These `.ui` files need to be converted into importable python modules. This can be done by using the new `contrib/generate-ui.sh` script.

Future work:
* Do Core things automatically
* Allow send pin to be used externally by GUI programs so there is a built in way to to the pin stuff

Screenshots:
image

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-exclude = *.pyc,__pycache__,hwilib/devices/btchip/,hwilib/devices/ckcc/,hwilib/devices/trezorlib/,test/work/
+exclude = *.pyc,__pycache__,hwilib/devices/btchip/,hwilib/devices/ckcc/,hwilib/devices/trezorlib/,test/work/,hwilib/ui
 ignore = E261,E302,E305,E501,E722,W5
 per-file-ignores = setup.py:E122
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ test/emulator.img
 test/work
 pip-wheel-metadata
 .mypy_cache/
+
+# Qt stuff
+hwiqt.pyproject.user
+hwilib/ui/ui_*.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,7 +112,6 @@ jobs:
         -   name: macOS binary distribution (no tests)
             stage: test
             os: osx
-            osx_image: xcode7.3
             language: generic
             addons:
                 artifacts:
--- a/contrib/build_bin.sh
+++ b/contrib/build_bin.sh
@@ -11,9 +11,9 @@ poetry install
 
 # We now need to remove debugging symbols and build id from the hidapi SO file
 so_dir=`dirname $(dirname $(poetry run which python))`/lib/python3.6/site-packages
-find ${so_dir} -name '*.so' -type f -execdir strip '{}' \;
+strip ${so_dir}/hid*.so
 if [[ $OSTYPE != *"darwin"* ]]; then
-    find ${so_dir} -name '*.so' -type f -execdir strip -R .note.gnu.build-id '{}' \;
+    strip -R .note.gnu.build-id ${so_dir}/hid*.so
 fi
 
 # We also need to change the timestamps of all of the base library files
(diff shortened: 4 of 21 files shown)


instagibbs:
getting

```
./hwi-qt.py 
  File "./hwi-qt.py", line 11, in <module>
    from ui.ui_getxpubdialog import Ui_GetXpubDialog
  File "/home/instagibbs/HWI/ui/ui_getxpubdialog.py", line 47
    self.buttonBox.raise()

```

on `Python 3.6.9`

achow101:
What are the contents of ui_getxpubdialog.py? (it's generated by the script)

instagibbs:
https://gist.github.com/instagibbs/ba62cecb0e5a3ccf3400ee19ed6d881a

achow101:
Looks like there's a bug in pysides2/Qt 5.14.0, so I guess just use Qt 5.13.2.

instagibbs:
`pip3 install pyside2==5.13.2` fixed it for me

instagibbs:
seems to actually mostly work :) nice. testnet option would be dandy, at some point we also need to support regtest bech32.

instagibbs:
note: looks like you're filtering the path text to always make sure the next keystroke is a valid path. Makes it a bit random/difficult to for example replace a single number with another number in the middle of two `/`. Still, nice to know whatever is there is valid, and not sure I have a valid counter-proposal.

achow101:
Had to change the PySide2 version to 5.14.0. Also reorganized some stuff and did a few bug fixes.

instagibbs:
stupid question but is there a way to install the qt instance using pip?

achow101:
When you install HWI using pip, the `hwi-qt` command should become available too. At least now it should, I forgot to update the setup.py earlier.

instagibbs:
there we go, thanks!

achow101:
Added another check to disable things for uninitialized devices. Setup, wiping, and restoring will be added in a followup PR.

instagibbs:
Could this get some README installation directions? I'm going to start putting this through serious paces since I think having the feature merged earlier is better.

instagibbs:
so I don't forget, after installation using pip I get

```
$hwi-qt
raceback (most recent call last):
  File "/home/instagibbs/elements-dev/HWI/hwi_env_test/bin/hwi-qt", line 11, in <module>
    load_entry_point('hwi==1.0.3', 'console_scripts', 'hwi-qt')()
  File "/home/instagibbs/elements-dev/HWI/hwi_env_test/lib/python3.6/site-packages/pkg_resources/__init__.py", line 490, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/instagibbs/elements-dev/HWI/hwi_env_test/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    raise ImportError("Entry point %r not found" % ((group, name),))
ImportError: Entry point ('console_scripts', 'hwi-qt') not found
```

instagibbs:
1) set passphrase button is active even with devices that don't support it
2) Looks like the keypool field is returning "python" `True/False` rather than `true/false` which stops it from being useful unless users go change the capitalization.

achow101:
I've added a commit which adds the argument parser for command line args. Another commit allows for an initial passphrase with the cli arg `--password` and for testnet mode with `--testnet`.

Also fixed the json thing.

instagibbs:
tACK https://github.com/bitcoin-core/HWI/pull/291/commits/0982c1e29e5e13a8acaa4afe47245009b33ad743

Everything seems to work as expected. It handled device errors just fine with legible errors reported to the user. `--testnet` works as expected I also quickly checked the code changes itself to make sure it doesn't obviously break or confuse anything pre-existing :)

PR #293, opened 2020-01-03
Add an expert mode and display decoded xpub information

achow101:
Adds an `--expert` switch to enable an expert mode. This mode can be used to hide potentially dangerous or otherwise expertise needed behavior. It can also hide output that isn't useful to most users, but may be useful to an expert.

For now, the only thing hidden behind this expert mode switch is newly added decoded xpub information. In `getxpub` and `getmasterxpub`, when `--expert` is used, the xpub will be decoded and that decoded information added to the returned dict. This includes which network the xpub is for, its depth, parent fingerprint, child number, chaincode, and the pubkey itself. This is being hidden behind `--expert` because it isn't generally useful to users, but maybe useful to experts who may need xpubs decoded, e.g. to get just the pubkey.

### DIFF
--- a/hwilib/base58.py
+++ b/hwilib/base58.py
@@ -8,11 +8,17 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 #
 
-from .serializations import hash256
+import hashlib
 import struct
 from binascii import hexlify, unhexlify
 from typing import List
 b58_digits: str = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
+
+def sha256(s):
+    return hashlib.new('sha256', s).digest()
+
+def hash256(s):
+    return sha256(sha256(s))
 
 def encode(b: bytes) -> str:
     """Encode bytes to a base58-encoded string"""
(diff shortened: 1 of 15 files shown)


gwillen:
Thanks, this is really helpful to have. Reviewed-but-not-tested ACK. NOTE: I did not double check the prefix byte strings at the top of ExtendedKey (but I assume you tested all four.)

It seems like there's some refactoring that could be done to reduce code duplication, but I leave that to your sense of taste.

achow101:
Had to rebase.

gwillen:
utACK fbecda2.

instagibbs:
is it possible to get a test for this? :)

achow101:
Added a test

gwillen:
utACK 32f0430

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/293/commits/32f043086454b516a7aa74839cabd2aefd51262b, will merge when travis passes

achow101:
Well travis didn't like that. I think I fixed it.

gwillen:
Is there any way for me to see either of (1) what changed in the force-push, or (2) what the travis error was from the previous version? I assumed I could do (1) somehow, but it looks like github is extremely uncooperative, unless there's a trick I am missing. And for (2) I figured it would show up in the Travis history but I don't see it there.

gwillen:
utACK bc05212, looks like a couple of your travis jobs got shot for timeout though.

Also, I wonder if there is a less error-prone way to test this... although I can't think of anything that would easily catch those issues without having the full suite of simulators set up, which I assume you don't locally?

PR #294, opened 2020-01-03
Add dependencies for Centos

jhfrontz:
Added some maybe-less-than-obvious dependencies for Centos (which may also apply for RHEL and FC).


instagibbs:
looks like you need to rebase you might have checked out an older commit

instagibbs:
utACK otherwise

achow101:
ACK 562e736c65a15fd50bb4a7dc4d3679d092671ee8

PR #295, opened 2020-01-03
List features

### LINKED ISSUE #273: [Feature] Providing device support information via CLI
The information on https://github.com/bitcoin-core/HWI#device-support should be hard coded and able to be consumed via the HWI CLI so we can properly activate/deactivate feature in the GUI based on the detected device.

Because right now, I am scrapping github README page to have that in my middleware.

achow101:
Adds a `getfeatures` command to retrieve the feature support for a given device. The result of this should match the table in the README. To handle the 3 levels of support (supported, not supported by HWI, and not supported by device firmware), a `DeviceFeature` enum is used. We can add more possible levels of support to this in the future. A `SupportedFeatures` class is also added to handle the support listing for a device so that we are sure that no feature is not listed for a given device.

Because some device models that have the same type have different feature support, additional changes are made to have model specific `HWWClient` classes which will have their own feature listings. This will also make it easier in the future to do different things for different models if/when they diverge. In particular, this is used for Trezor 1 and Trezor T as Trezor T has some firmware bugs preventing it from having the same feature set as the Trezor 1.

Closes #273

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -5,6 +5,7 @@ from .commands import (
     displayaddress,
     enumerate,
     find_device,
+    get_client_class,
     get_client,
     getmasterxpub,
     getxpub,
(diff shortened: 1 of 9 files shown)


instagibbs:
needs rebase

PR #297, opened 2020-01-07
coldcard: Fix needs_pin_sent and needs_passphrase_sent and consolidate simulator enumeration

achow101:
`needs_pin_sent` was missing. `needs_passphrase_sent` was misnamed as `needs_passphrase`.

A test has been added to ensure that all enumerate fields are present in the enumerate result. However, because Coldcard and Ledger simulators had separate enumeration logic for simulators, additional changes were made to merge those code paths with the normal enumeration logic. This is achieved by adding the simulator to the list of devices enumerated, and just ignoring it if an error occurs while trying to connect to it so it is ignored if it isn't running.

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -246,7 +246,8 @@ def enumerate(password=''):
         d_data['type'] = 'coldcard'
         d_data['model'] = 'coldcard'
         d_data['path'] = path
-        d_data['needs_passphrase'] = False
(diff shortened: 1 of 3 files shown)


instagibbs:
light code review and tACK on Ledger https://github.com/bitcoin-core/HWI/pull/297/commits/fd21d404f411247592b3b262f173e147683a32ff

Test coverage is good.

PR #298, opened 2020-01-07
travis: remove pyqt5 dependency

achow101:
We don't need it and it's causing travis to fail.


achow101:
This will be merged once travis passes

instagibbs:
ACK

PR #300, opened 2020-01-16
Implement updatefirmware command and firmwaredl binary

### LINKED ISSUE #112: feature request: updatefirmware command
It would be very nice to not be forced to use the centralized wallet portals to update, so putting the various python libraries to use to do the updates instead under one roof is desired.

We would likely have to put caveats on these commands since they'll likely "phone home" in some respect.

Ledger update would also have to include steps to update the bitcoin app itself(how?)

achow101:
A new command, `updatefirmware` is added which takes a firmware file and loads it onto the device. For some devices, verification of the firmware hashes and signatures is done. However, AFAICT, all devices will also verify signature themselves before applying the firmware. Adding this command requires a few new dependencies and the requisite dependencies and libraries have been added. Additionally, a full dependency bump was done.

A new binary, `firmwaredl` has been added to download the firmware for a particular device model. The `-f` option can be used to specify a particular version to download. A `--bitcoinonly` option allows you to download the Bitcoin only firmware if one is available. This is a separate binary as it requires network access and having the dependencies for that in `hwi` itself is not desirable. So it is separate. For the most part, the current bootloader and firmware versions are not needed to know what firmware is downloaded. There are some devices where sometimes an upgrade to a particular version is needed before upgrading to later versions. For now, users will need to figure that out themselves. In the future, we can modify `firmwaredl` to also connect to a device and figure out what firmware versions can be installed based on that.

NOTE: Ledger devices are not supported in this PR. You cannot download their firmware or apply firmware to their devices, and I am not sure whether that will be possible to do. Further information and discussion about Ledger firmware updating can be found in #112

Closes #112


justinmoon:
I was able to successfully update firmware for ColdCard mk2, Trezor One, and BitBox1 using both `firmwaredl.py` on my Arch Linux machine. 

I also built and tested binaries for MacOS, Windows and Linux. I also tried various combinations of `--firmware-version` and `--bitcoinonly`.  

They all seemed to work.

Bugs:
- Calling `poetry run python hwi.py --fingerprint <fingerprint> updatefirmware trezor-1.8.3.bin` while my Trezor One was plugged in and in bootloader mode caused HWI to freeze. It froze on [this line](https://github.com/achow101/hardware-wallet-interface/blob/update-firmware/hwilib/cli.py#L239).
- I observe this when calling `hwi enumerate` with a bootloader-mode Trezor One plugged: `[{"type": "trezor", "path": "webusb:003:1:1:4:3", "error": "Could not open client or get fingerprint information: 'NoneType' object has no attribute 'lower'", "code": -13}]`. Seems like a bug.
- I couldn't download ColdCard firmware version 3.0.0 but I could download 3.0.1:
```
$ poetry run python firmwaredl.py coldcard --firmware-version 3.0.0
{"error": "404 Client Error: Not Found for url: https://github.com/Coldcard/firmware/blob/master/releases/2019-10-09T1356-v3.0.0-coldcard.dfu?raw=true", "code": -13}
```

Notes:
- Would caller of `firmwaredl` be expected to verify binary signatures for some firmware binaries? My understanding is that Trezor (and maybe others) does this on the device, but ColdCard (and maybe others) publish GPG signatures and that should be checked.
- You might want to update the table in the README to mention that firmware updates are supported for all devices except Ledger.
- I like how you implemented it with separate "download" and "install" steps. This way we can also let user install firmware via SD card, for example.
- The JSON returned by the `firmwaredl` isn't very ergonomic: `{"success": true, "message": "Trezor 1 firmware version 1.8.3 downloaded to /home/justin/dev/HWI/trezor-1.8.3.bin"}`. If the caller wants to find out where the file was saved, they need to parse the contents of the `message` field. Perhaps the filename could be a separate key in the JSON object?
- `firmwaredl` takes a "model" string as argument. It took me a second to realize this was the same as the `model` key in `enumerate`. Maybe mention this in the help string?
- Can HWI tell us what firmware version a given device is running? Without this we won't know when to prompt user that an upgrade is available.

achow101:
> I couldn't download ColdCard firmware version 3.0.0 but I could download 3.0.1:

It looks like 3.0.0 doesn't exist anymore.

> Would caller of `firmwaredl` be expected to verify binary signatures for some firmware binaries?

Yes. Perhaps we should also download verifiable signatures if possible?

justinmoon:
> Yes. Perhaps we should also download verifiable signatures if possible?

Let's do that

PR #304, opened 2020-01-23
Pin additional dependency versions to make builds reproducible

### LINKED ISSUE #303: Reproducibility issues for historical releases
It seems that I am unable to reproduce the results of existing releases.

As noted by https://github.com/spesmilo/electrum/issues/5859#issuecomment-577121504, part of this is related to Cython which is a build requirement for hidapi. Changing the version of this does result in different builds. However that is not the only thing causing issues. Even when forcing cython to a particular version (by adding `poetry run pip install cython==0.29.13` to `contrib/build_bin.sh`), the result is different from the published builds. So there must be something else that is not version locked that is causing these issues.

achow101:
Adds additional version pinning so that builds are more reproducible. Poetry for the build also has a fixed version number. Cython is now a dependency so that hidapi is built reproducibly.

Fixes #303


achow101:
It seems that you can't install older packages on debian, so pinning the package versions like that isn't really helpful.

We'll probably need to build those libraries too and add them to the pyinstaller stuff.

achow101:
Dropped the package version pinning. I'm working on either excluding them from the result, or deterministically building them ourselves.

PR #306, opened 2020-01-23
Remove several system libraries from PyInstaller build

achow101:
PyInstaller was including a bunch of system libraries that we don't need. Removing them makes the final build smaller and lets us not have to worry about them when it comes to reproducibility. These libraries are not used by HWI but are imported by python built-ins that we aren't using. These all have handling for the libraries not being there so it is fine to exclude them.

The libraries being excluded are:
* libz
* libbz2
* libzma
* libreadline

libz (zlib) has to be handled more specially when python is being built during the docker container setup. The python compilation will include it if it detects it, so we apply an additional patch to CPython for it to not do that.


achow101:
Alternatively, instead of trying to hack CPython and make it not build things we don't use, we could try to find (or make) a python distribution that has a reduced standard lib where we can add modules for the things we use.

PR #307, opened 2020-01-23
Add get fingerprint function to device class

TheCharlatan:
As pre-work for the bitbox02 integration, add a function to the device class to get the master fingerprint. It can later be overloaded by the bitbox02 definitions.

Also removes the `get_xpub_fingerprint_as_id` function, since it does the same as `get_xpub_fingerprint_hex`.

### DIFF
--- a/hwilib/base58.py
+++ b/hwilib/base58.py
@@ -82,11 +82,6 @@ def get_xpub_fingerprint_hex(xpub: str) -> str:
     fingerprint = data[5:9]
     return hexlify(fingerprint).decode()
 
-def get_xpub_fingerprint_as_id(xpub: str) -> str:
-    data = decode(xpub)
-    fingerprint = data[5:9]
-    return hexlify(fingerprint).decode()
-
 def to_address(b: bytes, version: bytes) -> str:
     data = version + b
     checksum = hash256(data)[0:4]
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -6,7 +6,7 @@ import importlib
 import platform
 
 from .serializations import PSBT
-from .base58 import get_xpub_fingerprint_as_id, get_xpub_fingerprint_hex, xpub_to_pub_hex
+from .base58 import xpub_to_pub_hex
 from .errors import UnknownDeviceError, BAD_ARGUMENT, NOT_IMPLEMENTED
 from .descriptor import Descriptor
 from .devices import __all__ as all_devs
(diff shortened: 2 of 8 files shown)


achow101:
Concept ACK.

There's some more cleanup here that can be done. All of the devices (except Coldcard) still do the manual extraction of fingerprint from the xpub. So those can be changed to use this new method. The Coldcard has a method `_get_fingerprint_hex` that can be changed to `get_master_fingerprint`.

There's also the `get_xpub_fingerprint()` function in `base58.py` that returns the fingerprint as an int from an xpub. This is used in the Coldcard, Digital Bitbox, and Trezor. This function is the same trick that is being consolidated here. So I think we could add another method that also does that, maybe instead converting the fingerprint hex to the int instead of getting an xpub again and decoding it.

TheCharlatan:
> There's some more cleanup here that can be done. All of the devices (except Coldcard) still do the manual extraction of fingerprint from the xpub. So those can be changed to use this new method. The Coldcard has a method `_get_fingerprint_hex` that can be changed to `get_master_fingerprint`.

I force-pushed a change to implement this.

> There's also the `get_xpub_fingerprint()` function in `base58.py` that returns the fingerprint as an int from an xpub. This is used in the Coldcard, Digital Bitbox, and Trezor. This function is the same trick that is being consolidated here. So I think we could add another method that also does that, maybe instead converting the fingerprint hex to the int instead of getting an xpub again and decoding it.

If I understood the code correctly the fingerprint is already set in the class variable when the respective devices need the fingerrpint in "int" form, so is it good enough for to just expect it to be set and convert it from hex to "int"? Then we can just cast it whenever it is needed.

achow101:
> If I understood the code correctly the fingerprint is already set in the class variable when the respective devices need the fingerrpint in "int" form, so is it good enough for to just expect it to be set and convert it from hex to "int"? Then we can just cast it whenever it is needed.

I'm not sure that that is a reasonable assumption.

I've also completely missed the fact that we even have a `fingerprint` member variable, in which case, we don't really need to have these `get_master_fingerprint_hex` functions. I think we either should make sure that that is always set (e.g. by doing it in the class constructor) and always use that, or we should get rid of it and use the functions. Either way doesn't bother me, but the fact that we do both does.

TheCharlatan:
I went with the member function route and removed the fingerprint variable from the hardware client class.

achow101:
There's a couple places in `hwilib/commands.py` where `client.fingerprint` is accessed/modified. Could you remove/replace those with `client.get_master_fingerprint_hex()`?

achow101:
@TheCharlatan Could you clean up the remaining `client.fingerprint` uses? I would like to get this merged.

achow101:
ACK 06b5bf07a46b5a014c7507e41a7a9a9ea1d31d8d

PR #308, opened 2020-01-31
Add "active" to getkeypool's output to match importdescriptors

achow101:
importdescriptors uses "active" instead of "keypool" to indicate that new addresses can be fetched from that import object.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -104,6 +104,7 @@ def getkeypool_inner(client, path, start, end, internal=False, keypool=True, acc
     this_import['timestamp'] = 'now'
     this_import['internal'] = internal
     this_import['keypool'] = keypool
+    this_import['active'] = keypool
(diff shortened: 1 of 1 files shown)


achow101:
Merging this as it's a simple change, the `importdescriptors` api probably won't change, and having extra key-value pairs in a JSON object doesn't hurt anything.

instagibbs:
post-hom tACK, just noticed this :)

jb55:
great :+1: needed this today when testing `importdescriptors`. Although now I kind of wish there was a way to dump all relevant descriptor requests with a single command, something like getdescriptors but in importdescriptors/getkeypool format?

achow101:
> dump all relevant descriptor requests with a single command

The `--all` option? #316

jb55:
> The `--all` option? #316 

excellent

PR #311, opened 2020-03-02
Update firmware build scripts and tests to work with new firmwares

achow101:
For the past 2 weeks, travis has been failing because hardware wallet vendors have been updating their firmware (and emulators) in a way that was incompatible with the HWI tests. I've finally gotten around to fixing that.

Things that were updated:
* Keepkey build script. They switched to Python 3 and require a newer version of nanopb. Some other build commands had to be modified.
* Trezor 1 build script. `script/setup` needs to be run in the pipenv.
* Trezor T master branch incompatibility. Trezor T has made some major change that makes HWI incompatible with their current master branch (and probably future firmware versions too). For now, we checkout the 2.2.0 (latest as of this PR) firmware for testing.
* Speculos command line arguments. Speculos changed their command line argument names.
* Speculos patched to just do the key presses internally. Instead of having us to run another thread to send the correct key presses at the right time, we modify speculos to do that itself.
* Coldcard command line arguments. Coldcard changed some command line argument names.
* Some ismine thing is "broken" in Core, added a workaround for that.
* Each devices tests are run with their own test runner. Global emulators will only be run for their device's tests.
* MacOS build will use travis' defaul xcode version so that we don't run into issues with it being outdated and no longer supported by travis


instagibbs:
for posterity: this is the change in Core that broke HWI's tests: https://github.com/bitcoin/bitcoin/pull/18067

achow101:
I've made a large number of other changes to the tests in this PR. Many of these changes are to reduce the chance that a travis job times out.

* Each device is run with its own test runner. This allows us to have global emulators (emulators that are live for an entire device's tests as opposed to a new emulator instance for each test case) that are being run only for that device's tests instead of all the time. This should reduce the load on the travis VMs.
* Ledger test will use a single Speculos instance instead of one per test case. This should reduce the risk of speculos timing out
* The non-device tests will no longer build all the emulators. Those tests don't need the emulators so building them is pointless

Once we get a successful travis run, this will be merged.

PR #312, opened 2020-03-04
Doc: Readme: mention calling --help after command name

harding:
Reading through the tutorials, it was apparent to me that commands could accept parameters that weren't documented in the main `--help` text, but it was unclear to me how I could learn about those parameters without reading the source code; after some digging I figured it out and so this PR adds a short description to the readme about how to use the help.


achow101:
ACK f24eb9e5ab4b233ecbe2b08d5541a913becb6022

PR #313, opened 2020-03-20
tests: Add missing self.seen_msg_hash to speculos patch

achow101:
This was missing and could cause problems in the future. We appear to not run into issues with it due to test order.


instagibbs:
yeah sure why not

ACK https://github.com/bitcoin-core/HWI/pull/313/commits/f719449b4e918bf62865cc5207d81ff68b7c2726

PR #314, opened 2020-03-24
tests: emulator logging and speculos updates

achow101:
Instead of redirecting the output of the emulators to /dev/null, redirect them to log files. If travis fails, print out those log files.

Speculos has pushed a fix for the timeout issues that we were seeing. Something about race conditions. So the Ledger tests should no longer timeout. Our patch to Speculos needs to be updated because of this fix. This also means that we can enable the signtx tests, so those have been enabled for Ledger as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ pip-wheel-metadata
 # Qt stuff
 hwiqt.pyproject.user
 hwilib/ui/ui_*.py
+
+*.stderr
+*.stdout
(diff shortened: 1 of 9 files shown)


instagibbs:
> AssertionError: 'code' unexpectedly found in {'error': 'Exception : Proxy connection failed', 'code': -3}

in ledger big sign test

achow101:
> > AssertionError: 'code' unexpectedly found in {'error': 'Exception : Proxy connection failed', 'code': -3}
> 
> in ledger big sign test

I think the cache needs to be cleared, but I can't do that right now for some reason

achow101:
Fixed the speculos patch again. Looks like travis is working now.

instagibbs:
ACK https://github.com/bitcoin-core/HWI/pull/314/commits/ab86822ea9139c0e8d9a5ede103d1047e47c7b8c

instagibbs:
love that green

Issue #315, opened 2020-04-01
getkeypool: return all BIPXX chains for descriptor wallets


instagibbs:
For Core with descriptor wallets, it's much less dangerous/janky to simply give all 6 BIP44/49/84 chains to the wallet, for compatibility reasons(e.g., auditing/migrating from Ledger Live).

Obviously this is problematic for Core up til master today, so maybe a new option `--all` with proper documentation?

PR #316, opened 2020-04-01
getkeypool: --all option for descriptor wallets

instagibbs:
needs tests, putting this up for concept ACKs first. Seems to work(TM). 

resolves https://github.com/bitcoin-core/HWI/issues/315

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -10,6 +10,12 @@ from .base58 import xpub_to_pub_hex
 from .errors import UnknownDeviceError, BAD_ARGUMENT, NOT_IMPLEMENTED
 from .descriptor import Descriptor
 from .devices import __all__ as all_devs
+from enum import Enum
+
+class AddressType(Enum):
+    PKH = 1
+    WPKH = 2
+    SH_WPKH = 3
 
 # Get the client for the device
 def get_client(device_type, device_path, password='', expert=False):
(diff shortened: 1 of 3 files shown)


achow101:
Concept ACK

hwilib/commands.py
```diff
@@ -108,7 +106,7 @@ def getkeypool_inner(client, path, start, end, internal=False, keypool=True, acc
     this_import['watchonly'] = True
     return [this_import]
 
-def getdescriptor(client, master_xpub, testnet=False, path=None, internal=False, sh_wpkh=False, wpkh=True, account=0, start=None, end=None):
+def getdescriptor(client, master_xpub, testnet=False, path=None, internal=False, addr_type="wpkh", account=0, start=None, end=None):
```
achow101, line 109: Maybe use an enum instead of strings?

instagibbs: yep

instagibbs:
Pushed a basic test, no QT support currently because I can't be bothered.

hwilib/commands.py
```diff
@@ -157,21 +165,40 @@ def getdescriptor(client, master_xpub, testnet=False, path=None, internal=False,
...
+    return Descriptor(master_fpr, path_base.replace('m', ''), client.xpub_cache.get(path_base), path_suffix, client.is_testnet, is_sh_wpkh, is_wpkh)
+
+def getkeypool(client, path, start, end, internal=False, keypool=True, account=0, sh_wpkh=False, wpkh=True, addr_all=False):
 
-# wrapper to allow both internal and external entries when path not given
-def getkeypool(client, path, start, end, internal=False, keypool=True, account=0, sh_wpkh=False, wpkh=True):
+    if sh_wpkh and wpkh:
+        return {'error': 'Both `--wpkh` and `--sh_wpkh` can not be selected at the same time.', 'code': BAD_ARGUMENT}
+    if addr_all and (sh_wpkh or wpkh or (path is not None) or internal):
+        return {'error': 'Both `--all` and specific address/change types can not be selected at the same time.', 'code': BAD_ARGUMENT}
```
achow101, line 175: Instead of having the checks like this and requiring multiple bool args, we should just add a mutually exclusive group to the argument parser itself. That should do what we want and should be able to get it to give us the correct enum too.

instagibbs: Still punting enum list construction until `getkeypool` internally, but otherwise done(I think?)

hwilib/commands.py
```diff
@@ -180,9 +202,9 @@ def getdescriptors(client, account=0):
 
     for internal in [False, True]:
         descriptors = []
-        desc1 = getdescriptor(client, master_fpr=master_fpr, testnet=client.is_testnet, internal=internal, sh_wpkh=False, wpkh=False, account=account)
```
achow101, line 183: `master_xpub` is an unrelated change. You may need to rebase.

instagibbs: should be fixed, was an artifact of rebasing incorrectly onto master

achow101:
ACK 15b213a6dc0e9193fc624973b22a411723cf1974

PR #318, opened 2020-04-12
Support Trezor and KeepKey toggle passphrase protection

ben-kaufman:
Adds `togglepassphrase` command which toggles the passphrase protection functionality for Trezor and KeepKey devices. This will allow to enable and disable passphrase protection without having to wipe the device as is now required when using the HWI tool.

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -242,6 +242,10 @@ class ColdcardClient(HardwareWalletClient):
     def send_pin(self, pin):
         raise UnavailableActionError('The Coldcard does not need a PIN sent from the host')
 
+    # Toggle passphrase
+    def toggle_passphrase(self):
+        raise UnavailableActionError('The Coldcard does not support toggling passphrase from the host')
+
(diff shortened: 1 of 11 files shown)


hwilib/devices/coldcard.py
```diff
@@ -242,6 +242,10 @@ def prompt_pin(self):
     def send_pin(self, pin):
         raise UnavailableActionError('The Coldcard does not need a PIN sent from the host')
 
+    # Toggle passphrase
+    def toggle_passphrase(self):
+        raise NotImplementedError('The Coldcard does not implement this method')
```
achow101, line 247: To be consistent with other commands, this should be `UnavailableActionError` unless it is possible to implement this feature for the other devices.

ben-kaufman: Thanks @achow101, I modified the error for `ColdCard`, `Ledger` and `Digital Bitbox`, and squashed the commits.

achow101:
ACK c8d9db693b721f0ff85fdd05ee0de80e950f0954

PR #319, opened 2020-04-17
Trezors: Update to work with firmware 2.3 and 1.9

achow101:
Trezor firmware versions 2.3 and 1.9 changed a couple of things that were causing HWI to not work with them. This PR fixes those things.

* Changed `promptpin` to use `GetPublicKey` message. The `Ping` message used previously no longer has a pin prompt option.
* Updated messages and passphrase callback to match new passphrase workflow
* Check for a session being initialized by doing a `GetPublicKey` and seeing if a `FirmwareError` is returned. If so, send an `Initialize` to start a session.


prusnak:
> Due to the passphrase and session changes to the Trezor T, users using firmware 2.3 will be required to enter the passphrase on the host in the same way that is done with the Trezor 1

Entering passphrase on the device is definitively possible via python-trezor 0.12, see https://github.com/spesmilo/electrum/pull/6064

matejcik:
i don't understand why the previous code disallowed entering passphrase on host, and the new code enforces passphrase on host. If entering on device *ever* worked, there is no reason it should not work now. If entering on host works for T1, there is no reason it should not work for TT.

matejcik:
I think I understand why you disabled passphrase-on-host for TT. Is it because there is no place to prompt to enter on host / on device?

If that's the case, I'd suggest a different behavior: if passphrase is not specified on the command line, ask for passphrase on device. Otherwise use the supplied passphrase.

(Alternately, you could reuse `promptpin` to invoke the passphrase-on-device dialog on TT.)

I still don't understand why entering on host was briefly disallowed.

achow101:
> Entering passphrase on the device is definitively possible via python-trezor 0.12, see [spesmilo/electrum#6064](https://github.com/spesmilo/electrum/pull/6064)

The session changes mean that a user must enter their passphrase for every single command. Entering it every time on the device is a pain in the ass.

> If that's the case, I'd suggest a different behavior: if passphrase is not specified on the command line, ask for passphrase on device. Otherwise use the supplied passphrase.

That's not a bad idea.

> I still don't understand why entering on host was briefly disallowed.

I prefer that as much sensitive user info be entered on the device itself. So the preference is to have PINs and passphrases entered on the device. Previously the T would let you do that and it would cache the passphrase so users only had to do it once. So entering on the host was disallowed in order to require users enter the passphrase on the device. But with the session changes users would have to enter the passphrase with every single command they do. I found this experience to be cumbersome and annoying so I reversed the requirement: users must now enter on the host (this reversal is partially because we have no way to prompt users for which they want to do).

prusnak:
> The session changes mean that a user must enter their passphrase for every single command. 

Not if you implement sessions like we suggested earlier.

matejcik:
Dne pá 17. 4. 2020 18:31 uživatel Andrew Chow <notifications@github.com>
napsal:

> The session changes mean that a user must enter their passphrase for every
> single command. Entering it every time on the device is a pain in the ass.
>
this has been true for as long as TT exists though. there is no change. I
thought you resolved that by never ever calling Initialize.

the only difference is that now you can't call GetPublicNode on a freshly
plugged-in device, because there is no default session.
once you or something else calls Initialize for the first time (so a
session is created), that session remains open for as long as you don't
call Initialize again.

>

achow101:
> Not if you implement sessions like we suggested earlier.

Perhaps I misunderstand how sessions work. Or perhaps you are misunderstanding how HWI works. I'll try to explain why I don't want to do that.

As I understand it, sessions require the host software to store state. The device returns a 32 byte session id in a `Features` message and this session id must be sent in an `Initialize` message at the beginning of each connection in order to use the passphrase cached for that session. If it is not sent, the device begins a new session which means that the passphrase must be entered again (or a different passphrase entered).

Now the way that HWI works is that it connects to a device, does whatever is needed to complete the action specified by the user, and then disconnects. It must disconnect because it is a command line tool like `trezorctl`. For example. if the user does `getxpub m/0h/0`, then it connects to the device, sends `Initialize`, handles the response, sends a `GetPublicKey` message, handles the passphrase callback, etc. At the end, it gets the xpub, closes the device connection, returns the xpub, and exits. In this scenario, the user had to enter his passphrase. If we forced passphrase on device, he would have entered it on the device. Now suppose that user does `getxpub m/0h/1`. We do the same thing as above and the user has to enter his passphrase again. This is not a good experience.

So to have the good experience of the passphrase being cached on the first `getxpub`, we need to send that session id given in the `Features` message the first time he did `getxpub`. But how do we get that session id from the first invocation of `getxpub` to the second invocation? We can't just "store it in a variable" as you have suggested previously; the program exits entirely so anything that was in memory is lost. We'd have to make HWI store some state on disk. This means that we need to now add an entire framework for storing data on disk. We need to know where the data is going to be, allow the user to change the location, handle file I/O errors, etc. It's a whole lot more work just to store a single 32 byte string.

The crux of the problem is that I (and others) think that HWI shouldn't store anything on disk. It shouldn't need to store any state. No other device requires this. The problem it is trying to solve does not need to store state except for this one specific thing that you have recently added to your firmware. Frankly, I'm tired of having to constantly add Trezor specific things just to keep supporting your devices. This isn't a Trezor utility, it's a general purpose hardware wallet tool and it really should not have device specific commands and options (we already have 3 Trezor specific commands and I'd really like to get rid of them).

So if storing the session id in memory is not possible, and storing it on disk is unwanted, what other options are there? The only thing I can think of is that we make the user remember the session id and enter it on the command line every time. So now the user has to remember a 64 character hex string and remember to pass it in with a Trezor specific command line argument. But this workflow is ostensibly the same thing as making the user give the passphrase on the command line every time. Sure it is not the passphrase itself, but currently the session id is effectively just a proxy for the passphrase. However user passphrases are easier for the user to remember and does not change over time. The user experience between passing in a session id and passing in their passphrase is the same - on device they don't have to enter their passphrase for every command. It just has to be part of the command invocation which is easy to copy and paste.

This is why I've chosen to require the user enter the passphrase on the host. If I have misunderstood how sessions work or if you have any other ideas for how we could actually make use of sessions, please let me know.

achow101:
>> The session changes mean that a user must enter their passphrase for every
>> single command. Entering it every time on the device is a pain in the ass.
>>
> this has been true for as long as TT exists though. there is no change. I
> thought you resolved that by never ever calling Initialize.
> 
> the only difference is that now you can't call GetPublicNode on a freshly
> plugged-in device, because there is no default session.
> once you or something else calls Initialize for the first time (so a
> session is created), that session remains open for as long as you don't
> call Initialize again.

This was not true in my testing. I guess that's a bug.

In my experimentation, every time I did something without sending `Initialize` after opening the connection, I always got a non-specific `FirmwareError`. When I looked into it, this error was because there was no session for that connection.

prusnak:
Your utility is being used essentially as a library for projects like Wasabi, BTCPay and maybe others. If it's not acceptable to pass session_id as a parameter to your utility, then it is not meant to be used as a library and you should communicate this clearly in the readme. I will then point these projects to read that particular line in the readme and recommend them using something else.

prusnak:
> So now the user has to remember a 64 character hex string and remember to pass it in with a Trezor specific command line argument.

I am not sure what's the problem here. If I understand it correctly, a user needs to provide ` -d <path>` on everycall. This is not different to providing `-d <path> -s <session>`. You can even combine these two into one string. I am pretty users copy-paste the path from the enumerate call anyway as they can change on every device connection. From that regard, it is just making the "path" identifier longer.

nopara73:
@molnard @lontivero This conversation does not look good. Could you handle the situation?

hwilib/devices/trezor.py
```diff
@@ -414,7 +414,7 @@ def prompt_pin(self):
             raise DeviceAlreadyUnlockedError('The PIN has already been sent to this device')
         print('Use \'sendpin\' to provide the number positions for the PIN as displayed on your device\'s screen', file=sys.stderr)
         print(PIN_MATRIX_DESCRIPTION, file=sys.stderr)
-        self.client.call_raw(proto.Ping(message=b'ping', button_protection=False, pin_protection=True, passphrase_protection=False))
+        self.client.call_raw(proto.GetPublicKey(address_n=[0x80000000], ecdsa_curve_name=None, show_display=False, coin_name=None, script_type=proto.InputScriptType.SPENDADDRESS))
```
prusnak, line 417: You want this to be a standard path, such as `m/44'/1`/0` (first legacy account on Testnet):

```
address_n=[0x8000002c, 0x80000001, 0x80000000]
```

harding:
@prusnak 

> If I understand it correctly, a user needs to provide -d <path> on every call now.

No.  You can get away with only specifying the master xpub fingerprint, which is unchanging as long as the seed stays the same.  E.g.:

```text
$ ./hwi.py enumerate
[{"type": "trezor", "path": "webusb:002:2", "model": "trezor_t", "needs_pin_sent": false, "needs_passphrase_sent": false, "fingerprint": "01234567"}]

$ ./hwi.py -f 01234567 getdescriptors
{"receive": ["pkh([01234567/44h/0h/0h]xpub[...]

# <physically disconnect device, reconnect, enter pin>

$ ./hwi.py -f 01234567 getdescriptors
{"receive": ["pkh([01234567/44h/0h/0h]xpub[...]
```

Having to run `./hwi.py enumerate` each time I connect the device and then modify any commands I want to reuse from my bash history with a new sessionid sounds annoying; it'd be nice if that could be avoided.

nopara73:
@prusnak Would it be possible to implement backwards compatibility with a new firmware?

It would be really uncomfortable for both of us getting numerous support requests regarding it :/

> If a change results in user programs breaking, it's a bug in the kernel. - Linus

instagibbs:
@harding I typically do the poor-man's version as well: `-t <device type>` which of course only works if there is one of each type plugged in

prusnak:
> which is unchanging as long as the seed stays the same

You ignore the passphrase completely in the scenario you are describing. The whole discussion here is because of the passphrase.

> Would it be possible to implement backwards compatibility with a new firmware?

We are backward compatible. If your client doesn't understand the new API it will fall back to entering the passphrase on the host. What we are discussing here is to bring a new feature: a user can enter the passphrase on the device and they do not have to enter it again for future invocations of HWI if they provide the correct session_id.

harding:
@prusnak 

> You ignore the passphrase completely in the scenario you are describing. The whole discussion here is because of the passphrase.

That's fair, I've never used a passphrase with a Trezor.  However, it seems likely to me given @instagibbs's [comment](https://github.com/bitcoin-core/HWI/pull/319#issuecomment-615387760) that `-d <device_id>` doesn't need to be specified for every command even if a passphrase is in use.  (I don't have a spare Trezor to use to test that assertion, though.)

hwilib/devices/trezor.py

achow101:
Following up on what @matejcik said, I have been able to restore the previous behavior of having the passphrase entered on device once. After some further testing, the non-specific `FirmwareError` was only occurring if an `Initialize` had never been sent. The most recent changes are to do a `GetPublicKey` to see if it fails, and if it does, then an `Initialize` will be sent. So we are able to keep the current session open. I no longer was able to reproduce that non-specific `FirmwareError` on new connections to the device, so I will just call that either a failure in testing methodology, or a bug in a previous firmware revision.

What users experience with these changes should be identical to what they experienced previously.

> You ignore the passphrase completely in the scenario you are describing.

I don't think he does. The scenario he describes with using the fingerprint is a common use case. The fingerprint is dependent on the passphrase as it is derived from the key at `m/0h`. With sessions, the user will always have to provide the session id in order to get the same fingerprint.

> We are backward compatible.

While the bulk of this discussion has been about the passphrases, there are other issues this PR fixes which are due to the API not being backwards compatible. Namely the changes to `Ping` are not backwards compatible as we were expecting the `pin_protection` field to still be in `Ping` and have some effect. But that is no longer the case.

lontivero:
@prusnak as you mentioned above, HWI is being used by a couple of projects like Wasabi, BTCPayServer and maybe others too. I know that many of our users love to be able to use their Trezor wallets with Wasabi and the same has to be the case with other projects too.

Two different Wasabi users reported yesterday that they cannot use Trezor with Wasabi after upgrading the firmware and I don't know how many other users could have been affected. Anyway, this is one of those situations where all our projects lose so, could it be possible for Trezor to release a new firmware that makes the interface behaves as before?

prusnak:
> Two different Wasabi users reported yesterday that they cannot use Trezor with Wasabi

Unusable Trezor with Wasabi will be fixed when this PR is merged. 

> could it be possible for Trezor to release a new firmware that makes the interface behaves as before?

No, sorry. Take your time to understand the discussed issue and you'll see there is no regression. And with the change we introduced, the Wasabi UX will be much better for the user. (As they won't have to enter the passphrase again even when switching between different wallets).

molnard:
Thank you for the cooperation here @prusnak @achow101 @matejcik and @harding ! 

I agree that we should focus on long-term goals. With all respect, HWI and projects built on the top should be taken into account when checking the compatibility of the upcoming firmware. This issue struct out of the blue and surprised us, more importantly, the users. That is the worst thing can happen on any term. 
It seems we can fix this current issue. Also, it attracts the attention that these kinds of changes cannot be done one-sidedly. Preventing cooperation is required.

@achow101 elegant solution! Let me know if I can do some testings on it. Wasabi will publish a new hotfix release with the fix, hopefully soonish.

achow101:
Here are builds of this PR for testing
[hwi-mac.tar.gz](https://github.com/bitcoin-core/HWI/files/4496810/hwi-mac.tar.gz)
[hwi-linux.tar.gz](https://github.com/bitcoin-core/HWI/files/4496812/hwi-linux.tar.gz)
[hwi-win.tar.gz](https://github.com/bitcoin-core/HWI/files/4496813/hwi-win.tar.gz)

NicolasDorier:
This version works on old firmware of Trezor T, this is good enough ACK for me.

Testing the new version of firmware now

NicolasDorier:
Updated my Trezor, and it still work. Thank @achow101 !

molnard:
Tested with numerous HW wallets and firmwares. Works well!

ben-kaufman:
Tested with the binary on macOS but got `Could not open client or get fingerprint information: FirmwareError: Firmware error", "code": -13`.

However, when I tested (on macOS) from source it worked well for me, both models One and T, both older and the new firmware and both with and without passphrases.

@molnard just to be sure, was the passphrase enabled when you plugged in the device?
For me it does trigger the passphrase input.

molnard:
I tested with the binaries only (updated according to the edited comment https://github.com/bitcoin-core/HWI/pull/319#issuecomment-615884187)
@ben-kaufman yes the same device is working on windows, there I can enter my passphrase during the enumerate phase. 
I have another Trezor_T without the passphrase. It works fine on both OS.

ben-kaufman:
@molnard Yes, I get the same error with the binary, but not when using from the source.



> Here are builds of this PR for testing
> [hwi-mac.tar.gz](https://github.com/bitcoin-core/HWI/files/4496822/hwi-mac.tar.gz)
> [hwi-linux.tar.gz](https://github.com/bitcoin-core/HWI/files/4496812/hwi-linux.tar.gz)
> [hwi-win.tar.gz](https://github.com/bitcoin-core/HWI/files/4496813/hwi-win.tar.gz)
> 
> edit: wrong mac build.

@achow101 are you sure the new macOS build linked is correct?

achow101:
> @achow101 are you sure the new macOS build linked is correct?

Nope. I don't have a mac so I can't test it out myself. The macOS builds come from travis, so I probably grabbed the wrong one. But now I can't find the build from this PR...

If the source works, then it should be fine. It looks like there are also some other issues with macOS builds currently.

molnard:
No problem, it is working - good job, thx!

molnard:
On Linux I am getting the following error after running our unit tests on HWI:

```
Error Message:
   WalletWasabi.Hwi.Exceptions.HwiException : 'hwi --testnet enumerate' exited with incorrect exit code: 255.
```

Zero is the desired. Did you change the exit code?

molnard:
Installed an Ubuntu 18.04 and dig deeper. With the new release it is not working.

```
[1810] Error loading Python lib '/tmp/_MEInzgskp/libpython3.6m.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/_MEInzgskp/libpython3.6m.so.1.0)
2020-04-21 15:52:43 ERROR	ConnectHardwareWalletViewModel (140)	WalletWasabi.Hwi.Exceptions.HwiException: 'hwi enumerate' exited with incorrect exit code: 255.
   at WalletWasabi.Hwi.HwiClient.SendCommandAsync(IEnumerable`1 options, Nullable`1 command, String commandArguments, Boolean openConsole, CancellationToken cancel, Boolean isRecursion) in /home/user/WalletWasabi/WalletWasabi/Hwi/HwiClient.cs:line 57
   at WalletWasabi.Hwi.HwiClient.EnumerateAsync(CancellationToken cancel, Boolean isRecursion) in /home/user/WalletWasabi/WalletWasabi/Hwi/HwiClient.cs:line 249
   at WalletWasabi.Gui.Tabs.WalletManager.HardwareWallets.ConnectHardwareWalletViewModel.EnumerateIfHardwareWalletsAsync() in /home/user/WalletWasabi/WalletWasabi.Gui/Tabs/WalletManager/HardwareWallets/ConnectHardwareWalletViewModel.cs:line 414
   at WalletWasabi.Gui.Tabs.WalletManager.HardwareWallets.ConnectHardwareWalletViewModel.<>c__DisplayClass6_0.<<-ctor>b__13>d.MoveNext() in /home/user/WalletWasabi/WalletWasabi.Gui/Tabs/WalletManager/HardwareWallets/ConnectHardwareWalletViewModel.cs:line 129

```

achow101:
There weren't any changes to exist codes. I'm not seeing anything unusual with it on Arch.

If you use HWI by itself, do you get the same issue? Is there any error output besides the exit code?

achow101:
There weren't any changes to exit codes. I'm not seeing anything unusual with it on Arch.

If you use HWI by itself, do you get the same issue? Is there any error output besides the exit code?

molnard:
According to @lontivero: 

> Yes, Ubuntu 18.04 comes with libc6 (2.27) and this new hwi requires (2.28).

This upgrade is necessary? It will break HWI for Ubuntu 18.04 users.

achow101:
> According to @lontivero:
> 
> > Yes, Ubuntu 18.04 comes with libc6 (2.27) and this new hwi requires (2.28).
> 
> This upgrade is necessary? It will break HWI for Ubuntu 18.04 users.

Hmm. That's odd. The docker container that does the build should be using debian stretch which has an older libc6. I'll investigate that.

achow101:
@molnard @lontivero I've redone a build from scratch (nuked the docker containers and rebuilt them) and uploaded the results. These match the travis results too (the previous ones didn't) so the containers were probably dirty from build system changes I was experimenting with. Can you try the new binaries? They should be linked to an older libc.

lontivero:
Where are the new binaries?

achow101:
On the releases page. I overwrote the original release.

lontivero:
It works in the Ubuntu 18.04 (no libc6 error). Thank you!

molnard:
Tested, works well! Thx!

PR #322, opened 2020-04-21
Fix macOS hwi-qt build

ben-kaufman:
Fix `sed` command used in the `contrib/generate-ui.sh` script for generating the `hwi-qt` build.

The current build exists in the [release](https://github.com/bitcoin-core/HWI/releases/tag/1.1.0) is broken on macOS, generating a new one with this fix should solve the issue.

### DIFF
--- a/contrib/generate-ui.sh
+++ b/contrib/generate-ui.sh
@@ -5,6 +5,6 @@ for file in *.ui
 do
     gen_file=ui_`echo $file| cut -d. -f1`.py
     pyside2-uic $file -o $gen_file
-    sed -i 's/raise()/raise_()/g' $gen_file
+    sed -i '' -e 's/raise()/raise_()/g' $gen_file
 done
 popd


instagibbs:
tACK on Mac https://github.com/bitcoin-core/HWI/pull/322/commits/57633c6114fdb6075269637531290cdf9b4747f2

achow101:
@ben-kaufman Can you check that this build works as expected: https://hwi-travis-builds.s3.amazonaws.com/bitcoin-core/HWI/1401/1401.10/hwi-1.1.0-mac-amd64.tar.gz

ben-kaufman:
@achow101 Yes, just checked and it works well now, thanks!

achow101:
Looks like this is causing `generate-ui.sh` to fail on linux now. Getting `sed: can't read : No such file or directory`.

instagibbs:
Is it difficult to add the build in a test?

achow101:
> Is it difficult to add the build in a test?

It already is. Seems like the bash scripts just aren't setup to cause a travis failure if they fail to run successfully, and we have no way to test the `hwi-qt` binary.

I have a fix for this.

ben-kaufman:
@achow101 I added in #334 a condition to check if the OS used is macOS, and if so use the code from here, otherwise use the previous one which worked on Linux. Is that a good solution?

PR #323, opened 2020-04-26
Toggle Passphrase GUI

ben-kaufman:
Adds a "Toggle Passphrase" button to the GUI, enabled for Trezor and Keepkey devices (0075e7f).

This PR also fixes Keepkey error where the command finished with an error message while it should have told the user to use `sendpin` to confirm the change, since unlike Trezor, Keepkey requires to send the pin to confirm the change of device settings (ff6cdfa).


achow101:
ACK 0075e7fde2c9bf904035490bdefe2399a6e3bfd8

PR #324, opened 2020-04-29
Multisig display address

ben-kaufman:
Added support for multisig addresses in the `displayaddress` command for Coldcard, Trezor, and Keepkey.

This PR is based on #279 and should replace it.
For now, I mostly preserved justinmoon's API as he described it in [this message](https://github.com/bitcoin-core/HWI/pull/279#issue-341248821). I'll just add a few notes on that:
- To prevent Coldcard format of sending all paths from being inconsistent with that of Trezor I allowed sending multiple paths to the Trezor too and it will simply find the correct one. That shouldn't be necessary but I just chose to add that to maintain consistency. That's not mandatory on Trezor though, so it still works just fine with either a single or multiple paths.
- I also left the `--sh_wpkh` and `--wpkh` flags for now. I think it will be best to rename them to something like `--wrapped-segwit` and `bech32` or `--native-segwit`, but wanted to raise that up here too before changing.
- ~~As for descriptor support, I did not add it yet since I didn't want this PR to be too big. I plan to add that support in a separate PR after this one is done, but if you think I should have it all in this PR that's possible too.~~
**Edit:** I also added descriptor support for multisig, [see this comment](https://github.com/bitcoin-core/HWI/pull/324#issuecomment-621926077).

The tests should cover Coldcard, Trezor, and Keepkey with 2-3 multi-sig. The setup for the test is borrowed mostly from the `sign_tx` tests code.

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -40,7 +40,7 @@ def backup_device_handler(args, client):
     return backup_device(client, label=args.label, backup_passphrase=args.backup_passphrase)
 
 def displayaddress_handler(args, client):
-    return displayaddress(client, desc=args.desc, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh)
+    return displayaddress(client, desc=args.desc, path=args.path, sh_wpkh=args.sh_wpkh, wpkh=args.wpkh, redeem_script=args.redeem_script)
 
 def enumerate_handler(args):
     return enumerate(password=args.password)
@@ -177,6 +177,7 @@ def process_commands(cli_args):
     group.add_argument('--path', help='The BIP 32 derivation path of the key embedded in the address, default follows BIP43 convention, e.g. m/84h/0h/0h/1/*')
     displayaddr_parser.add_argument('--sh_wpkh', action='store_true', help='Display the p2sh-nested segwit address associated with this key path')
     displayaddr_parser.add_argument('--wpkh', action='store_true', help='Display the bech32 version of the address associated with this key path')
+    displayaddr_parser.add_argument('--redeem_script', help='P2SH redeem script')
     displayaddr_parser.set_defaults(func=displayaddress_handler)
 
     setupdev_parser = subparsers.add_parser('setup', help='Setup a device. Passphrase protection uses the password given by -p. Requires interactive mode')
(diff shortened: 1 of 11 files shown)


instagibbs:
>I also left the --sh_wpkh and --wpkh flags for now. I think it will be best to rename them to something like --wrapped-segwit and bech32 or --native-segwit, but wanted to raise that up here too before changing.

`bech32` is ambiguous, `wpkh` only has one interpretation of script type. So ACK not coupling those changes.

ben-kaufman:
@instagibbs so maybe `--native-segwit`? 
The problem with keeping it just `wpkh` is that now it could mean either `wpkh` or `wsh` which makes the name somewhat misleading. We could also just completely separate that by adding `--sh_wsh` and `--wsh` options? Though that would add some code that could be spared.

instagibbs:
> The problem with keeping it just wpkh is that now it could mean either wpkh or wsh

Oh you mean the way you implemented it means that, sure. This could be deprecated in favor of descriptor input only since really that's what we're re-engineering here?

ben-kaufman:
With descriptors it’s not needed so whenever a descriptor is used these options are disallowed. It’s only necessary when using path (which I don’t think should be deprecated as it’s also good to have).
Also this PR doesn’t support yet descriptors anyway (I thought to separate that to another PR, but can also add that here).

ben-kaufman:
I added now support also for multisig descriptors.
Now when using the `displayaddress` command, it will be possible to just pass a descriptor instead of `--path` and `--redeemscript` (which will be calculated automatically from the descriptor). I also added tests to cover that.

ben-kaufman:
@achow101 @instagibbs just pinging you here as I think it would be really useful to have this feature. Would very much appreciate your review here when you have time, and will be happy to make changes if you have any comments.

achow101:
ACK on the most recent changes, but needs rebase. Sorry I missed this earlier.

ben-kaufman:
@achow101 thanks, I resolved the conflicts.
I see though that the tests are failing both here and on master due to some issue with ColdCard. Should I look into that here, or should I ignore it for this PR?

achow101:
> Should I look into that here, or should I ignore it for this PR?

I'm pretty sure that's unrelated and can be ignored for this PR.

Can you drop the merge commit so that the commits are based on current master instead of merged into it?

ben-kaufman:
Sure, I rebased it now so it's just the 4 commits as before.

achow101:
ACK 967138c2db19d6f82368661f7d5dbba5d00c0b0f

PR #330, opened 2020-05-08
Added Coldcard libraries & tests for HSM mode

ceyonur:
Recently they announced a new feature HSM that can automatically sign transactions. HWI can support this mode with an updated ckcc version. Some commands are blocked in HSM mode but with the proper configuration they can be used. 

For now it's up to user to configure a proper setting for HSM (as in docs). They need to tell the HSM mode to share xpubs and addresses. In the default mode HSM only shares master xpub ('m'), which was different than the `get_master_xpub`(m/44'/0'/0') in the `hwwclient.py`. I did not want to ruin current flow for the non-HSM coldcard so did not change `coldcard.py`. However if we can find a workaround and work with path (m) only, we can remove the configuration limitation from the docs. 

Seems like `backup` won't work in any case with HSM.

Whitelisted commands:
```
HSM_WHITELIST = frozenset({
    'logo', 'ping', 'vers',     # harmless/boring
    'upld', 'sha2', 'dwld', 'stxn',     # up/download/sign PSBT needed
    'mitm','ncry',              # maybe limited by policy tho
    'smsg',                     # limited by policy
    'blkc', 'hsts',             # report status values
    'stok', 'smok',             # completion check: sign txn or msg
    'xpub', 'msck',             # quick status checks
    'p2sh', 'show',             # limited by HSM policy
    'user',                     # auth HSM user, other user cmds not allowed
    'gslr',                     # read storage locker; hsm mode only, limited usage
})
```
https://github.com/Coldcard/firmware/blob/2020-04-30T1612-v3.1.3/shared/usb.py#L50

More information: https://coldcardwallet.com/docs/ckbunker-hsm


instagibbs:
candidate for an auto-approve Wasabi mixing device?

ceyonur:
@instagibbs thats a possibility. but not sure if this PR needed anymore. @achow101 already updated the ckcc lib to 1.0.2. so it should've been covered in master.

PR #332, opened 2020-05-16
Mark pyside2 as optional and update dependencies

achow101:
Marks pyside2 as optional and part of an extra `qt`. In doing so, the dependencies in the poetry.lock file are also updated. `setup.py` is regenerated to pick up this change.

Closes #331

PR #334, opened 2020-05-25
Update generate-ui.sh

ben-kaufman:
Fix Linux issue reported in #322


achow101:
It should be possible to have a single command valid for both implementations of `sed`. I've tried this in https://github.com/bitcoin-core/HWI/pull/332/commits/71e16fb6ca83dde0cfee2a03d4ef66232de94a2d (#332)

Issue #335, opened 2020-05-25
Mac Build Failure


achow101:
Travis is currently failing to build the Mac binaries: https://travis-ci.org/github/bitcoin-core/HWI/jobs/690750926 which is a problem as we use Travis to build the Mac binaries for release.

Unfortunately I don't have a Mac to debug this. Can someone with a Mac take a look at that and try to figure out what's wrong? Alternatively, I should get a Mac (or someone can buy me one).

instagibbs:
what are the specific steps to reproduce? I might be able to take a look.

achow101:
Looks like it's just with running `contrib/build_bin.sh`. You might want to install pyenv first though. It's all in `travis.yml`.

instagibbs:
Looks like my build on OSX 10.15.4 works fine :/

PR #336, opened 2020-05-26
Strip all local symbols to avoid xcode complaint

instagibbs:
Fixes https://github.com/bitcoin-core/HWI/issues/335 at least locally for my build on Catalina. Didn't test if it breaks linux et al but the option appears to be the same.

Alternative I tried was `-u -r` which ended up with identical binaries.


instagibbs:
Other builds are having serious trezor issues(I think) but osx build completed 👍

achow101:
ACK 14a7892d8ab776e3163ebaed9d6eb102de1f5c98

PR #348, opened 2020-06-27
annotated

fametrano:
As exercise to explore, understand, and possibly extend hwi, I have annotated the HardwareWalletClient and moved code comments as proper doc-strings. Of course I might have been missing something, so suggestions/corrections are welcome and I will update the annotations accordingly.

Being the main class, HardwareWalletClient is the first file a developer would look at and IMO deserves to be annotated to easily explain the class intended usage.

If this contribution is appreciated I am willing to annotate more (eventually all) code, on this PR branch or using a separate follow up PR.

### DIFF
--- a/hwilib/hwwclient.py
+++ b/hwilib/hwwclient.py
@@ -1,82 +1,181 @@
+from typing import Dict, Optional, Union
+
 from .base58 import get_xpub_fingerprint_hex
+from .serializations import PSBT
 
-# This is an abstract class that defines all of the methods that each Hardware
-# wallet subclass must implement.
+
 class HardwareWalletClient(object):
+    """Create a client for a HID device that has already been opened.
 
-    # device is an HID device that has already been opened.
-    def __init__(self, path, password, expert):
+    This abstract class defines the methods
+    that hardware wallet subclasses should implement.
+    """
+
+    def __init__(self, path: str, password: str, expert: bool) -> None:
         self.path = path
         self.password = password
         self.message_magic = b"\x18Bitcoin Signed Message:\n"
         self.is_testnet = False
-        self.fingerprint = None
-        self.xpub_cache = {}
+        self.fingerprint: Optional[str] = None
+        # {bip32_path: <xpub string>}
+        self.xpub_cache: Dict[str, str] = {}
         self.expert = expert
 
-    # Get the master BIP 44 pubkey
-    def get_master_xpub(self):
-        return self.get_pubkey_at_path('m/44\'/0\'/0\'')
+    def get_master_xpub(self) -> Dict[str, str]:
+        """Return the master BIP44 public key.
 
-    # Get the master fingerprint
-    def get_master_fingerprint_hex(self):
-        master_xpub = self.get_pubkey_at_path('m/0h')['xpub']
+        Retrieve the public key at the "m/44h/0h/0h" derivation path.
+
+        Return {"xpub": <xpub string>}.
+        """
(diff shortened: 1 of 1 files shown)


hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
+
+        Return {"xpub": xpub}
+        """
+        # FIXME is "m/44'/0'/0'" just an example?
+        # how is testnet handled?
+        # shouldn't this be
+        # return self.get_pubkey_at_path(self.path)
+        # or according to get_master_fingerprint_hex
+        # return self.get_pubkey_at_path("m/0h")
+        return self.get_pubkey_at_path("m/44'/0'/0'")
```
achow101, line 34: Re your questions: This follows the BIP 44 default, although it indeed should be different for testnet.

hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
+    def get_pubkey_at_path(self, bip32_path: str) -> Dict[str, str]:
+        """Return a dictionary with the xpub.
+
+        Retrieve the public key at the specified BIP32 derivation path.
+
+        Return {"xpub": xpub}
+        """
+        raise NotImplementedError("not implemented by HardwareWalletClient base class")
+
+    def sign_tx(self, tx) -> str:
```
achow101, line 50: This actually returns a `Dict[str, str]`. All of these functions should return a dict.

hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
+        """Return a hex string with the signed transaction.
```
achow101, line 51: The return value is a base64 encoded PSBT, not hex. The original comment is out of date.

it returns a dict of the form

    {'psbt': <base64 psbt string>}

hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
+
+        The tx must be in the combined unsigned transaction format.
```
achow101, line 53: The tx is a `PSBT` object (see `serializations.py`)

hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
+        """
+        raise NotImplementedError("not implemented by HardwareWalletClient base class")
+
+    def sign_message(self, message: str, bip32_path: str) -> str:
+        """Return a base64 encoded string with the signed message.
```
achow101, line 58: This returns a dict of the form

    {'signature': <base64 signature string>}

hwilib/hwwclient.py
```diff
@@ -1,82 +1,167 @@
...
-    def toggle_passphrase(self):
-        raise NotImplementedError('The HardwareWalletClient base class does not implement this method')
+    def get_pubkey_at_path(self, bip32_path: str) -> Dict[str, str]:
+        """Return a dictionary with the xpub.
+
+        Retrieve the public key at the specified BIP32 derivation path.
+
+        Return {"xpub": xpub}
+        """
+        raise NotImplementedError("not implemented by HardwareWalletClient base class")
```
achow101, line 48: We should avoid changing these `raise` lines to preserve the git blame. The error message should also remain as it was.

fametrano:
Hopefully I've addressed all points and I look forward to other comments/suggestions

achow101:
ACK, please squash.

fametrano:
done

achow101:
ACK 465f4fd8ed1800cb42197cdff1058af00511ad83

PR #349, opened 2020-06-27
fixed formatting

fametrano:
As exercise to explore, understand, and possibly extend hwi, I have formatted the code base to remove *hundreds* of warnings from flake8 / mypy / pylint. Not all of them have been removed, as I didn't want to touch any "real" code.

I am fully aware that coding style formatting can be quite controversial, anyway I used black as formatter: "black is opinionated, so you don't have to"; I personally advocate black as a very good common ground for collaborative open-source projects.

While many files have been touched, all the changes are just formatting. I skipped the hardware vendor folders/files, just in case that code is maintained by vendor developers which might prefer to copy/paste from not-so-clean codebases.

I also used isort for fixing the order of import directives (see .isort.cfg) and adopted better flake8 defaults.

If this contribution is appreciated, the residual issues might be tackled in different PRs, as they might deserve a case-by-case analysis.


achow101:
NACK. Sorry, formatting and general refactors are not usually accepted because they will force rebases and break git blame.

fametrano:
I hoped a relatively young project could live with a one-off solution of continuity... but I fully understand the point.
I will just use this effort to submit issues/PRs once I figure out the relevance of the residual warnings.

jonatack:
I agree with @achow101. This PR would also have required a non-trivial review effort that can be applied more productively (review is a scarce resource already) and introduce subtle but vital errors that sneak past review.

instagibbs:
Linters against new lines of code (that are being changed anyways) are a
different beast.

PR #351, opened 2020-07-02
Fix Ledger Bitcoin App 1.4.0 multple segwit input signing

achow101:
A workaround for the [multiple segwit input signing issue](https://github.com/LedgerHQ/app-bitcoin/issues/154) was added to btchip-python. So we update our copy of btchip-python to include this fix.

Also updates the speculos and bitcoind patches so that travis works.

Note that speculos doesn't have a 1.4.0+ bitcoin app.


instagibbs:
tACK https://github.com/bitcoin-core/HWI/pull/351/commits/d0de7b13266bc1ac36bc47ae5f49b55efd4fc177 tested multiple input segwit tx on ledger nano s with *non-trusted inputs*, bitcoin app 1.4.2, so at least nothing is broken for old flow.

PR #353, opened 2020-07-03
Support multisig xpub descriptors on Trezor

ben-kaufman:
Trezor Model T has an option when displaying a multisig address to also show all xpubs used as cosigners of the multisig. Currently, the display address function did not support using xpubs, so here I added support in Trezor for such descriptors, which will make xpubs show up correctly.

### DIFF
--- a/test/test_device.py
+++ b/test/test_device.py
@@ -13,6 +13,7 @@ import unittest
 from authproxy import AuthServiceProxy, JSONRPCException
 from hwilib.base58 import xpub_to_pub_hex
 from hwilib.cli import process_commands
+from hwilib.descriptor import AddChecksum
 from hwilib.serializations import PSBT
 
 # Class for emulator control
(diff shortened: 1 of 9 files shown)


stepansnigirev:
Just tried it on Trezor T - works. `xpubs` are displayed correctly.

achow101:
Can you add tests for this? This needs a test for Trezor and a test for the descriptors module.

ben-kaufman:
I'll add test for the descriptor module, but I don't think there's a way to test the xpubs displayed on Trezor, as the emulator doesn't return them.

achow101:
The Trezor will respond to HWI with the address it is displaying, so we can check that. Have a look at the existing display address tests.

ben-kaufman:
Oh ok, I thought you meant testing the xpubs displayed.
I'll add a test checking for the multisig descriptor address returned.

ben-kaufman:
@achow101 I added the tests, but due to #370, it doesn't even get to the Trezor test I added. I assume #370 should be handled in a separate PR, right? If so maybe you could try running the relevant test locally and check that it works for you?

achow101:
> I assume #370 should be handled in a separate PR, right?

Yes. I'll check the test locally with an older version of the firmware.

test/test_device.py
```diff
@@ -662,6 +662,14 @@ def test_display_address_multisig_descriptor(self):
         sh_wsh_multi_desc = sh_wsh_multi_desc.replace("'", "h").split('#')[0]
         wsh_multi_desc = wsh_multi_desc.replace("'", "h").split('#')[0]
 
+        # descriptor with xpubs
+        if self.full_type == 'trezor_t':
+            account_xpub = self.do_command(self.dev_args + ['getxpub', 'm/48h/0h/0h/2h'])['xpub']
+            result = self.do_command(self.dev_args + ['displayaddress', '--desc', 'wsh(multi(2, [' + self.fingerprint + '/48h/0h/0h/2h]' + account_xpub + '/0/0,[' + self.fingerprint + '/48h/0h/0h/2h]' + account_xpub + '/1/0))'])
+            self.assertNotIn('error', result)
+            self.assertNotIn('code', result)
+            self.assertIn('address', result)
```
achow101, line 671: For a sanity check, can you put the descriptor through the `deriveaddresses` RPC and check that bitcoind returns the same address as the device?

ben-kaufman: Sure, added. Let me know if that works for you.

test/test_device.py
```diff
@@ -657,11 +657,21 @@ def test_display_address_multisig_descriptor(self):
...
 
+        # descriptor with xpubs
+        if self.full_type == 'trezor_t':
+            account_xpub = self.do_command(self.dev_args + ['getxpub', 'm/48h/0h/0h/2h'])['xpub']
+            desc = 'wsh(multi(2, [' + self.fingerprint + '/48h/0h/0h/2h]' + account_xpub + '/0/0,[' + self.fingerprint + '/48h/0h/0h/2h]' + account_xpub + '/1/0))'
+            result = self.do_command(self.dev_args + ['displayaddress', '--desc', desc])
+            self.assertNotIn('error', result)
+            self.assertNotIn('code', result)
+            self.assertIn('address', result)
+            self.assertEqual(self.wrpc.deriveaddresses(desc)[0], result['address'])
```
achow101, line 673: `deriveaddresses` needs the checksum. Also regtest and testnet have different prefixes. This diff should handle that, and also use BIP 45 (for some reason 48 wasn't working with my emulator).

Don't forget `from hwilib.descriptor import AddChecksum` at the top of `test_device.py`

```suggestion
            account_xpub = self.do_command(self.dev_args + ['getxpub', 'm/45h/0h/0h/2h'])['xpub']
            desc = 'wsh(multi(2,[' + self.fingerprint + '/45h/0h/0h/2h]' + account_xpub + '/0/0,[' + self.fingerprint + '/45h/0h/0h/2h]' + account_xpub + '/1/0))'
            result = self.do_command(self.dev_args + ['displayaddress', '--desc', desc])
            self.assertNotIn('error', result)
            self.assertNotIn('code', result)
            self.assertIn('address', result)
            addr = self.wrpc.deriveaddresses(AddChecksum(desc))[0]
            # removes prefix and checksum since regtest gives
            # prefix `bcrt` on Bitcoin Core while wallets return testnet `tb` prefix
            self.assertEqual(addr[4:58], result['address'][2:56])
```

ben-kaufman:
Thanks, updated. Sorry about the issue there, for some reason, I can't build the simulators on my machine so can't test it locally...

achow101:
ACK 78a8801f949cb8cb12b351f81942daaf86a0a737

PR #354, opened 2020-07-03
Update Ledger udev rules

achow101:
From https://github.com/LedgerHQ/udev-rules/blob/master/20-hw1.rules


nopara73:
@molnard @yahiheb @lontivero @MaxHillebrand Be aware of this that Ledger & Linux & Wasabi users may need to update their udev rules in case someone asks.

PR #355, opened 2020-07-15
Trezor: Add multisig change address verification

ben-kaufman:
Trezor has the ability to verify that the change address of a multisig transaction is derived from the same cosigners `xpubs` used in the inputs, to make sure it was not modified into having different cosigners without the user knowing so.
Here @stepansnigirev and I added support for this type of multisig change address verification by parsing the `PSBT_GLOBAL_XPUB` fields of the PSBT (if available) and using it to verify the change address.

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -38,7 +38,9 @@ from .trezorlib import (
 )
 from .trezorlib import messages as proto
 from ..base58 import (
+    encode as base58_encode,
     get_xpub_fingerprint,
+    hash256,
     to_address,
     xpub_main_2_test,
 )
@@ -58,6 +60,7 @@ from types import MethodType
 import base64
 import logging
 import sys
+import struct
 
 py_enumerate = enumerate # Need to use the enumerate built-in but there's another function already named that
 
@@ -94,6 +97,38 @@ def parse_multisig(script):
     # Build MultisigRedeemScriptType and return it
     multisig = proto.MultisigRedeemScriptType(m=m, signatures=[b''] * n, pubkeys=pubkeys)
     return (True, multisig)
+
+# Parses the PSBT_GLOBAL_XPUB fields of a PSBT as multisig pubkeys
+def parse_multisig_xpubs(tx, psbt_in_out, multisig):
+    try:
+        old_pubs = [k.node.public_key for k in multisig.pubkeys]
+        if len(old_pubs) != sum([len(xpubs) for xpubs in tx.global_xpubs.values()]):
+            # global xpubs weren't found
+            return multisig
+        new_pubs = []
+        xpubs = [item for sublist in tx.global_xpubs.values() for item in sublist]
+        derivations = [list(struct.unpack("<" + "I" * (len(der) // 4), der)) * len(tx.global_xpubs[der]) for der in tx.global_xpubs.keys()]
+        for pub in old_pubs:
+            der = list(psbt_in_out.hd_keypaths[pub])
+            for i, derivation in py_enumerate(derivations):
+                if der[0] == derivation[0]:
+                    idx = i
+                    for i in range(len(derivation)):
+                        if der[i] != derivation[i]:
+                            # derivations mismatch
+                            return multisig
+                    break
+            xpub = xpubs[idx]
+            address_n = der[len(derivations[idx]):]
+            xpub_obj = ExtendedKey()
+            xpub_obj.deserialize(base58_encode(xpub + hash256(xpub)[:4]))
+            hd_node = proto.HDNodeType(depth=xpub_obj.depth, fingerprint=der[0], child_num=xpub_obj.child_num, chain_code=xpub_obj.chaincode, public_key=xpub_obj.pubkey)
+            new_pub = proto.HDNodePathType(node=hd_node, address_n=address_n)
+            new_pubs.append(new_pub)
+        return proto.MultisigRedeemScriptType(m=multisig.m, signatures=multisig.signatures, pubkeys=new_pubs)
+    except:
+        # If not all necessary data is available or malformatted, return the original multisig
+        return multisig
 
 def trezor_exception(f):
     def func(*args, **kwargs):
@@ -255,8 +290,7 @@ class TrezorClient(HardwareWalletClient):
                 # Check for multisig
                 is_ms, multisig = parse_multisig(scriptcode)
                 if is_ms:
-                    # Add to txinputtype
-                    txinputtype.multisig = multisig
+                    txinputtype.multisig = parse_multisig_xpubs(tx, psbt_in, multisig)
                     if not is_wit:
                         if utxo.is_p2sh:
                             txinputtype.script_type = proto.InputScriptType.SPENDMULTISIG
@@ -330,10 +364,9 @@ class TrezorClient(HardwareWalletClient):
                     else:
                         raise BadArgumentError("Output is not an address")
 
-                # Add the derivation path for change, but only if there is exactly one derivation path
+                # Add the derivation path for change
                 psbt_out = tx.outputs[i]
-                if len(psbt_out.hd_keypaths) == 1:
-                    _, keypath = next(iter(psbt_out.hd_keypaths.items()))
+                for _, keypath in psbt_out.hd_keypaths.items():
                     if keypath[0] == master_fp:
                         wit, ver, prog = out.is_witness()
                         if out.is_p2pkh():
@@ -349,7 +382,9 @@ class TrezorClient(HardwareWalletClient):
                                 txoutput.script_type = proto.OutputScriptType.PAYTOP2SHWITNESS
                                 txoutput.address_n = keypath[1:]
                                 txoutput.address = None
-
+                        is_ms, multisig = parse_multisig(psbt_out.witness_script if wit else psbt_out.redeem_script)
+                        if is_ms:
+                            txoutput.multisig = parse_multisig_xpubs(tx, psbt_out, multisig)
                 # append to outputs
                 outputs.append(txoutput)
 
--- a/hwilib/serializations.py
+++ b/hwilib/serializations.py
@@ -758,6 +758,7 @@ class PSBT(object):
         self.inputs: List[PartiallySignedInput] = []
         self.outputs: List[PartiallySignedOutput] = []
         self.unknown: Dict[bytes, bytes] = {}
+        self.global_xpubs: Dict[bytes, List[bytes]] = {}
 
     def deserialize(self, psbt: str) -> None:
         psbt_bytes = base64.b64decode(psbt.strip())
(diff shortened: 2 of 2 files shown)


instagibbs:
concept ACK

instagibbs:
any way to have automatic tests for the emulator? (Asking out loud)

stepansnigirev:
I am not sure about automatic testing with a simulator, but I can give a few test vectors I was using.

Signer 1:
mnemonic: `anxiety caught grab depend come feel auto sell aisle brisk grow wagon`
multisig derivation: `m/48h/1h/0h/2h`
xpub: `tpubDEZutCnZY4syurTf9bwP6cWC7ikZmHWdyzgK7wqE9CSKTgsBBLLa1yosC6Z7nWTS5GKfn6B3fBRgdwYnkgM6F8Hwh9UkvrKYtTbBnHhv9o5`

Signer 2:
mnemonic: `more deputy dolphin twelve excite crush cause purchase twelve pill name number ceiling spawn latin voice caught erode`
multisig derivation: `m/48h/1h/0h/2h`
xpub: `tpubDF3Nnrvc5wjeiPc9yfcGAqg5Co3obX5BXS5Cn1Azs8eitQD9mwKzCFVkMavgWcVBmre9cKhrk7sGcov34RhKkDp46d2r8DnJDBA7m8Uz68D`

PSBT with witness utxo, non-witness utxo and xpubs: (1-of-2 multisig)
`cHNidP8BAH0CAAAAARyU+KOy5BHVN/+yykeT1RMV0/luK+BcVjtOR85PENBcAAAAAAD+////AhAJkwAAAAAAIgAgFQ3T27HsLW6F9w+KmEuGdbiC2n5fH9aJcsE4QXOQ+dnz4AEAAAAAABYAFG/wUHGvbEs/3RQIpNhYdh/09gwmAAAAAE8BBDWHzwRaBl9FgAAAAgqNKI9KQB0co4VtKMgXYmGwCJHXVNomV6XLJLISJSiqA40bX7z5LIGR89K0ptVuB5vVXpZAvznMN/Xwyi7RlpYWFL8nvXswAACAAQAAgAAAAIACAACATwEENYfPBJpyfx2AAAACWQNr8kXKMtaSlSm1zNDU2K7g1y2uJsED7YUgjxT7m2wC4wCUMFkmKHgu8tTU9WLM0hfNMZUioZGhRI+3Pae8gXgUjM5j+DAAAIABAACAAAAAgAIAAIAAAQD9EQECAAAAAAEBmIPAR1hdKOwey/Yaj/WDeDxS7O2xuLhX42z7JxYEM/kBAAAAAP7///8CwPaUAAAAAAAiACBJXrCQI9okIGvUOgHDffK6JVBks2XWhQNjtMPesy76DgOTAwAAAAAAFgAUb/BQca9sSz/dFAik2Fh2H/T2DCYDAEcwRAIgPTDowe9vcGvAjBSefcwu6F5WoFU4n4xnvcrLIfCm1yECIApZHlp5vqzmVc50pe+JsYP1yd4S8yg5HO6HqhPwMuSeAUdRIQLQjk6VDHTx7y7G1zTS6plioiIoatB6PQlSUGMRjKXGQiEC7deljS/x5IPTX5KjLlNgdCP5NrKb+VYTyrJLC3+S4PFSrgAAAAABASvA9pQAAAAAACIAIElesJAj2iQga9Q6AcN98rolUGSzZdaFA2O0w96zLvoOAQVHUSECTGmR7fjieeJ4cV7v5SH26sJNI+0/fhbbjKCsqlu8DbwhA0hH+lU7ZfnD1k5xM5xzZmCiRhcD1iaKyAJTdwN+qoytUq4iBgJMaZHt+OJ54nhxXu/lIfbqwk0j7T9+FtuMoKyqW7wNvByMzmP4MAAAgAEAAIAAAACAAgAAgAEAAAAAAAAAIgYDSEf6VTtl+cPWTnEznHNmYKJGFwPWJorIAlN3A36qjK0cvye9ezAAAIABAACAAAAAgAIAAIABAAAAAAAAAAABAUdRIQKQtDFITWCH33zEXk11rdDLWTZtUIigtXyGa3+dwoE5/yEDwc1ODyUBPxNZKCGjaMqzBMEco8camSnw66SeEkxoA3FSriICApC0MUhNYIfffMReTXWt0MtZNm1QiKC1fIZrf53CgTn/HIzOY/gwAACAAQAAgAAAAIACAACAAQAAAAIAAAAiAgPBzU4PJQE/E1koIaNoyrMEwRyjxxqZKfDrpJ4STGgDcRy/J717MAAAgAEAAIAAAACAAgAAgAEAAAACAAAAAAA=`

hwilib/devices/trezor.py
```diff
@@ -95,6 +97,39 @@ def parse_multisig(script):
...
+        derivations = [[value[4 * i:4 * i + 4] for i in range(0, len(value) // 4)] for value in derivations]
+        derivations = [[int.from_bytes(v, 'little') for v in der] for der in derivations]
+        new_pubs = []
+        for pub in old_pubs:
+            # derivation
+            der = list(psbt_in_out.hd_keypaths[pub])
+            for i, derivation in py_enumerate(derivations):
+                if der[0] == derivation[0]:
+                    idx = i
+                    for i in range(derivation):
```
stepansnigirev, line 116: `range(len(derivation))`

hwilib/devices/trezor.py
```diff
@@ -95,6 +97,39 @@ def parse_multisig(script):
...
 
+# Parses the PSBT_GLOBAL_XPUB fields of a PSBT as multisig pubkeys
+def parse_multisig_xpubs(tx, psbt_in_out, multisig):
+    try:
+        old_pubs = [k.node.public_key for k in multisig.pubkeys]
+        xpubs = [xpub for xpub in tx.unknown.keys() if xpub.startswith(b"\x01")]
+        derivations = [tx.unknown[xpub] for xpub in xpubs]
+        # unpack
+        derivations = [[value[4 * i:4 * i + 4] for i in range(0, len(value) // 4)] for value in derivations]
+        derivations = [[int.from_bytes(v, 'little') for v in der] for der in derivations]
```
stepansnigirev, line 108: `import struct` at the top and then here
`derivations = [list(struct.unpack("<"+"I"*(len(value)//4),value)) for value in derivations]`

hwilib/devices/trezor.py
ben-kaufman: Opps... right, added. (shouldn't do changes of the last moment when submitting PRs...)

ben-kaufman:
@achow101 hi, this and #353 might have gotten a bit buried under other PRs, but would be really great to have them (I can make changes if needed)...

achow101:
I think we should update the PSBT implementation in general to handle `GLOBAL_XPUB` instead of trying to extract these from the unknowns from within the Trezor module. Doing so would also let us be able to include the xpubs in tests without needing bitcoind to support this.

ben-kaufman:
Makes sense, I'll move it there.

ben-kaufman:
@achow101 I moved the xpubs extraction to be in the PSBT serialization itself, hope that's matches what you meant.

achow101:
I was looking for more of actually handling the key type properly in deserialization as we do for all of the other fields. What you have is still pulling things out of the unknowns when we could be doing it in the big `if-else` block for global key deser.

ben-kaufman:
Oh ok yes I see that now, moved it there.

hwilib/serializations.py
```diff
@@ -758,6 +758,7 @@ def __init__(self, tx: Optional[CTransaction] = None) -> None:
         self.inputs: List[PartiallySignedInput] = []
         self.outputs: List[PartiallySignedOutput] = []
         self.unknown: Dict[bytes, bytes] = {}
+        self.global_xpubs: Dict[bytes, bytes] = {}
```
achow101, line 761: I think this is supposed to be `Dict[bytes, list[int]]`.

I think it would be better to actually flip this around though. The proposed implementation to Bitcoin Core has the global xpubs as a multimap. So the key is the derivation path, and the value is a list of xpubs for that derivation path (because there could be collisions). The type ends up being `Dict[list[int], list[bytes]]`. I think doing it this way makes it easier later when we want to do lookups by derivation path.

ben-kaufman: Ok, but as it's not possible to use `list` as a key for `dict`, I moved unpacking of the derivation outside and kept it as bytes in the PSBT deser, so the type is `Dict[bytes, list[bytes]]`.

ben-kaufman:
@achow101 is this ok now with 9d91e2f99ab378d306d03e46a56c46b13f1b7c16 ?

hwilib/devices/trezor.py
```diff
@@ -95,6 +98,38 @@ def parse_multisig(script):
     multisig = proto.MultisigRedeemScriptType(m=m, signatures=[b''] * n, pubkeys=pubkeys)
     return (True, multisig)
 
+# Parses the PSBT_GLOBAL_XPUB fields of a PSBT as multisig pubkeys
+def parse_multisig_xpubs(tx, psbt_in_out, multisig):
+    try:
+        old_pubs = [k.node.public_key for k in multisig.pubkeys]
+        if len(old_pubs) != sum([len(xpubs) for xpubs in tx.global_xpubs.values()]):
```
achow101, line 105: I'm not sure that this check is necessary. There could be more xpubs given than in the particular multisig being examined.

hwilib/devices/trezor.py
```diff
@@ -95,6 +98,38 @@ def parse_multisig(script):
...
+        for pub in old_pubs:
+            der = list(psbt_in_out.hd_keypaths[pub])
+            for i, derivation in py_enumerate(derivations):
+                if der[0] == derivation[0]:
+                    idx = i
+                    for i in range(len(derivation)):
+                        if der[i] != derivation[i]:
+                            # derivations mismatch
+                            return multisig
+                    break
```
achow101, line 120: Wouldn't a dictionary lookup be more suitable here?

hwilib/devices/trezor.py
```diff
@@ -95,6 +98,38 @@ def parse_multisig(script):
...
+            xpub = xpubs[idx]
```
achow101, line 121: I think we would want to make sure that the xpub derives the pubkey we are looking at.

achow101:
Have a look at #376. It adds a `KeyOriginInfo` class (actually part of #374 which it is based on) which we can use for the derivation path stuff. That should make it easier to have derivation path as the key in the dict. It also extends `ExtendedKey` to be able to derive child keys so we can use that to check that an xpub is correct by deriving the child pubkey.

instagibbs:
Does this need someone to pick it up?

prusnak:
@ben-kaufman @stepansnigirev Are you still interested in working on this or would you like someone else to pick this up?

ben-kaufman:
Hi, don't have much time currently and I'm not too familiar with the latest changes of HWI, so I think it'd be better if someone else could take over this PR.

stepansnigirev:
> @ben-kaufman @stepansnigirev Are you still interested to work on this or would you like someone else to pick this up?

Rewrote it for HWI v2 here: https://github.com/bitcoin-core/HWI/pull/497

ben-kaufman:
Closing as replaced by #497

PR #357, opened 2020-07-16
Trezor and Keepkey: Fix empty passphrase behavior

ben-kaufman:
Fix empty passphrase behavior for Trezor and Keepkey. Currently, it is not possible to get the details of a passphrase-enabled Trezor One or KeepKey with `enumerate` for an empty passphrase, while other commands use an empty passphrase by default. This PR differentiates between an empty passphrase (i.e. an empty string), and no passphrase (`None`). 
This, on the one hand, will allow using an empty passphrase with all commands, while by default, if no passphrase was passed at all, the device will not be usable (if passphrase is enabled), as compared to the present state where it is usable for all commands except enumerate.

As for the default passphrase option, I decided to set it to `None` as, while it could be just an empty passphrase, I think it is better to enforce user input for passphrase if is enabled. But if you think an empty passphrase as default is better I could change that, but choosing either of these options is better than the current state for consistency and usability.

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -69,7 +69,9 @@ def coldcard_exception(f):
 # This class extends the HardwareWalletClient for ColdCard specific things
 class ColdcardClient(HardwareWalletClient):
 
-    def __init__(self, path, password='', expert=False):
+    def __init__(self, path, password=None, expert=False):
+        if password is None:
+            password = ''
         super(ColdcardClient, self).__init__(path, password, expert)
         # Simulator hard coded pipe socket
         if path == CC_SIMULATOR_SOCK:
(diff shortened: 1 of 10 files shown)


hwilib/devices/coldcard.py
```diff
@@ -69,7 +69,9 @@ def func(*args, **kwargs):
 # This class extends the HardwareWalletClient for ColdCard specific things
 class ColdcardClient(HardwareWalletClient):
 
-    def __init__(self, path, password='', expert=False):
+    def __init__(self, path, password=None, expert=False):
+        if password in None:
```
achow101, line 73: `s/in/is`

ben-kaufman: right, fixed.

achow101:
This is breaking the tests.

This seems to be a breaking change since now a passphrase needs to always be provided.

ben-kaufman:
@achow101 I'm unable to fix the tests now for some reason. Seems like there it behaves differently than how my physical devices I tried it with behave...
I'm unable to setup the emulators properly on my own machine (macOS) so it's somewhat hard for me to debug what's the issue with the tests currently. I'll close this for now then and try to fix the tests, and open again once I can get the tests to pass.

PR #358, opened 2020-07-21
Implementation for D'CENT 

KyungeunKim:
Implements HWI for the D'CENT Hardware wallet.
In addition, `protobuf` is required. 

Also updates the documentation for the D'CENT.

https://dcentwallet.com/


achow101:
It would be preferable if we didn't have to add protobuf as a dependency. Trezor is able to avoid this by using a minimal implementation of protobuf, could we do a similar thing here?

Is an emulator/simulator available for this device which we can use in automated tests?

Are the files in `dcentlib/` available in an upstream repo that we can pull from for future updates?

KyungeunKim:
> It would be preferable if we didn't have to add protobuf as a dependency. Trezor is able to avoid this by using a minimal implementation of protobuf, could we do a similar thing here?
> 
> Is an emulator/simulator available for this device which we can use in automated tests?
> 
> Are the files in `dcentlib/` available in an upstream repo that we can pull from for future updates?


@achow101 
We solved the issues that you comment.

1. We removed 'protobuf' and instead use Trezor's.
2. We are not offering simulator as a policy. **MUST** we provide a simulator?
3. We create a new repository https://github.com/DcentWallet/HWI-dcentlib
   `dcentlib/` copied to this repository.  Can we add this repository in HWI as submodule?

instagibbs:
My opinions only:

> We are not offering simulator as a policy. MUST we provide a simulator?

We currently don't have a specific policy on this, although for all currently-included hww they have them. Ledger for example fairly recently released theirs, they were the last. Without one it's a bit harder to give QoS guarantees.

> We create a new repository https://github.com/DcentWallet/HWI-dcentlib
dcentlib/ copied to this repository. Can we add this repository in HWI as submodule?
That's an idea, to submodule or subtree this.

All this imo can work with the expectation that your team(I'm assuming you are with d'cent) is willing to maintain this implementation going forward. It would be on you(r team) to ensure that the d'cent implementation is ready and working prior to each release at a minimum, and if support drops off indefinitely without stated reason support may be discontinued if it becomes a maintenance burden or we deem usage of the device too high risk.

achow101:
@KyungeunKim @lix82 Since it seems like we will be adding protobuf as a dependency for Bitbox02, I think we can revert the change to use trezorlib's protobuf. This would also make sure that we don't have cross device dependencies (i.e. removing support for one device completely does not require modifying another device's implementation). Sorry for the noise.

instagibbs:
needs rebase. I think this can go ahead https://github.com/bitcoin-core/HWI/pull/358#issuecomment-674207595 now on master

prusnak:
Repo https://github.com/DcentWallet/HWI-dcentlib does not have licence information available. @KyungeunKim will you update the license in this repo? Since you are using Trezor LGPLv3 code I suggest you use the same license.

achow101:
AFAICT the d'cent wallet is not open source and there is no explanation of why. This does not meet the [device support policy](https://hwi.readthedocs.io/en/latest/devices/index.html#support-policy).

Issue #158, opened 2019-04-25
Handle when user is not in the Bitcoin app on Ledger Nano S


achow101:
We need to make sure that we give a proper error when the user is not in the Bitcoin app on the Ledger Nano S. I was unable to find a way to  detect this. It seems like what happens is that there's just an error that gets thrown with an unhelpful error message, so maybe we just need to make the error messages better?

@instagibbs Can you take a look?

instagibbs:
sounds like better error codes aren't coming until 1.5 firmware update

achow101:
Looks like some people are running into this: https://github.com/zkSNACKs/WalletWasabi/issues/3931#issuecomment-659412817

Any idea if there are better error codes now or a check we can do?

promag:
Have you tried to exchange "get app name" APDU `cla=0xb0` `ins=0x01`?

PR #362, opened 2020-07-25
Expose app name of ledger devices

promag:
Requirement for #158.

### DIFF
--- a/hwilib/devices/btchip/btchip.py
+++ b/hwilib/devices/btchip/btchip.py
@@ -26,8 +26,10 @@ from binascii import hexlify, unhexlify
 
 class btchip:
 	BTCHIP_CLA = 0xe0
+	BTCHIP_CLA_COMMON_SDK = 0xb0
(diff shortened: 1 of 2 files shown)


achow101:
I'm not sure that it is useful to expose this information. I think it would be better if we just didn't enumerate a device if it wasn't in the Bitcoin or Bitcoin Testnet apps.

promag:
@achow101 ok, so just ensure app matches the command network and fail otherwise?

achow101:
For mainnet, make sure it only matches "Bitcoin". For testnet, I think it should match both "Bitcoin" and "Bitcoin Test" (for testing, sometimes it's easier to just use the mainnet app with `--testnet`).

instagibbs:
I wouldn't mind it throwing an error if the app name isn't what's expected. Right now it's semi mysterious at points.

achow101:
Superseded by #473

PR #363, opened 2020-07-27
add BitBox02  support

benma:
This adds cli and gui support for the BitBox02. For now, only single sig support is added, with the plan to add multisig support soon after this is merged.

This PR has had an internal review over here before: https://github.com/benma/hwi/pull/1, and I tested it quite extensively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,29 +90,29 @@ The below table lists what devices and features are supported for each device.
 
 Please also see [docs](docs/) for additional information about each device.
 
-| Feature \ Device | Ledger Nano X | Ledger Nano S | Trezor One | Trezor Model T | Digital BitBox | KeepKey | Coldcard |
+| Feature \ Device | Ledger Nano X | Ledger Nano S | Trezor One | Trezor Model T | BitBox01 | BitBox02 | KeepKey | Coldcard |
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
-| Support Planned | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| Implemented | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| xpub retrieval | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| Message Signing | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| Device Setup | N/A | N/A | Yes | Yes | Yes | Yes | N/A |
-| Device Wipe | N/A | N/A | Yes | Yes | Yes | Yes | N/A |
-| Device Recovery | N/A | N/A | Yes | Yes | N/A | Yes | N/A |
-| Device Backup | N/A | N/A | N/A | N/A | Yes | N/A | Yes |
-| P2PKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2SH-P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2SH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2SH-P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| Bare Multisig Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A |
-| Arbitrary scriptPubKey Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A |
-| Arbitrary redeemScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A |
-| Arbitrary witnessScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A |
-| Non-wallet inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| Mixed Segwit and Non-Segwit Inputs | N/A | N/A | Yes | Yes | Yes | Yes | Yes |
-| Display on device screen | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| Support Planned | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| Implemented | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| xpub retrieval | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| Message Signing | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| Device Setup | N/A | N/A | Yes | Yes | Yes | Yes | Yes | N/A |
+| Device Wipe | N/A | N/A | Yes | Yes | Yes | Yes | Yes | N/A |
+| Device Recovery | N/A | N/A | Yes | Yes | N/A |  Yes | Yes | N/A |
+| Device Backup | N/A | N/A | N/A | N/A | Yes |  Yes | N/A | Yes |
+| P2PKH Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2SH-P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| P2SH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2SH-P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| Bare Multisig Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary scriptPubKey Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary redeemScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary witnessScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Non-wallet inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| Mixed Segwit and Non-Segwit Inputs | N/A | N/A | Yes | Yes | Yes | Yes | Yes | Yes |
+| Display on device screen | Yes | Yes | Yes | Yes | N/A | Yes | Yes | Yes |
 
 ## Using with Bitcoin Core
 
--- a/docs/bitbox02.md
+++ b/docs/bitbox02.md
@@ -1,0 +1,49 @@
+# BitBox02
+
+The BitBox02 is supported by HWI.
+
+Current implemented commands are:
+
+* `signtx`
+* `getxpub`
+* `displayaddress`
+* `setup`
+* `wipe`
+* `restore`
+* `backup`
+* `togglepassphrase`
+
+Multisig (P2WSH only) is supported by the BitBox02, but is not ingerated into HWI yet. Coming
+soon^{tm}.
+
+# Usage Notes
+
+## Strict keypaths
+
+The BitBox02 has strict keypath validation.
+
+The only accepted keypaths for xpubs are:
+
+- `m/49'/0'/<account'>` for `p2wpkh-p2sh` (segwit wrapped in P2SH)
+- `m/84'/0'/<account'>` for `p2wpkh` (native segwit v0)
+- `m/48'/0'/<account'>/2` for p2wsh multisig (native segwit v0 multisig).
+
+`account'` can be between `0'` and `99'`.
+
+For address keypaths, append `/0/<address index>` for a receive and `/1/<change index>` for a change
+address. Up to `10000` addresses are supported.
+
+In `--testnet` mode, the second element must be `1'` (e.g. `m/49'/1'/...`).
+
+## Signing with mixed input types
+
+The BitBox02 allows mixing inputs of different script types (e.g. and `p2wpkh-p2sh` `p2wpkh`), as
+long as the keypaths use the appropriate bip44 purpose field per input (e.g. `49'` and `84'`) and
+all account indexes are the same.
+
+Multisig and singlesig inputs cannot be mixed.
+
+## getmasterxpub and legacy addresses not supported
+
+`getmasterxpub` is the same as `getxpub` at the legacy keypath `m/44'/0'/0'`. Legacy xpub, addresses
+and inputs are not supported.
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -88,29 +88,29 @@ The below table lists what devices and features are supported for each device.
...
+| P2SH-P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| P2SH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2SH-P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| Bare Multisig Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary scriptPubKey Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary redeemScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Arbitrary witnessScript Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
+| Non-wallet inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
```
benma, line 111: What is a non-wallet input? @achow101

docs/bitbox02.md
```diff
@@ -0,0 +1,49 @@
...
+* `getxpub`
+* `displayaddress`
+* `setup`
+* `wipe`
+* `restore`
+* `backup`
+* `togglepassphrase`
+
+Multisig (P2WSH only) is supported by the BitBox02, but is not ingerated into HWI yet. Coming
+soon^{tm}.
```
TheCharlatan, line 17: What is the blocker here?

benma: - Passing the full descriptor like in https://github.com/bitcoin-core/HWI/pull/353 so the setup can be registered
- I figured smaller PRs are better. Multisig will add quite a bit of code, but this PR is already big enough to look over

README.md
achow101: Non-wallet inputs are inputs which this device cannot sign, i.e. Coinjoins.

benma: Good to know. Shouldn't most of those be N/A then? Without ownership proofs, ignoring external inputs can be dangerous.

See e.g. https://github.com/trezor/trezor-firmware/issues/37

benma:
@achow101 could you test this PR with your BitBox02 already?

achow101:
Is there a simulator for Bitbox02 that we can use in CI?

benma:
> Is there a simulator for Bitbox02 that we can use in CI?

Not yet. Until one is there (and after), I am happy to keep a close eye on the BitBox02 integration and ensure that it stays functional.

achow101:
For the device libraries, we have been preferring to copy the library source into `hwilib/devices/` so that we can modify them to reduce dependencies. It seems like we could do that here and drop the `base68` dependency and potentially the `protobuf` one as well. For protobuf, it would be preferable if this could do something like what Trezor does with protobuf (they have a mini implementation of protobuf instead of including the library).

benma:
@achow101 

It is possible to vendor the bitbox02 Python library here, though I am sure that is easier. At some point there will be updates to the library, and having poetry/pip take care of that process is more streamlined. I can drop the `base58` dependency too, it is only used in a util function, if that helps.

> so that we can modify them 

Modifying the lib here could make it harder to update it in the future.

If you insist though, I will of course vendor it :) Just seems easier to use poetry/pip as with the other dependencies that you already have.

As to protobuf: what would be the purpose of hand-rolling another implementation instead of using the `protobuf` package? It would be quite a big effort to re-implement it (and integrate it into our package, test it properly, etc.), which I would like to a avoid. The lib that is inside trezorlib is not sufficient for the bitbox02, as it makes use of more message types.

instagibbs:
My opinions only:

We should make reasonable effort to reduce dependencies where we can but:
1) If HWI is getting direct support from the manufacturer this mitigates some of that issue with respect to upkeep/making modifications to support CI. AFAIU only bitbox and d'cent have made those directed efforts for HWI inclusion so this is somewhat new for us. A little more discussion here https://github.com/bitcoin-core/HWI/pull/358#issuecomment-671969764
2) How large do the deps make deterministic builds? Provided the imports are self-contained I'm less worried because people can just elide modules if they don't want to use those devices.

benma:
@instagibbs thanks, and agree. Happy to maintain it and make sure that the bitbox02 plugin is functional continuously (which even a simulator in the CI can't guarantee; but hopefully we can provide one someday anyway).

Since the protobuf dep seems to be used now by at least two plugins, it would be good to accept it as a dependency.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,678 @@
...
+    DeviceNotReadyError,
+    UnavailableActionError,
+    DEVICE_NOT_INITIALIZED,
+    handle_errors,
+    common_err_msgs,
+)
+
+import hid  # type: ignore
+
+from .trezorlib.tools import parse_path
```
achow101, line 42: I think we should avoid doing cross device imports like this. In theory, bitbox02 support should still work even if all of the trezor stuff were deleted.

It might be useful to add some of our own utility functions like this that all devices can share.

benma: Good point. Fixed in a new commit. I can factor out common functionality in a future PR.

benma:
Looks like the CI error is independent of this PR and currently affects all PRs.

benma:
Bump. Would be great to make progress here, as some BitBox02 customers are waiting for this so they can use it with e.g. Wasabi. :pray:

achow101:
I'll test this when my USB C female to A male cable arrives.

In the meanwhile, there's a merge conflict with the readme.

hwilib/devices/bitbox02.py
```diff
@@ -135,6 +133,40 @@ def attestation_check(self, result: bool) -> None:
             )
 
 
+def _parse_path(nstr: str) -> Sequence[int]:
```
achow101, line 136: Related PR: #374

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+class KeypathError(UnavailableActionError):
+    def __init__(self, keypath: str, is_testnet: bool):
+        """
+        Keypath error exception with formatting and docs.
+        """
+        network = "testnet" if is_testnet else "mainnet"
+        msg = "The BitBox02 does not support the keypath {} on {}. Supported keypaths are:\n".format(
+            keypath, network
+        )
+        msg += "m/49'/0'/<account'> for p2wpkh-p2sh\n"
```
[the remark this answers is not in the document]
benma: True, but they are shown nicely in hwi-qt.py and possibly also by other downstream UIs. But it is safer to go without `\n`, will fix.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+        )
+
+    def _get_coin(self) -> bitbox02.btc.BTCCoin:
+        if self.is_testnet:
+            return bitbox02.btc.TBTC
+        return bitbox02.btc.BTC
+
+    def _get_xpub(self, keypath: List[int]) -> str:
+        expected_coin = 1 + HARDENED if self.is_testnet else 0 + HARDENED
+        if len(keypath) == 3:
```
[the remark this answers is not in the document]
benma: Good point. I will instead give a more general error message in case the device returns an error; the issue currently is that the BitBox02 does not give very granular error codes, so it is a bit hard to identify what the actual issue was (keypath, unsupported script, etc.?). 

I will remove the keypath validation here and the rest of the plugin.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+            assert keypath is not None
+            found_pubkeys.append(found_pubkey)
+
+            # TOOD: validate keypath
+
+            if bip44_account is None:
+                bip44_account = keypath[2]
+            elif bip44_account != keypath[2]:
+                raise BadArgumentError(
+                    "The bip44 account index must be the same for all inputs and changes"
```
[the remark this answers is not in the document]
benma: Our users very often complained that it is confusing to deal with Bitcoin Segwit (p2wpkh-p2sh), Bitcoin native segwit (p2wpkh), etc., so in a recent update we allow combining inputs with multiple types, as long as they are from the same "account".

This allows us to have a unified "Bitcoin" account in the app, and still allow for multiple accounts.

See also "Unified accounts" in the [release blog post](https://medium.com/shiftcrypto/bitbox-07-2020-lidernen-update-493f670860aa).

As to why the firmware disallows it: we have taken the general approach to disallow by default, and allow only validated use cases based. Allowing too much by default has led to plenty of vulnerabilities in the past across many vendors.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+                    },
+                }
+            )
+
+        outputs: List[bitbox02.BTCOutputType] = []
+        for output_index, (psbt_out, tx_out) in builtins.enumerate(
+            zip(psbt.outputs, psbt.tx.vout)
+        ):
+            _, keypath = find_our_key(psbt_out.hd_keypaths)
+            is_change = keypath and keypath[-2] == 1
```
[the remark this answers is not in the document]
benma: Sending the output as `bitbox02.BTCOutputInternal` tells the device that it is a change output. 

The BitBox02 enforces that change outputs have the `change` part of the keypath be `1`, hence this check. 

By adding this check, the output is shown to the user. Removing this check would mean the whole transaction will fail.

I guess if there was a tx that sent the change on a strange keypath, we should let it fail instead of showing an address to the user that they cannot effectively verify. So I will remove this additional check.

hwilib/devices/bitbox02.py
benma: Will include this after that PR is merged (can also be in a future PR).

hwilib/devices/bitbox02.py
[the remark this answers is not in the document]
benma: Fixed.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+        if not bb02.device_info()["initialized"]:
+            raise UnavailableActionError("Not initialized")
+        return bb02.root_fingerprint().hex()
+
+    def prompt_pin(self) -> Dict[str, Union[bool, str, int]]:
+        raise UnavailableActionError(
+            "The BitBox02 does not need a PIN sent from the host"
+        )
+
+    def send_pin(self) -> Dict[str, Union[bool, str, int]]:
```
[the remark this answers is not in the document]
benma: Fixed in a separate commit.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+
+    def _get_xpub(self, keypath: List[int]) -> str:
+        expected_coin = 1 + HARDENED if self.is_testnet else 0 + HARDENED
+        if len(keypath) == 3:
+            # singlesig
+            purpose, coin, account = keypath
+            if coin != expected_coin or not _keypath_check_account(account):
+                raise ValueError()
+            try:
+                # Actually we want to show ypub... or zpub... for segwit-p2sh or native segwit,
```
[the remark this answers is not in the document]
benma: Makes sense. I simplified the code accordingly.

hwilib/devices/bitbox02.py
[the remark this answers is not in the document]
benma: Done.

hwilib/devices/bitbox02.py
```diff
@@ -0,0 +1,681 @@
...
+        bb02 = self.init()
+        info = bb02.device_info()
+        if info["mnemonic_passphrase_enabled"]:
+            bb02.disable_mnemonic_passphrase()
+        else:
+            bb02.enable_mnemonic_passphrase()
+        return {"success": True}
+
+    @bitbox02_exception
+    def setup_device(
```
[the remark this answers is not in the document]
benma: The issue was that the fingerprint is not available on an uninitialized device, so the call to get the fingerprint failed.

I fixed it in the commit `commands/find_device: only get the device fingerprint if required`, which I used when testing this, but accidentally forgot to include in this PR. Please try again.

About the `wrong state`: this happens when a call is made on a uninitialized device when it should be initialized, or vice versa. Added better error handling in the commit `bitbox02: consistent errs when the device is (not) initialized`.

benma:
@achow101 thanks for the nice review. I addressed all points, please take a look.

hwilib/devices/bitbox02.py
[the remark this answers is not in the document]
benma: I took a look, I have to keep the check. The reason is that Electrum (and others?) seem to include the keypath even for non-change outputs if you send to yourself, so the presence of a keypath in the output is not sufficient to determine if it actually is a change (automatically added), or a deliberate recipient address.

The other HW wallets seem to treat it as change and not display it on their screens, which is totally fine, but the BitBox02 shows all outputs manually added by a user for verification, even if it is a send-to-self.

hwilib/gui.py
```diff
@@ -207,6 +211,41 @@ def toggle_account(self, checked):
             self.ui.path_lineedit.setEnabled(True)
             self.ui.account_spinbox.setEnabled(False)
 
+class BitBox02PairingDialog(QDialog):
```
achow101, line 214: When is this dialog shown? I can't seem to find a way to show it.

achow101: Ah, I was able to get it to show if the device was not initialized. However it was just blank until *after* I accepted on the device.

benma: The dialog is shown on first pairing, after which the pairing info is persisted and re-used. You can force a new pairing by deleting the pairing info, which on Linux is at `~/.config/bitbox/bitbox02/bitbox02.json` and on macOS is at `~/Library/Application Support/bitbox/bitbox02/bitbox02.json`.

About the blank dialog: it works for me most of the times, but after trying a few more times I can see the issue appear. I'll try to figure out how to make Qt draw the window before proceeding with the pairing.

benma: I attempted a fix in `hwilib/gui: make sure the bb02 pairing window is painted hwilib/gui: make sure the bb02 pairing window is painted`. Please try again and lmk if it worked.

achow101:
ACK b24f2aaa2debbe90bf102fb96c778733ce81259b

Going to wait and see if someone else wants to review this before I merge.

hwilib/gui.py
instagibbs: @achow101 is this resolved for you?

achow101: Yes, seems to work now.

hwilib/devices/bitbox02.py
```diff
@@ -135,6 +133,40 @@ def attestation_check(self, result: bool) -> None:
             )
 
 
+def _parse_path(nstr: str) -> Sequence[int]:
+    """
+    Adapted from trezorlib.tools.parse_path.
+    Convert BIP32 path string to list of uint32 integers with hardened flags.
+    Several conventions are supported to set the hardened flag: -1, 1', 1h
```
instagibbs, line 140: just curious: `-1` who uses this?

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/363/commits/b24f2aaa2debbe90bf102fb96c778733ce81259b

I'd like to personally check the GUI changes for non-bb but I see achow already tested so ok for now.

hwilib/devices/bitbox02.py
benma: I was also surprised to see that :shrug:

benma:
Awesome, thanks again! I'll get started on the multisig part of the plugin soon then.

PR #365, opened 2020-08-05
Fix descriptor fingerprint parsing when no derivation used

ben-kaufman:
This is a small fix for parsing a descriptor that specifies origin fingerprint but no origin derivation (depth=0).
E.g. `[90abcdef]xpub.../0/*`.

Currently, it will set origin_fingerprint to `None`, while the correct behavior would be to save the fingerprint.

### DIFF
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -147,6 +147,9 @@ class Descriptor:
                     origin_path = match.group(2)
                     # Replace h with '
                     origin_path = origin_path.replace('h', '\'')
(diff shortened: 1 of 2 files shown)


achow101:
Do you have a test case for this that we can add to `test_descriptor.py`?

ben-kaufman:
@achow101 sure, just added a test for that.

achow101:
ACK 67a9d7a322f35b3685ed48bede3c8990da116460

PR #366, opened 2020-08-06
extended sign_message to also sign bytes, not only string

fametrano:
Bitcoin message signing is usually used to sign an input string which, as first step, is encoded to bytes. The rest of the implementation assumes bytes, operates on bytes, ecdsa-signs bytes.

This PR proposes to skip the string encoding if bytes, instead of a string, are provided as input: this allow to "bitcoin message sign" an arbitrary byte sequence (note that arbitrary byte sequences cannot be always decoded to strings). 

The changes to trezorlib/btc.py are only cosmetic and that file could actually be left untouched

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -1,5 +1,7 @@
 # Coldcard interaction script
 
+from typing import Dict, Union
+
 from ..hwwclient import HardwareWalletClient
 from ..errors import (
     ActionCanceledError,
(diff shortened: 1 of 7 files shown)


instagibbs:
concept ACK. Would be a good time to add some really basic tests for signmessage?

fametrano:
> concept ACK. Would be a good time to add some really basic tests for signmessage?

sure! At least it should be ensured that the same result is obtained by passing a string or its byte encoding.
I can do that if there is a general interest in merging this PR.

instagibbs:
Should be simple to pipe it through `verifymesage` on bitcoind in the tests too.

achow101:
It seems like it would be easier to just make it always `bytes` instead of accepting both `str` and `bytes`. It would require less processing logic for each device.

achow101:
Concept ACK. Could you remove the unrelated changes?

fametrano:
> Could you remove the unrelated changes?
 I restored the unused variables in coldcadr.py

Anything else to restore?

I could leave trezorlib/btc.py untouched as that is not-relevant middle layer software; anyway, it is involved in the sign_message code flow and IMO would benefit to be accordingly documented.

The comments I removed should have been docstrings; anyway, since they would document a derived class implementation of an abstract base class method, they are redundant (most documentation generation tools, Sphinx and Epydoc included, already pull parent method docstrings).

stepansnigirev:
cACK
I think it would be nice to use hardware wallets to sign i.e. hashes of the firmware or software, or other binary files (i.e. PDFs).
Signing this kind of stuff definitely requires bytes, not strings.

achow101:
ACK d16bdf794f7090a5c7015c1e3b3398440df8a440

fametrano:
See also
- https://github.com/Coldcard/ckcc-protocol/pull/12
- https://github.com/trezor/trezor-firmware/pull/1225

PR #368, opened 2020-08-11
Add disclaimer to README about (lack of) endorsement

instagibbs:
We should be saying something like this since day 1, and now with discussion of newer devices even more prudent.


fametrano:
ACK

achow101:
ACK 3016d4a8616b042bae82bca64c2253ea2d128a19

PR #369, opened 2020-08-12
added py.typed so that static type checkers use type hints (PEP 561)

fametrano:
- Problem: mypy can't find type hints when importing from hwi
- Solution: add a py.typed file to make hwi PEP-561-compatible

At the moment hwi has type hints, but they are useless to packages importing from hwi because there is no py.typed file to tell mypy they're there.

PEP 561 specifies that packages which contain either inline type hints or type stubs should indicate their support for type hints via including a blank file named py.typed in the root of the package. Package maintainers who wish to support type checking of their code MUST add it to their package supporting typing.


achow101:
The setup.py file is automatically generated. Please modify the pyproject.toml file instead.

Not everything has inline types yet, will adding this marker file cause issues for/because of the code that hasn't been typed yet?

fametrano:
> The setup.py file is automatically generated. Please modify the pyproject.toml file instead.

done. Aside note: I'm not really familiar with pyproject.toml, but I assumed that if it used, then setup.py should be removed (or, to allow editable installs, reduced to a shim; see https://snarky.ca/what-the-heck-is-pyproject-toml/)

> Not everything has inline types yet, will adding this marker file cause issues for/because of the code that hasn't been typed yet?

When importing from hwi, as of now type checkers fail to detect type hints and issue errors, unless type checking is disabled globally or just for hwi imports. To disable import type checking is not recommended, as everything is then assumed to accept/return  ` Any `.

With the py.typed marker file, type hints are used when provided and are implicitly deduced whenever possible: for my own usage of hwi this is perfectly fine and generates no error/warning. Other users might import/use some part of the hwi codebase which has not been properly annotated yet (commands.py probably being the most relevant one): they will simply disable type-checking, report the issue, and/or contribute the required annotation. Anyway, they will not be in a worse situation compared to the current one.

Even the python standard libraries are not fully annotated yet, this is a work in progress and nobody would expect all libraries to be fully annotated. Anyway, one can have a legitimate expectation that any recent python library should be at least type-checking aware.

achow101:
ACK 287bce98942483de0889cb152d714a39713e5e44

PR #371, opened 2020-08-17
added gitignore directives for virtual environments

fametrano:
virtual environments are routinely created in the root project (sub)folder(s) and should be ignored by Git


achow101:
ACK 7a896e792be4bf33a1511dfc8ba03352f08c5bcb

Sure, why not.

PR #372, opened 2020-08-17
Cleaned up log messages

fametrano:
In my team, we comment out that line to avoid Trezor writing to our LOG.info

As they say "your mileage may vary", but a final user's LOG.info should not be polluted by third party library.

### DIFF
--- a/hwilib/devices/trezorlib/client.py
+++ b/hwilib/devices/trezorlib/client.py
@@ -72,7 +72,7 @@ class TrezorClient:
     """
 
     def __init__(self, transport, ui=None, state=None):
-        LOG.info("creating client instance for device: {}".format(transport.get_path()))
+        LOG.debug("creating client instance for device: {}".format(transport.get_path()))
         self.transport = transport
         self.ui = ui
         self.state = state
--- a/hwilib/devices/trezorlib/device.py
+++ b/hwilib/devices/trezorlib/device.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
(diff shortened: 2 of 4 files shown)


achow101:
HWI should be configured to mute that so I don't think this is necessary. Are you seeing these log lines being written anywhere?

fametrano:
I'm working on a project based on HWI and if/when I enable logging as in
```
import logging
from hwilib import commands

LOGGER = logging.getLogger(__name__)
LOGGER.info("my log message")
commands.enumerate()
```
my LOG receives the LOG.info issued by Trezor.

achow101:
Ah I see. I don't think commenting out this code is the right approach though as this logging can sometimes be helpful in debugging.

fametrano:
Yes, but it is intrusive for any user's debugging :-)
Moreover, why is only the Trezor code to be that log-invasive?
 
Anyway, I can leave with just commenting this out in my forks, unless a better approach is available,

achow101:
What about just setting it (and others) to `LOG.debug`?

fametrano:
done.

I've downgraded all LOG.info to LOG.debug, but the one in hwilib/devices/trezorlib/transport/hid.py which has been upgraded to LOG.warning (similarly to hwilib/devices/trezorlib/transport/webusb.py) 

I've also re-enabled the LOG.debug that was commented out in hwilib/devices/trezorlib/device.py

Finally, I've renamed this pull request and, if needed, I can edit the first messages of this conversation which do not make sense anymore.

achow101:
ACK 5e296ca4dca8fbf3f126ecbff507ff926808f13a

fametrano:
https://github.com/trezor/trezor-firmware/pull/1201

PR #373, opened 2020-08-18
Avoided bare except

fametrano:
Using a bare `except:` is not a good coding practice: it catches all exceptions inheriting from `BaseException`, including `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit` (which is not an error and should not normally be caught by user code).
```
BaseException
  - Exception
  - GeneratorExit
  - KeyboardInterrupt
  - SystemExit
```
This patch is sub-optimal because only specific exceptions should be catched... anyway, in situations where one wants to catch all “normal” exceptions, at least restrict to explicitly catch the `Exception` base class, instead of implicitly catching any `BaseException`.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -69,7 +69,7 @@ def find_device(password='', device_type=None, fingerprint=None, expert=False):
                 client.close()
                 continue
             return client
-        except:
+        except Exception:
             if client:
                 client.close()
             pass # Ignore things we wouldn't get fingerprints for
--- a/hwilib/devices/btchip/btchip.py
+++ b/hwilib/devices/btchip/btchip.py
@@ -84,7 +84,7 @@ class btchip:
 				self.scriptBlockLength = 50
 			else:
 				self.scriptBlockLength = 255
-		except:
+		except Exception:
 			pass			
 
 	def getWalletPublicKey(self, path, showOnScreen=False, segwit=False, segwitNative=False, cashAddr=False):
@@ -271,7 +271,7 @@ class btchip:
 					response = self.dongle.exchange(bytearray(apdu))
 					offset += dataLength
 				alternateEncoding = True
-			except:
+			except Exception:
 				pass
 		if not alternateEncoding:
 			apdu = [ self.BTCHIP_CLA, self.BTCHIP_INS_HASH_INPUT_FINALIZE, 0x02, 0x00 ]
--- a/hwilib/devices/btchip/btchipComm.py
+++ b/hwilib/devices/btchip/btchipComm.py
@@ -142,7 +142,7 @@ class HIDDongleHIDAPI(Dongle, DongleWait):
 		if self.opened:
 			try:
 				self.device.close()
-			except:
+			except Exception:
 				pass
 		self.opened = False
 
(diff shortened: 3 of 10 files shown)


achow101:
When possible, I think we should avoid making changes to the libraries that we've pulled in from upstream (i.e. trezorlib, ckcc, btchip)

instagibbs:
@fametrano I think I agree even though these changes are limited. Would be best if we patched upstream then updated the vendoring? Keeping close to upstream is optimal unless there's a serious bug being fixed here.

fametrano:
Somebody has defined "bare except" the most diabolical python antipattern (https://realpython.com/the-most-diabolical-python-antipattern/) and I personally see it as the worst offending Python coding practice from a CS/engineering point of view.

Among other annoyances, this behaviour is preventing keyboard interrupts to stop any process involving HWI.

I agree about minimizing changes to libraries pulled in from upstream, but this is a very relevant/urgent one; moreover, it must be said, most of those libraries are in a very bad state when it comes to code style/quality.

I would love to see the bitcoin community to gradually move to decent modern Python 3 standards. I'm more than willing to push this changeset upstream to trezorlib, ckcc, btchip, etc. (upstream PR links would be appreciated).

instagibbs:
No arguments on badness. I don't do it ever.

> Among other annoyances, this behaviour is preventing keyboard interrupts to stop any process involving HWI.

Ok, that's good enough for me. ACK

fametrano:
there is one bare except in HWI, 2 in the tests, and 7 in vendor code.
The first three should be a no-brain fix... unfortunately without fixing all of them most of the related problems will still be there.

Anyway, I think it should be not to hard to have the vendors accept equivalent PRs to their codebase

fametrano:
See:

- https://github.com/LedgerHQ/btchip-python/pull/47
- https://github.com/Coldcard/ckcc-protocol/pull/10
- need github link for digitalbitbox
- need github link for trezor (the Trezor Firmware Monorepo does not seem to be the one...)

achow101:
> need github link for digitalbitbox

We don't really use the digitalbitbox upstream. Also, I believe it is now unmaintained. Anyways: https://github.com/digitalbitbox/mcu/tree/master/py

> need github link for trezor (the Trezor Firmware Monorepo does not seem to be the one...)

It's in the monorepo: https://github.com/trezor/trezor-firmware/tree/master/python

fametrano:
- Trezor already fixed that in the past; as for digitalbox
- I cannot find the equivalent upstream code to fix: https://github.com/digitalbitbox/mcu/search?l=Python&q=except
- Ledger has not provided feedback yet
- Coldcard rightfully complained about my PR being syntactically wrong (my bad)

Anyway, I would not consider upstream acceptance as a prerequisite: the changeset is small, its rationale is strong, the current situation is very bad as it prevents keyboard interrupts to stop any process involving HWI.

achow101:
ACK 44dabcfa805797bef310471fc566f29ab67bb6d5

TamtamHero:
FYI upstream patch at Ledger has been merged and deployed with btchip-python 0.1.31: 
https://github.com/LedgerHQ/btchip-python/pull/47
Thanks for the PR and sorry for the delay.

PR #374, opened 2020-08-18
Add KeyOriginInfo class to handle derivation paths

achow101:
Instead of passing around key origins (fingerprint and derivation path combos) as lists of ints, put these into a new class `KeyOriginInfo`. This is basically what we do in Bitcoin Core. All of the PSBT derivation path stuff should be operating on `KeyOriginInfo`. This is placed in a new `hwilib/key.py` file along with the `ExtendedKey` class.

Additionally, the `parse_path` function in `trezorlib.tools` has been copied to `key.py` so that we can use it in other devices without requiring `trezorlib`.

### DIFF
--- a/hwilib/devices/bitbox02.py
+++ b/hwilib/devices/bitbox02.py
@@ -10,8 +10,6 @@ from typing import (
     Sequence,
     TypeVar,
 )
-from binascii import unhexlify
-import struct
 import builtins
 import sys
 from functools import wraps
@@ -35,6 +33,9 @@ from ..errors import (
     DEVICE_NOT_INITIALIZED,
     handle_errors,
     common_err_msgs,
+)
+from ..key import (
+    KeyOriginInfo,
 )
 
 import hid  # type: ignore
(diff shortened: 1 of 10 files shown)


hwilib/key.py
```diff
@@ -0,0 +1,158 @@
...
+            d['privkey'] = binascii.hexlify(self.privkey).decode()
+        else:
+            d['pubkey'] = binascii.hexlify(self.pubkey).decode()
+        return d
+
+
+class KeyOriginInfo(object):
+    def __init__(self, fingerprint: bytes, path: List[int]) -> None:
+        self.fingerprint: bytes = fingerprint
+        self.path: List[int] = path
```
benma, line 81: `Sequence` is preferred over `List` as it only supports random access, no mutation.

achow101: Changed.

achow101:
Rebased and added Bitbox02 changes. @benma Can you test those?

hwilib/devices/bitbox02.py
```diff
@@ -299,13 +267,16 @@ def init(self, expect_initialized: bool = True) -> bitbox02.BitBox02:
     def close(self) -> None:
         self.transport.close()
 
-    def get_master_fingerprint_hex(self) -> str:
+    def get_master_fingerprint(self) -> bytes:
```
benma, line 270: What is the point of splitting this into two functions?

benma (approved):
Tested with the BitBox02, tACK. Nice improvement!

benma:
> Rebased and added Bitbox02 changes. @benma Can you test those?

Sorry, this got lost in my inbox. I tested it now and it works.

hwilib/devices/bitbox02.py
achow101: `get_master_fingerprint_hex` is needed to match the api, but it's better to use `get_master_fingerprint` internally.

PR #376, opened 2020-08-18
Derive child keys from an ExtendedKey

achow101:
It might be useful to us to be able to derive further child keys from an xpub. This PR re-introduces the BIP 32 code that was removed in a3cdd53ddaab86cf9b7cb12f73bc99a081b450a6.

Based on #374 for organizational reasons.

### DIFF
--- a/hwilib/devices/coldcard.py
+++ b/hwilib/devices/coldcard.py
@@ -117,8 +117,7 @@ class ColdcardClient(HardwareWalletClient):
         else:
             result = {'xpub': xpub}
         if self.expert:
-            xpub_obj = ExtendedKey()
-            xpub_obj.deserialize(xpub)
+            xpub_obj = ExtendedKey.deserialize(xpub)
             result.update(xpub_obj.get_printable_dict())
         return result
 
--- a/hwilib/devices/digitalbitbox.py
+++ b/hwilib/devices/digitalbitbox.py
@@ -355,8 +355,7 @@ class DigitalbitboxClient(HardwareWalletClient):
         else:
             result = {'xpub': reply['xpub']}
         if self.expert:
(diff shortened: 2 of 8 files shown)


fametrano:
Would you consider using [btclib](https://github.com/btclib-org/btclib) instead? Its BIP32 support is pretty robust and the library has also spawn a proposed addition to the BIP32 specification: https://github.com/bitcoin/bips/pull/921

Moreover, btclib might progressively allow the removal of the base58/bech32/mnemonic codebase/dependency (probably even ecdsa could go), leaving HWI to focus on its core.
[Next release will also include PSBT, but on this @achow101's code might be few steps ahead, at least in term of reliability ;-)]

I know the library is new, but it should be appreciated that there are no alternatives tested so extensively (100% coverage), reproducing results from all major formal and informal sources. It does not hurt that btclib is fully annotated and try to be very pythonic.

Last but not least, any contribution to the btclib codebase would be welcome and quickly processed: the library does need more contributors to really take off.

achow101:
> Would you consider using [btclib](https://github.com/btclib-org/btclib) instead?

I've been avoiding including similar all-in-one libraries in HWI as they contain lots of functionality that isn't useful to us. I don't think we use enough things to warrant the inclusion of such a library.

> Moreover, btclib might progressively allow the removal of the base58/bech32/mnemonic codebase/dependency (probably even ecdsa could go), leaving HWI to focus on its core.

Unfortunately many of the dependencies that we currently include are because of the libraries that we have brought into the codebase. `mnemonic` and `ecdsa` are required because they are used by `trezorlib`. While I'd like to get rid of them, the upstream libraries need to do that before I can.

Base58 and bech32 are small enough modules that I think it's fine to have them.

PR #379, opened 2020-08-20
Update docs regarding Trezor T mixed input support

mmilata:
AFAICT Trezor T does support mixed segwit and non-segwit inputs, there's couple of tests for that: https://github.com/trezor/trezor-firmware/blob/core/v2.3.2/tests/device_tests/test_msg_signtx_mixed_inputs.py


mmilata:
Thanks!

PR #390, opened 2020-09-15
Allow passing None to BitBox2 passphrase

ben-kaufman:
Check for emptiness instead of an empty string when initializing BitBox2.
A very small change but I think that'd be the expected behavior with regards to how others (like Ledger) treat the passphrase.

A separate question, are you planning to make a new release soon with the BitBox2 in it?


achow101:
ACK 6a6c46f6834b4dac5e67d0c92a9ef791619e9dce

> A separate question, are you planning to make a new release soon with the BitBox2 in it?

Yes. There is a potential bug with the Coldcard that I would like to resolve before making the release.

PR #399, opened 2020-11-04
bitbox02: fix enumerate regression

benma:
6a6c46f6834b4dac5e67d0c92a9ef791619e9dce checks that a password must
exist instead of that a password must not exist.

That commit does not describe the reason for the change. Password
cannot be None anyway, as the type of password is `str`, not `Optional[str]`.

### DIFF
--- a/hwilib/devices/bitbox02.py
+++ b/hwilib/devices/bitbox02.py
@@ -214,7 +214,7 @@ class Bitbox02Client(HardwareWalletClient):
         Initializes a new BitBox02 client instance.
         """
         super().__init__(path, password=password, expert=expert)
-        if not password:
+        if password:
             raise BadArgumentError(
                 "The BitBox02 does not accept a passphrase from the host. Please enable the passphrase option and enter the passphrase on the device during unlock."
             )


benma:
cc @ben-kaufman

ben-kaufman (approved):
Right, that looks like a mistake.
ACK b8cf8578836cf92918c619f6f763c69697926810

benma:
@ben-kaufman what was the reason for the original commit? The type prevents `None` anyway.

ben-kaufman:
I think it was just that on Specter we were passing None when the passphrase was empty, and it was strange at first that it failed on BitBox02. Didn't notice the type check actually, bitbox02 seems to be the only one with it. It just seemed an unexpected failure compared to how others handle None passphrase.

I have no idea how come it says `not` though, as even the code we have in Specter for it doesn't have it, so must be some typing mistake on my side or something.

I think it might be worth making the passphrase type optional though, since a change as I was hoping to do in #357 would need it to distinguish between no passphrase passed and the passphrase passed was an empty string.

benma:
Making it explicitly optional makes sense, though empty passphrase and no passphrase is the same thing within bip39.

benma:
@achow101 - please merge this before making another release.

achow101:
ACK b8cf8578836cf92918c619f6f763c69697926810

PR #400, opened 2020-11-09
Add serialization for multisig descriptors

stepansnigirev:
With this PR multisig descriptors can be parsed and serialized as well.


achow101:
The descriptors overhaul done in  #412 added multisig descriptor serialization.

PR #401, opened 2020-11-10
descriptor: fix wrong handling of missing origin in multisig

benma:
Example (second key is missing an origin):

wsh(multi(2,[00000001/48'/0'/0'/2']tpubD6NzVbkrYhZ4WaWSyoBvQwbpLkojyoTZPRsgXELWz3Popb3qkjcJyJUGLnL4qHHoQvao8ESaAstxYSnhyswJ76uZPStJRJCTKvosUCJZL5B/0/0,tpubDFHiBJDeNvqPWNJbzzxqDVXmJZoNn2GEtoVcFhMjXipQiorGUmps3e5ieDGbRrBPTFTh9TXEKJCwbAGW9uZnfrVPbMxxbFohuFzfT6VThty/0/0))

Before, the fingerprint and origin keypath of the second key was the
same as the first one, as the loop parsing it did not reset it.

Now, the fingerprint and origin keypath of the second key is None.

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -242,7 +242,8 @@ def displayaddress(client, path=None, desc=None, sh_wpkh=False, wpkh=False, rede
             redeem_script = format(80 + int(descriptor.multisig_M), 'x')
             xpubs_descriptor = False
             for i in range(0, descriptor.multisig_N):
-                path += descriptor.origin_fingerprint[i] + descriptor.origin_path[i]
+                if descriptor.origin_fingerprint[i] and descriptor.origin_path[i]:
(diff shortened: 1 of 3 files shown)


benma:
cc @ben-kaufman , according to the git history you are the best person to tag :)

achow101:
I have a complete reimplementation of the Descriptor class that should resolve this problem. Currently how descriptors are implemented is incredibly hacky and isn't very extensible.

benma:
I see. I am currently implementing multisig support for the BitBox02 in HWI, making use of the the `descriptor` param in `display_address()`. When do you expect it to land? Should I build on what's on master or on your PR/branch?

PR #404, opened 2020-11-18
bitbox02: unlock during enumerate()

benma:
Make it consistent with other HW wallets. Nunchuk relies on having an
unlocked device after enumerate() finishes.


benma:
cc @hugohn

@achow101 is unlock during enumerate() okay, or should Nunchuk call e.g. `getxpub` (or any subcommand) to invoke the unlock (previous behavior)?

achow101:
Concept ACK. Will review in more detail later.

It would be more preferable for the BitBox02 to only be enumerable after it is unlocked, in the same manner as the Coldcard or Trezor T, but I digress.

hugohn:
Concept ACK.

Agreed with @achow101.

achow101:
ACK e1cc1937179a7fdb8e9807a0ef9e6f9b8ece8df0

PR #408, opened 2020-12-01
Pastable udev Shell script

molnard:
Modified the UDEV rule install script in a way that the user can easily copy/paste as a whole, enter his password, and press enter.


nopara73 (approved):

achow101:
ACK 98967d1c65571a0595d15738522e2f2e169484df

PR #411, opened 2020-12-09
improve code pythonic readability

fametrano:
While at https://github.com/bitcoin-core/HWI/pull/409 and https://github.com/bitcoin-core/HWI/pull/410 I just couldn't resist to try and improve some pythonic readability.

Namely, I 
- converted for loop into call to sum()
- inlined variable that is immediately returned
- lifted return into if
- merged duplicate blocks in conditional, removing redundant conditional
- replaced unused for index ('txout') with underscore
- replaced if statement with if expression
- replaced multiple comparisons of same variable with in operator
- replaced unneeded comprehension with generator
- simplified if expression by using or
- simplified conditional into return statement
- used any() instead of for loop, inverted any/all to simplify comparisons

As I know stylistic PR are not welcome, let me stress here that modern python coding should leverage mypy, black, flake8, pylint, bandit, isort, and sourcery. These tools **combined** provide a powerful contribution, helping your residual bug not to be lost among hundreds of warnings.

This PR includes https://github.com/bitcoin-core/HWI/pull/409 but could be made independent from it; if both PRs are accepted, it should be merged later to separate the functionality change from the cosmetic changes.

If this is rejected... well I got some fun :-) and, actually, the relevant PR is really https://github.com/bitcoin-core/HWI/pull/410

### DIFF
--- a/hwilib/serializations.py
+++ b/hwilib/serializations.py
@@ -19,7 +19,6 @@ from .errors import PSBTSerializationError
 from .key import KeyOriginInfo
 
 import struct
-import binascii
 import hashlib
 import copy
 import base64
@@ -35,6 +34,7 @@ from typing import (
     Tuple,
     TypeVar,
     Callable,
+    Union,
 )
 from typing_extensions import Protocol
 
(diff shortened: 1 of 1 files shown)


achow101:
Closing as up for grabs

PR #415, opened 2021-01-20
Replace --sh_wpkh and --wpkh with --addr-type

achow101:
In `getkeypool` and `displayaddress` where we have `--sh_wpkh` and `--wpkh` to determine the address type, replace those options with a new `--addr-type` option. This allows us to remove mutual exclusion checks for `sh_wpkh` and `wpkh`. It will also be easier to handle new address types in the future.


instagibbs (approved):
utACK 357b59f

wish we had automated tests that properly covered these. Rely on RCs I guess!

PR #417, opened 2021-01-20
ledger: Fix KeyOriginInfo is not subscritable error

achow101:
Fixes a bug in Ledger `sign_tx` implementation which would cause it to error with `{"error": "'KeyOriginInfo' object is not subscriptable", "code": -13}` during change detection.

### DIFF
--- a/hwilib/devices/ledger.py
+++ b/hwilib/devices/ledger.py
@@ -206,12 +206,12 @@ class LedgerClient(HardwareWalletClient):
             # Find which wallet key could be change based on hdsplit: m/.../1/k
             # Wallets shouldn't be sending to change address as user action
             # otherwise this will get confused
-            for pubkey, path in tx.outputs[i_num].hd_keypaths.items():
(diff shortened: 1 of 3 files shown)


hwilib/devices/ledger.py
```diff
@@ -207,11 +207,11 @@ def sign_tx(self, tx):
             # Wallets shouldn't be sending to change address as user action
             # otherwise this will get confused
             for pubkey, path in tx.outputs[i_num].hd_keypaths.items():
```
instagibbs, line 209: good time to rename this `origin` tbh

instagibbs (commented):
looks right, a simple rename probably better though

hwilib/devices/ledger.py

achow101:
I've added a test case for this.

Sjors:
ACK (assuming you get CI to behave)

This is the same fix I came up with independently in #428, which is always a good sign.

PR #418, opened 2021-01-20
Remove keepkey

### LINKED ISSUE #386: [RFC] Drop Keepkey
Should we drop support for the Keepkey?

Keepkey is a clone of the Trezor, so we currently just reuse the Trezor implementation. As such, `trezorlib` and the `TrezorClient` needs a couple of modifications to support the Keepkey. However these changes do make updating Trezor things harder to ensure that Keepkey still work. Furthermore, doing so also means that we lose Trezor functionality.

The specific incompatibility between the Trezor and Keepkey that causes the issues is that the `Features` message differs. Using a stripped down `Features` message to support will result in lost functionality on the Trezor. I've tried a couple of other things to try to swap out the the module implementing the message either don't work, or involve hijacking some very internal things in Python that we probably shouldn't be doing. In the latter case, while it may work for the CLI tool, it's possible that library users won't work.

Instead of doing some weird Python thing, we could instead copy the trezorlib implementation and just make it Keepkey specific (note that the actual keepkeylib is not usable for us as it is based on an older trezorlib that phones home and would require even more significant modification to fit our standards). So Keepkey would be implemented as a separate device instead of as a Trezor device. However this will mean that there's a lot of duplicated code and we still have to go through the work of patching trezorlib to work with the keepkey every time it is updated.

But there's also the question of whether it is even worthwhile to keep the Keepkey. It does not seem like the Keepkey is actually used. So in that case, we could probably just drop it entirely without any adverse effect on the users of HWI.

achow101:
Closes #386


nopara73:
We never succeeded to make KeepKey work properly in Wasabi, so we wouldn't mind if you'd drop it.

ben-kaufman:
We do have KeepKey support on Specter. I don't know if or how many of our users use it though, so I'll have to try and survey.
We could probably maintain support for it in Specter codebase itself if we do need it, although it'll just be easier to have it in HWI.

achow101:
The primary reason for removing the Keepkey is that it makes it harder to maintain the Trezor stuff. Since Keepkey is a clone of the Trezor 1, the way support for it was implemented by reusing the Trezor implementation (by directly importing and subclassing `TrezorClient`) and trezorlib with some modifications. However Keepkey's communication protocol has slightly diverged from Trezor's so we lose a bit of Trezor functionality and updating trezorlib is quite a pain.

An alternative to removing the Keepkey is to duplicate the Trezor code and trezorlib, but that seems a bit wasteful. It also doesn't remove the need to modify trezorlib to handle the Keepkey. We could use keepkeylib instead, but it's based on an old trezorlib and contains some undesirable behavior, such as defaulting to phoning home for UTXOs, so that would still need modification to remove that.

instagibbs:
Could we put it behind a deprecation flag or something for a release cycle? If no complaints show up, remove?

achow101:
> Could we put it behind a deprecation flag or something for a release cycle? If no complaints show up, remove?

Since I am trying to do SemVer correctly, that seems like it would require 2 major releases to do that since it is ostensibly an API breaking change. The big promise of 2.0 is that we can break the API.

instagibbs:
Ok for 2.0 this is probably fine.

gruve-p:
I am using keepkey with Specter and Sparrow
Instead of removing Keepkey I would recommend to copy the trezorlib implementation and just make it Keepkey specific similar as Electrum has done (so Keepkey would be implemented as a separate device instead of as a Trezor device. And yes this will mean that there's a lot of duplicated code)

Maybe @keepkeyjoncan help/assist you through the work of patching trezorlib to work with the keepkey every time it is updated.

achow101:
Seems like we should keep Keepkey support.

PR #419, opened 2021-01-22
Expand CI jobs matrix and switch from Travis to Cirrus

achow101:
In Travis, we have a simple jobs matrix which only tests each interface. All devices are run for an interface in a single job. This has caused problems for us as any failure in a specific device fails the entire test and does not build or run the test for the other devices. This results in a single failure not informing us whether a change made to another device causes the tests for that device to not be run. Instead we can expand this jobs matrix to test each device and interface combination individually.

Additionally we switch from Travis to Cirrus as Cirrus appears to be more responsive and easier to debug. Since Cirrus uses docker containers, it is possible to replicate the CI environment locally. Additionally, the docker images used can be specified by us and can even use dockerfiles present in the repo. This allows us to have absolute control over the testing environment which makes it much easier to get the tests to work.

In order to facilitate the expanded jobs matrix, `setup_environment.sh` is modified to only build each simulator/emulator or bitcoind when given the correct command line arguments. These are `--trezor-1`, `--trezor-t`, `--coldcard`, `--ledger`, `--keepkey`, `--bitbox01`. Also `run_tests.py` is modified to have the same arguments names for devices, namely `--trezor` is now `--trezor-1` and `--bitbox` is `--bitbox02`. It also has a new `--device-only` option to disable the non-device unit tests.

This CI change reveals that the Coldcard and Trezor T tests and implementations are currently broken. It also revealed that our Ledger tests were slightly broken, but this was more easily resolved and has been fixed. For the Ledger, we switch to using Speculos' built in button automation instead of having a patch that does the button presses for us. Updating the configuration for that to include some additional text that is being displayed resolves the failures there. However, for Coldcard and Trezor T, the changes involved will be more complicated and so are left for a future PR.

For the final jobs matrix, we first have 3 jobs which don't need simulators/emulators or bitcoind:
* Linter
* Type checker
* Non-device unit tests

Then we have a set of builder jobs. These don't test anything but instead build the simulators, emulators, bitcoind, and release distribution. The results of these builders are cached and passed down to the test jobs:
* bitcoind builder
* Builder for each device simulator/emulator
* Linux and python distribution builder
* Windows distribution builder (this is to test that our Wine build script works. The results are not used or tested.)

Lastly we have the test jobs. These take the cached bitcoind and cached device simulator (only for the device being tested) and runs the device tests for each interface. As some interfaces require the linux and python distributions, the cached results from those are also used in those jobs:
* For each device
  * For each interface in library, cli, stdin, wheel installed cli, sdist installed cli, linux binary dist
    * Device test job


achow101:
Merging as expected tests are failing.

PR #421, opened 2021-01-26
Allow python 3.9, update dependencies, and bump version to 2.0.0-dev

achow101:
This is general pyproject.toml maintenance. Python 3.9 is the latest python version, so we should allow it. Ideally, we would not set a python version upper bound, but the pyside2 package sets a python version upper bound, so poetry requires us to also set that.

Additionally, the version number in the master branch was super out of date. 2.0.0-dev is more accurate - we are developing for 2.0.0 in master currently.

Lastly, all of the dependencies in `poetry.lock` are updated.


laanwj (approved):
Code review (except poetry.lock) ACK 225d10e75ab8fa2f5e95e68b7997c1526f5aee5a

PR #423, opened 2021-01-27
Refactor client.display_address into single sig and multisig variants

achow101:
Instead of having `display_address` handle both single sig and multisig address display, it is refactored into `display_singlesig_address` to display single sig addresses, and `display_multisig_address` to display multisig addresses. Additionally the `redeem_script`
parameter is removed as descriptors provide everything we need.

Supersedes #416. I think this approach is nicer to work with.

Based on #415

### DIFF
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -132,6 +132,27 @@ class PubkeyProvider(object):
             path = path[:-1] + str(pos)
         return path
 
+    def get_full_derivation_int_list(self, pos: int) -> List[int]:
+        """
+        Returns the full derivation path as an integer list at the given position.
+        Includes the origin and master key fingerprint as an int
+        """
+        path: List[int] = self.origin.get_full_int_list() if self.origin is not None else []
+        if self.deriv_path is not None:
+            der_split = self.deriv_path.split("/")
+            for p in der_split:
+                if not p:
+                    continue
+                if p == "*":
+                    i = pos
+                elif p[-1] in "'phHP":
+                    assert len(p) >= 2
+                    i = int(p[:-1]) | 0x80000000
+                else:
+                    i = int(p)
+                path.append(i)
+        return path
+
     def __lt__(self, other: 'PubkeyProvider') -> bool:
         return self.pubkey < other.pubkey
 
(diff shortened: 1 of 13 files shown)


instagibbs:
rebase plz

achow101:
Rebased

hwilib/descriptor.py
```diff
@@ -68,6 +68,16 @@ def __init__(
...
         self.deriv_path = deriv_path
 
+        # Make ExtendedKey from pubkey if it isn't hex
+        self.extkey = None
+        try:
+            unhexlify(self.pubkey)
+            # Is hex, normal pubkey
+        except Exception:
+            # Not hex, maybe xpub
+            if self.pubkey[0:4] == "xpub" or self.pubkey[0:4] == "tpub":
```
instagibbs, line 83: `if self.pubkey[0:4] in ["xpub", "tpub"]:`

hwilib/descriptor.py
```diff
@@ -172,6 +204,17 @@ def __init__(
     def to_string_no_checksum(self) -> str:
         return "{}({},{})".format(self.name, self.thresh, ",".join([p.to_string() for p in self.pubkeys]))
 
+    def expand(self, pos: int) -> "ExpandedScripts":
+        m = (self.thresh + 0x50).to_bytes(1, "big") if self.thresh > 0 else b"\x00"
```
instagibbs, line 229: might want to throw when self.thresh is larger than 16 since we're dealing with single pushes only for now

instagibbs (approved):
utACK

hwilib/descriptor.py
benma: It would be better to move the check inside of ExtendedKey.deserialize() and throw an Exception if the bytes are not a valid xpub (including checksum).

Reason: it is better to concentrate xpub parsing in one location instead of many locations, making future maintenance easier.

achow101: Changed to throw an exception in `ExtendedKey.deserialize`.

hwilib/descriptor.py
achow101: Changed to handle thresholds larger than 16

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/423/commits/58cb94aa05b9d071f632d89f36c68f882deafb2c

test failures appear to be unrelated timeouts

achow101:
Timeouts were related, fixed now.

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/423/commits/7650a57a1ab057cfcd1b2cc2b878376ce8844929

PR #424, opened 2021-01-28
Fix trezor T multisig and tests

achow101:
Fixes the Trezor T multisig tests. Part of this requires fixes to `displayaddress` and the Trezor implementation. The majority of the changes here are to make the multisig derivation paths conform to [Trezor's purpose 48](https://github.com/trezor/trezor-firmware/blob/master/docs/misc/purpose48.md).

Also fixes the Coldcard multisig tests. Coldcard now disallows the same fingerprint being used multiple times in a multisig. We add a patch which enables this behavior so our tests work. Additionally the patch which sets the default multisigs is updated to use the new derivation paths that we need to use.


achow101:
Merging as all of the tests that this is supposed to fix have been fixed.

PR #425, opened 2021-01-28
Remove poetry version restriction in build_*.sh

achow101:
I found a workaround to this problem by setting a UTF-8 locale, so the version restriction is no longer necessary.

Should also fix the wine_builder CI job.


achow101:
Merging as wine_builder now passes.

PR #426, opened 2021-01-28
bitbox02: update bitbox02 to 5.1.0

achow101:
Updated pyproject.toml and setup.py, then: `poetry update bitbox02 --lock`.

Also fixes the failing type check CI job.


achow101:
Merging as type check now passes

PR #427, opened 2021-01-28
Remove hidapi.so strip during build_bin.sh

achow101:
Poetry >=1.1.0 seems to be building the cython .so files
deterministically now so there is no need for us to make it
deterministic by stripping them. Additionally, stripping them when
poetry >=1.1.0 did the build is causing the resulting binaries to fail,
so this should fix that too.


achow101:
Merging as CI is all green(!)

PR #431, opened 2021-01-29
tests: refactor Trezor model handling

prusnak:


test/test_trezor.py
```diff
@@ -359,10 +357,20 @@ def trezor_test_suite(emulator, rpc, userpass, interface, model_t=False):
     parser.add_argument('emulator', help='Path to the Trezor emulator')
     parser.add_argument('bitcoind', help='Path to bitcoind binary')
     parser.add_argument('--interface', help='Which interface to send commands over', choices=['library', 'cli', 'bindist'], default='library')
+    parser.add_argument('--model_1', help='The emulator is for the Trezor One', action='store_true')
```
achow101, line 360: I think it would be better to have this and `--model_t` as a mutually exclusive group. Both of these options can then be using `action='store_const` and `dest=model`. This would remove the `if` block below.

e.g. add_argument('--model_1', help='The emulator is for the Trezor One', action='store_const', const='1', dest='model')

prusnak: Great! Didn't know about `store_const`. Seems definitively better.

Addressed in the fixup commit e03a1588efba3d90b2d690ceb7cfa1706d9ffc1d

test/test_trezor.py
```diff
@@ -359,10 +357,11 @@ def trezor_test_suite(emulator, rpc, userpass, interface, model_t=False):
     parser.add_argument('emulator', help='Path to the Trezor emulator')
     parser.add_argument('bitcoind', help='Path to bitcoind binary')
     parser.add_argument('--interface', help='Which interface to send commands over', choices=['library', 'cli', 'bindist'], default='library')
-    parser.add_argument('--model_t', help='The emulator is for the Trezor T', action='store_true')
+    parser.add_argument('--model_1', help='The emulator is for the Trezor One', action='store_const', const='1', dest='model')
+    parser.add_argument('--model_t', help='The emulator is for the Trezor T', action='store_const', const='t', dest='model')
```
achow101, line 361: Please use a [mutually exclusive group](https://docs.python.org/3/library/argparse.html#argument-groups).

```
group = parser.add_argument_group()
group.add_argument('--model_1', help='The emulator is for the Trezor One', action='store_const', const='1', dest='model')
group.add_argument('--model_t', help='The emulator is for the Trezor T', action='store_const', const='t', dest='model')
```

prusnak: Addressed in 267f2af7304354a90899c3cc517abf7fedc8891e

achow101:
ACK 267f2af7304354a90899c3cc517abf7fedc8891e

PR #435, opened 2021-02-01
Add some dockerfile steps for local testing

instagibbs:
I'd really like local testing to be easier. Ideally we'd use docker's layer/caching to take care of everything for us rather than rely on additional layers of caching, making it efficient to test hwi changes.

For now, I added some steps that can be uncommented out for quick local testing.

As a follow-up I'm looking at having these lines run in the base image, but have to make sure they don't conflict with other things such as the various simulator patches(coldcard only for now), and making sure the additional build time is worth it from a CI perspective since a number of simulator building steps are re-done for each cirrus job.

### DIFF
--- a/ci/cirrus.Dockerfile
+++ b/ci/cirrus.Dockerfile
@@ -33,7 +33,47 @@ RUN apt-get install -y \
     libboost-thread-dev \
     protobuf-compiler \
     cython3
+
 RUN pip install poetry flake8
+
+####################
+# Local build/test steps
+# -----------------
+# To install all simulators/tests locally, uncomment the block below,
+# then build the docker image and interactively run the tests
+# as needed.
+# e.g.,
+# docker build -f ci/cirrus.Dockerfile -t hwi_test .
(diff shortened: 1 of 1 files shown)


test/setup_environment.sh
```diff
@@ -127,6 +127,8 @@ if [[ -n ${build_coldcard} ]]; then
         fi
     fi
     # Apply patch to make simulator work in linux environments
+    git config --global user.email "you@example.com"
```
instagibbs, line 130: probably a better way somehow but it didn't like that I had nothing configured locally on docker

achow101: It's sufficient to have an environment variable named `EMAIL` set to the email. You can add that to the dockerfile.

instagibbs: done, and added some more notes to tell someone how to run it locally

achow101:
ACK 35ec4dec1d49dfe94d1b7c5aa4a1bc892a1cdeee

PR #438, opened 2021-02-01
Add label field to enumerate

prusnak:
I don't have other devices except Trezor, so maybe others can help with how to retrieve labels for the rest.


instagibbs:
sorry can the OP motivate what this is a bit more? What are we solving, example input etc.

prusnak:
Apps integrating HW wallet via HWI could show the device label in their UI.

instagibbs:
ah, so like "Bob's Trezor".

prusnak:
> ah, so like "Bob's Trezor".

Yes

achow101:
ACK 1168d6a5304eecf2243cdfda0d0abd361395d2d4

achow101:
For the other devices, you can do `d_data['label'] = None` and it will be output as `"label": null`

PR #440, opened 2021-02-02
[WIP] User docker to setup/cache simulators

instagibbs:
Based on https://github.com/bitcoin-core/HWI/pull/435

This is definitely going to break for now, at least for coldcard due to the multisig patch being applied every since invocation of `setup_environment.sh`.

### DIFF
--- a/ci/cirrus.Dockerfile
+++ b/ci/cirrus.Dockerfile
@@ -33,7 +33,47 @@ RUN apt-get install -y \
     libboost-thread-dev \
     protobuf-compiler \
     cython3
+
 RUN pip install poetry flake8
+
+####################
+# Local build/test steps
+# -----------------
+# To install all simulators/tests locally, uncomment the block below,
+# then build the docker image and interactively run the tests
+# as needed.
+# e.g.,
+# docker build -f ci/cirrus.Dockerfile -t hwi_test .
+# docker run -it --entrypoint /bin/bash hwi_tst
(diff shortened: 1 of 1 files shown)


instagibbs:
I do wonder if the dockerfile should eat and log simulator build errors, this would allow the dockerfile to finish building and be shuttled off to cirrus jobs for partial testing. This would be useful in case simulator repo(s) get in a bad state, and still allows for easy local testing.

@achow101 ?

instagibbs:
Talking offline achow notes that if we can simply have a single build flat for the dockerfile we can have cirrus build N different docker images instead, one for each device.

test/setup/setup_trezor_t.sh
```diff
@@ -0,0 +1,41 @@
...
+# Remove .venv so that poetry can symlink everything correctly
+find . -type d -name ".venv" -exec rm -rf {} +
+
+# Build trezor t emulator. This is pretty fast, so rebuilding every time is ok
+# But there should be some caching that makes this faster
+poetry install
+cd core
+poetry run make build_unix
+# Delete any emulator.img file
+find . -name "trezor.flash" -exec rm {} \;
```
prusnak, line 40:
```suggestion
# Delete emulator files
find . -name "trezor.flash" -exec rm {} \;
find . -name "trezor.sdcard" -exec rm {} \;
```

prusnak (changes requested):

achow101:
is this still relevant?

PR #437, opened 2021-02-01
PSBT: also parse and serialize PSBT_GLOBAL_XPUB

benma:
Needed for BitBox02 to check and perform the xpubs registration.

Extracted from #407 according to this comment: https://github.com/bitcoin-core/HWI/pull/407#issuecomment-768513068

PR #407, opened 2020-11-25
Bitbox02 multisig

benma:
This adds support for multisig `display_address` and `sign_tx` for segwit multisig (p2wsh and p2wsh-p2sh).

This is in high demand by our users and enables multisig support in downstream wallets such as Nunchuk, Specter, Sparrow, etc.

Firmware v9.3.0 is required.

PR #443, opened 2021-02-02
BitBox02 multisig signing

benma:
Extracted from #407 according to this comment: https://github.com/bitcoin-core/HWI/pull/407#issuecomment-768513068

Builds on https://github.com/bitcoin-core/HWI/pull/437 as we need access to the PSBT xpubs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ Please also see [docs](docs/) for additional information about each device.
 | P2PKH Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
 | P2SH-P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | P2WPKH Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| P2SH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
-| P2SH-P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | N/A | Yes | Yes |
+| P2SH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| P2SH-P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | P2WSH Multisig Inputs | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | Bare Multisig Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
 | Arbitrary scriptPubKey Inputs | Yes | Yes | N/A | N/A | Yes | N/A | N/A | N/A |
--- a/docs/bitbox02.md
+++ b/docs/bitbox02.md
@@ -13,20 +13,19 @@ Current implemented commands are:
 * `backup`
 * `togglepassphrase`
 
-Multisig (P2WSH only) is supported by the BitBox02, but is not ingerated into HWI yet. Coming
-soon^{tm}.
-
 # Usage Notes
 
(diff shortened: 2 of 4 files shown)


instagibbs:
concept ACK, since there's no emulator let's try to get #437 in with a test first

benma:
Rebased on #437 and master.

benma:
Rebased after #437 was merged.

instagibbs:
@achow101 failure seems unrelated but new to me:
```
pip install $(find dist -name "*.tar.gz" -a -not -name "*amd64*")
find: ‘dist’: No such file or directory
ERROR: You must give at least one requirement to install (see "pip help install")
����
```

instagibbs (commented):
code changes look fine enough, going to test

achow101:
The failure is the caching problem we were occasionally running into. The results from dist_builder aren't making it to that test machine which needs those results. But it's unrelated anyways.

instagibbs:
Kind of annoying to test with no utils adding xpubs to PSBTs on my end. @ben-kaufman can I get a speculative specter branch that does this(doesn't seem to right now)?

benma:
@instagibbs in the meantime you can also setup a multisig in Electrum and export PSBT's from there.

I can also add a seed and some PSBT fixtures for easier testing - could also be helpful for the future.

instagibbs:
A simple tool to graft xpub strings into the psbt in python would be sufficient for this testing for now. Also could be useful in test suite in lieu of support in Core for updating these fields.

ben-kaufman:
Tested, works. Haven't looked deep into the code but looks good from what I saw. I've created this branch on Specter where it allows selecting Bitbox02 for multisig, and also fills all the PSBT data in the Raw PSBT field under transaction details. So you can test with this branch by copying a PSBT from that Specter branch and signing with HWI of this PR.
https://github.com/ben-kaufman/specter-desktop/tree/test-bitbox02

instagibbs:
Does bitbox not support testnet/regtest derivation paths? Docs seem to suggest it doesn't, and I'm getting a very generic error:

```
{"error": "Input error: error code: 103, message: generic error. A keypath might be invalid. Supported keypaths are: m/49'/0'/<account'> for p2wpkh-p2sh; m/84'/0'/<account'> for p2wpkh; m/48'/0'/<account'>/2' for p2wsh multisig; m/48'/0'/<account'>/1' for p2wsh-p2sh multisig; m/48'/0'/<account'>' for any supported multisig; account can be between 0' and 99'; For address keypaths, append /0/<address index> for a receive and /1/<change index> for a change address.", "code": -9}
```

ben-kaufman:
It does I think @instagibbs 
At least I made a successful regtest transaction with it... Maybe you need to update the firmware?

instagibbs:
I guess it stands to reason that it's handing out keys for those paths, why wouldn't it sign? I'll try from scratch again.

instagibbs:
Ahhhhhhh, you need to pass `--testnet` otherwise it's checking against mainnet derivation paths.

This should really be documented or handled better.

instagibbs:
ACK https://github.com/bitcoin-core/HWI/pull/443/commits/55ae05ae5d3a72673d98f4cd5b5ffd6ab8ac283f

Would appreciate some way off better communicating to users that `--testnet` is required(or illegal) with respect to signing which I didn't expect.

benma:
> Would appreciate some way off better communicating to users that `--testnet` is required(or illegal) with respect to signing which I didn't expect.

This documentation already exists:

https://github.com/bitcoin-core/HWI/blob/master/docs/bitbox02.md#strict-keypaths

achow101:
Code Review ACK 55ae05ae5d3a72673d98f4cd5b5ffd6ab8ac283f

instagibbs:
@benma ok I didn't find the text flow informative enough, but also too lazy to suggest an improvement :)

PR #444, opened 2021-02-03
Update local trezorlib

achow101:
Updates the local trezorlib to the latest from their monorepo.

In the interest of making it easier to update this in the future, a significant number of changes to the way that we handle. Since trezorlib has added some functions that are useful for us during initializing, it is unnecessary for us to modify `init_device`. Instead we can do all of the logic that we were doing there previously and do it in another helper function in our `TrezorClient` class.

Additionally, instead of modifying the messages to allow for the Keepkey to work, I've implemented a way for the `KeepkeyClient` to override the message to type mapping used. This allows us to have separate implementations of Keepkey specific messages (added to `hwilib/devices/keepkey.py`) that don't conflict with the Trezor messages.

Some of the changes made to trezorlib will be upstreamed so that in the future, we can have a more minimal diff (or no diff at all).


prusnak:
Thanks achow101 for doing this!

@matejcik could you please have a look at the changes to:
- see if you can suggest some improvements
- understand how HWI is using our library, so we can make their way easier in the future by maybe doing some changes in upstream?

hwilib/devices/trezorlib/transport/webusb.py
```diff
@@ -105,19 +108,16 @@ def __init__(
...
+    def enumerate(cls, usb_reset=False, usb_ids: Optional[Set[Tuple[int, int]]] = TREZORS) -> Iterable["WebUsbTransport"]:
         if cls.context is None:
             cls.context = usb1.USBContext()
             cls.context.open()
             atexit.register(cls.context.close)
         devices = []
         for dev in cls.context.getDeviceIterator(skip_on_error=True):
             usb_id = (dev.getVendorID(), dev.getProductID())
-            if usb_id not in TREZORS:
+            if usb_ids is not None and usb_id not in usb_ids:
```
matejcik, line 120: using `usb_ids=None` this way will let you enumerate **all** vendor-class USB devices present on the system. Some of them might not be user accessible. This crashes your enumeration on my machine.

hwilib/devices/trezor.py
```diff
@@ -117,39 +117,108 @@ def func(*args, **kwargs):
...
+    def disallow_passphrase(self):
+        self.return_passphrase = False
+
+    def get_passphrase(self):
+        if self.return_passphrase:
+            return self.passphrase
+        raise ValueError('Passphrase from Host is not allowed for Trezor T')
+
+
+def get_path_transport(path: str):
```
matejcik, line 198: any reason why this exists, as opposed to `trezorlib.transport.get_transport` ?

matejcik: ...right, because there's no pass-through in `get_transport` that would allow finding non-Trezor USB IDs. hence also the `usb_ids=None` thing

achow101: I actually did this to avoid having to remove the `BridgeTransport` stuff from `all_transports`.

matejcik: that's easier to accomplish by setting BridgeTransport.ENABLED = False

achow101: `trezorlib/transport/bridge.py` is specifically excluded from the local copy because it requires `requests`. Including it either requires us to add `requests` as a dependency, or modify the file to not do that. I do not want to do either of those things.

hwilib/devices/trezorlib/transport/webusb.py
achow101: I've pushed a change to drop that.

hwilib/devices/trezor.py
matejcik: so i'm wondering, do you intend to include `trezorlib` as a dependency once your changes are in, or are you still going to vendor a copy? because obviously including it as a dependency would pull in a lot of things like requests

achow101: I'm still planning on vendoring a copy, however I would greatly prefer if all I had to do was copy over a specific set of files and not have to make changes to them.

PR #448, opened 2021-02-05
Bump Python patch version to 3.6.12

Sjors:
It's a pain on macOS Big Sur to run older patch versions, because PyEnv can't install them anymore.


achow101:
It looks like the wine_builder job is failing since python.org doesn't upload windows installer binaries for releases in the security fix only part of the lifecycle.

Sjors:
Maybe I should keep the original windows version?

achow101:
> Maybe I should keep the original windows version?

Yes

achow101:
ACK c8fd9fecaf3a25e71b355d567019e7705f7b6102

PR #451, opened 2021-02-09
Add --signet as alias for --testnet

Sjors:
I plan to modify https://github.com/bitcoin/bitcoin/pull/16546 to use `--signet`.


prusnak:
It might have been cleaner to introduce another option and then use `client.is_testnet = args.testnet or args.signet` below on line 268. But that's just my two satoshis.

achow101:
At this rate, we should just do `--chain <name>`. I want to avoid a repeat of the address type thing. We probably want to support regtest as well and that will actually need changes because the bech32 hrp is different.

Sjors:
That works for me, but is probably above my paygrade :-)

prusnak:
I created an alternative PR which implements `--chain` here: https://github.com/bitcoin-core/HWI/pull/452

PR #452, opened 2021-02-09
Introduce --chain command line argument (replacing --testnet)

prusnak:
Alternative to https://github.com/bitcoin-core/HWI/pull/451

This nukes `--testnet` argument. If we want to keep that, I'll reintroduce it in another commit. Just let me know. I just think it's fine to get rid of that now, since we are breaking the API anyway in HWI 2.0.

Also we should get rid of `client.is_testnet` across the whole codebase and check against the `client.chain`, but I'd prefer to leave that for later PR.


Sjors:
Concept ACK

If it's not too much trouble, it's nice to keep `--testnet` around.

Bitcoin Core uses `main` instead of `mainnet` and `test` instead of `testnet`.
Schermafbeelding 2021-02-09 om 20 00 05

prusnak:
> Bitcoin Core uses `main` instead of `mainnet` and `test` instead of `testnet`.

Addressed in 016d278345d52444a0a376be3806b6de50c15621

Sjors:
Looks like this breaks some tests...

instagibbs:
concept ACK, I'm ok with a clean break here since we're already getting rid of a lot of similar exclusive arguments like address types

achow101:
ACK a7d14ba40611749685901581ac9feddb8e2f8ecd

PR #453, opened 2021-02-10
Replace is_testnet by chain where possible (except key.py)

prusnak:
Continutation of https://github.com/bitcoin-core/HWI/pull/452

Replace `is_testnet` with `chain` everywhere except `key.py`

### DIFF
--- a/hwilib/cli.py
+++ b/hwilib/cli.py
@@ -20,6 +20,7 @@ from .commands import (
     wipe_device,
     install_udev_rules,
 )
+from .common import Chain
 from .errors import (
     handle_errors,
     DEVICE_CONN_ERROR,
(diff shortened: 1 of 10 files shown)


achow101:
I would prefer for this to be an enum rather than a string.

prusnak:
> I would prefer for this to be an enum rather than a string.

Enum introduced in 0dba2b63a637955e14ac8ef4844f9ad3cae0191c

hwilib/cli.py
```diff
@@ -117,14 +118,12 @@ def error(self, message):
...
 def process_commands(cli_args):
-    CHAINS = ['main', 'test', 'regtest', 'signet']
-
     parser = HWIArgumentParser(description='Hardware Wallet Interface, version {}.\nAccess and send commands to a hardware wallet device. Responses are in JSON format.'.format(__version__))
     parser.add_argument('--device-path', '-d', help='Specify the device path of the device to connect to')
     parser.add_argument('--device-type', '-t', help='Specify the type of device that will be connected. If `--device-path` not given, the first device of this type enumerated is used.')
     parser.add_argument('--password', '-p', help='Device password if it has one (e.g. DigitalBitbox)', default='')
     parser.add_argument('--stdinpass', help='Enter the device password on the command line', action='store_true')
-    parser.add_argument('--chain', help='Select chain to work with ({})'.format(', '.join(CHAINS)), default='main', choices=CHAINS)
+    parser.add_argument('--chain', help='Select chain to work with ({})'.format(', '.join(Chain.names())), default='main', choices=Chain.names())
```
achow101, line 126: It's unnecessary to list out the names again here. The `choices` option will do that automatically.

This could also do the same trick that `AddressType` does to do the automatic conversion from string to enum type without needing `Chain.by_name` below.

prusnak: Addressed in the comment below

achow101:
ACK 3e737644213657fd716fa9bcca495610654a346d

PR #449, opened 2021-02-05
Change HardwareWalletClient functions to return objects instead of dictioanries

achow101:
For the functions in HardwareWalletClient, it doesn't quite make sense to have it return the dictionary that for cli output. Instead they should return the correct objects and types, which the caller can convert to the dictionary for cli output.

PR #456, opened 2021-02-11
Type check (almost) everything

achow101:
Adds type hints and enables type checking for almost everything. All commands, cli functions, and device implementations now have full type hints and will be checked by the type checker CI job.

The only things missing type hints are the gui stuff (autogenerated ui files don't have type hints) and all local library copies. Even though trezorlib has type hints, some of it is wrong, and some is missing, so it is being ignored for now. The btchip and ckcc lack type hints.

Based on #449 as it adds type hints for a lot of stuff too.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -128,14 +128,26 @@ basic_test_task:
     - name: Type Check
       type_check_script: >
         poetry run
-        mypy --implicit-reexport --strict
+        mypy
+        hwi.py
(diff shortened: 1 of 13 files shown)


instagibbs:
typing good ACK https://github.com/bitcoin-core/HWI/pull/456/commits/ad1146dfa91d83e16bc3ef461281ea23d27a4262

PR #457, opened 2021-02-13
ledger: Ensure signed message signature is correct length

achow101:
Ledger returns a DER encoded signature for signed messages. As such, it is possible that the `r` and `s` values could be shorter than 32 bytes. This would cause us to incorrectly serialize the compact signature used for signed messages. This PR fixes this issue by interpreting the `r` and `s` values as ints and then re-serializing them into fixed 32 byte ints.

Also added a test case. With the seed that speculos uses, the message `285`, when signed with the path `m/44'/1'/0'/0/0`, the `s` value is 31 bytes rather than 32 which runs into this issue.

Electrum had a similar issue as well: https://github.com/spesmilo/electrum/pull/7004

### DIFF
--- a/hwilib/devices/ledger.py
+++ b/hwilib/devices/ledger.py
@@ -339,15 +339,10 @@ class LedgerClient(HardwareWalletClient):
 
         # Make signature into standard bitcoin format
         rLength = signature[3]
-        r = signature[4: 4 + rLength]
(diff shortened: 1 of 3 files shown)


instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/457/commits/969e5fd740041fb210961d4f18368c5949c4a851

PR #458, opened 2021-02-18
Make some modules internal and private

achow101:
Makes `cli.py`, `gui.py`, `base58.py`, and `bech32.py` private modules by prefixing them with an underscore. These modules do not need to be public as they are only used internally by HWI.


instagibbs:
looks fine enough, what problem does this solve? Want to make sure people don't rely on it externally moving forward?

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/458/commits/68996ab959799f569a653d26fdadf040578d0133

achow101:
> Want to make sure people don't rely on it externally moving forward?

Yes

PR #460, opened 2021-02-19
Handle all fingerprints as bytes

achow101:
Instead of sometimes handling the fingerprint as a hex string, and other times as bytes (e.g. in KeyOriginInfo), always handle the fingerprint as bytes. `enumerate` still returns it as a hex string since that is intended for output to the caller. The one exception to this rule is `find_device` which takes the fingerprint as a hex string, and matches it against fingerprints provided by `enumerate`.


instagibbs:
lint fail

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/460/commits/0d4b807c9c2629eff63b42d8a70431235e926e99

Looks like you got all the usage of hex-encoded pubkeys removed except where it makes sense(presentation to user as output)

PR #463, opened 2021-02-22
make more distinction between Yes and N/A in the readme table

prusnak:
... by using checkmark ✓ and cross ✗

The original formatting was making the table quite hard to grasp as there were so many letters present on screen at once.

This is how the rendered readme looks like: https://github.com/prusnak/HWI/tree/readme-table#device-support

We might even consider using ❌ and ✅ emojis, but these might not be as widely supported (they are in the Unicode 6.0 standard from 2010). The characters I used are part of Unicode 1.1 from 1993.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,29 +90,33 @@ The below table lists what devices and features are supported for each device.
 
 Please also see [docs](docs/) for additional information about each device.
 
+* `✓` - supported by the firmware and implemented in HWI
+* `✗` - supported by the firmware and not implemented in HWI
+* `-` - not supported by the firmware
+
 | Feature \ Device | Ledger Nano X | Ledger Nano S | Trezor One | Trezor Model T | BitBox01 | BitBox02 | KeepKey | Coldcard |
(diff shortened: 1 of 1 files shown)


achow101:
The table actual shows a tri-state. `Yes` means "supported by firmware and implemented in HWI". `No` means "supported by firmware and not implemented by HWI" (we don't have any on this table AFAICT). `N/A` means "not supported by the firmware". So just ✓ and ✗ is not sufficient.

A previous PR which implemented something similar was closed because of concerns with how this would be displayed in text editors, although I think that used Github's emojis rather than unicode. I think most terminals and text editors support unicode so this probably isn't an issue, especially with characters that are part of a very old unicode version.

prusnak:
Right. There is no "No" in the table. 

Should we use the cross "✗" for No. And dash (-) for N/A then?

(I'll add the legend under the table).

prusnak:
Right. There is no "No" in the table. 

I updated the PR to include the legend and replaced cross with dash for N/A.

achow101:
ACK 40c6887af5816e61ab6856af5099f3898c63eacc

PR #464, opened 2021-02-23
Sphinx docs

achow101:
Adds documentation that can be generated using [sphinx](https://www.sphinx-doc.org/en/master/). The documentation is moved almost entirely to docstrings for each function and class. Additionally all of our existing documentation is reformatted to use sphinx's rst format.

Once this is merged, I will enable automatic generation on readthedocs.org for this repo.


laanwj:
Didn't read the entire diff (checking the conversion from md to rst in detail) but concept ACK, this would be useful.

PR #470, opened 2021-02-26
Rename addrtype enum values to remove misnomer

achow101:
As noted previously, the current `AddrType` enums are a misnomer for multisig addresses, and in general, things that are not single keys. This PR fixes that by renaming them. PKH is now LEGACY, SH_WPKH is SH_WIT_V0, and WPKH is WIT_V0.

I decided to go wit_v0 to be more explicit about the segwit version. This will let us unambiguously have future address types that are other segwit versions.

Additionally this PR makes WIT_V0 the default address type instead of LEGACY.

### DIFF
--- a/docs/examples/bitcoin-core-usage.rst
+++ b/docs/examples/bitcoin-core-usage.rst
@@ -52,7 +52,7 @@ specified, both receiving and change address descriptors are generated.
 
 ::
 
-    $ ./hwi.py -f e5dbc9cb getkeypool --addr-type wit 0 1000
+    $ ./hwi.py -f e5dbc9cb getkeypool 0 1000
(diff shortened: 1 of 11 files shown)


instagibbs:
Not sure I love the renaming. Taproot(next likely address type) isn't a hash either. What does miniscript et al suppose they will do for naming conventions?

achow101:
> Not sure I love the renaming. Taproot(next likely address type) isn't a hash either. What does miniscript et al suppose they will do for naming conventions?

Taproot would be WIT_V1

Sjors:
Concept ACK on changing the default to native segwit v0.
* `--addr-type  legacy` looks good too
* `--addr-type sh_wit_v0` is a bit atrocious. Since Taproot doesn't have wrapped segwit addresses, we might as well call it `sh_wit`
* I guess `wit_v0` makes sense in order to make room for `wit_v1`, though `wit` and `tap` could work too, since who cares that taproot is technically v1 segwit?


If we add support for miniscript or policy language, then the main purpose of the `--addr` argument is to disambiguate between v0, v1 witness pay to script hash, sh_wit_v0 and legacy. Both naming conventions work for that.

achow101:
>     * `--addr-type sh_wit_v0` is a bit atrocious. Since Taproot doesn't have wrapped segwit addresses, we might as well call it `sh_wit`

I had specified the v0 because I thought it might be possible that we will see p2sh nested again. But that's probably not going to happen so I guess it isn't needed. Renamed to `sh_wit`

>     * I guess `wit_v0` makes sense in order to make room for `wit_v1`, though `wit` and `tap` could work too, since who cares that taproot is technically v1 segwit?

I guess that's true. I suppose any future ones will also have their own special names so the actual version number doesn't really matter. Renamed to `wit`.

Sjors:
I guess I have to change this on the Bitcoin Core side...

achow101:
> I guess I have to change this on the Bitcoin Core side...

Oh... oops.

Sjors:
Core already relies on the master branch anyway for `--chain`

Sjors:
Actually Core doesn't use the `--addr_type` argument, because it uses descriptors.

I created https://github.com/bitcoin/bitcoin/pull/21417 to keep an eye on breaking changes (none so far).

PR #473, opened 2021-02-26
ledger: ensure only Bitcoin and Bitcoin Test apps are active

### LINKED ISSUE #158: Handle when user is not in the Bitcoin app on Ledger Nano S

achow101:
When the user is in an app on the Ledger that is not Bitcoin or Bitcoin Test, we should not be enumerating the device or allowing them to use HWI as unexpected and unsupported behavior may occur. To do this, this PR pulls in (part of) #362 in order to get the current app name. Then it checks that the app name is one of `Bitcoin`, `Bitcoin Test`, or `app` (for the app included in Speculos). For older devices which do not implement this instruction, the name is returned as `None`. Such devices will be enumerated and accessible as previously as we do not want to lock out users using old devices.

Closes #158

### DIFF
--- a/hwilib/devices/ledger.py
+++ b/hwilib/devices/ledger.py
@@ -21,6 +21,7 @@ from ..errors import (
     DeviceConnectionError,
     DeviceFailureError,
     UnavailableActionError,
+    UnknownDeviceError,
     common_err_msgs,
     handle_errors,
 )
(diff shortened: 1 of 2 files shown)


hwilib/devices/ledger.py
```diff
@@ -136,6 +137,10 @@ def __init__(self, path: str, password: str = "", expert: bool = False) -> None:
 
         self.app = btchip(self.dongle)
 
+        app_name = self.app.getAppName()
+        if app_name is not None and app_name != "Bitcoin" and app_name != "Bitcoin Test" and app_name != "app":
```
instagibbs, line 141: `if app_name not in ["Bitcoin", "Bitcoin Test", "app"]`

instagibbs:
Seems like something we can cover in tests.

```
"error": "Could not open client or get fingerprint information: get_pubkey_at_path canceled"
```
this is coming from 1.4.7 app. What version supports the new message?

hwilib/devices/ledger.py

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/473/commits/34003f831420065fa5ee20d4520a00871e329da5

PR #475, opened 2021-03-02
Add support for the Blockstream Jade hww

JamieDriver:


achow101:
Is there a simulator or emulator available so that we can run the automated tests?

JamieDriver:
> Is there a simulator or emulator available so that we can run the automated tests?

Not right now, although we have a WIP branch which compiles to run on the 'qemu' hw/chip emulator - which may be ok if you are ok to download/run a prebuilt binary ?

docs/devices/index.rst
```diff
@@ -99,3 +99,5 @@ Device APIs
    :members:
 .. automodule:: hwilib.devices.coldcard
    :members:
+.. automodule:: hwilib.devices.jade
+  :members:
```
achow101, line 103: I think the indentation might be wrong here as it is not showing up in the docs preview.

hwilib/devices/jade.py
```diff
@@ -0,0 +1,323 @@
...
+        super(JadeClient, self).__init__(path, password, expert)
+        self.jade = JadeAPI.create_serial(path)
+        self.jade.connect()
+
+        # Push some host entropy into jade
+        self.jade.add_entropy(os.urandom(32))
+
+        # Do the PIN thing if required
+        # NOTE: uses standard 'requests' networking to connect to blind pinserver
+        while not self.jade.auth_user(self._network()):
```
achow101, line 102: Is this interactive (i.e. user enters something on the host computer)? If so, is this required?

docs/devices/index.rst
JamieDriver: Apologies - I hope it's now repaired.

hwilib/devices/jade.py
JamieDriver: The user enters the pin on the hww.  The host then mediates between the hww and the remote blind pin server to unlock the keys on the hw.  
It is necessary in a 'real-world' situation to decrypt the users wallet/keys.

With a debug build of the fw [ie in a CI test] we can skip this step and poke a mnemonic or a seed directly into the hw.

JamieDriver: To clarify further, once the hww is unlocked (whether via pinserver or via 'poke in seed') 'auth_user' is a near no-op which returns immediately without user action / network activity.

So when I am manually testing, the first time I 'enumerate' I may have to unlock the Jade - but for the rest of my testing this call is a no-op (until/unless Jade is turned off).
Usually I manually inject a known seed into jade before I start testing, so a) I don't get interrupted with having to unlock it mid-test, and b) I get known/deterministic addresses/signatures etc for testing.

If/when we do get a sim working, that's what we'd aim to do.

docs/devices/index.rst
achow101: Hmm. It isn't showing up in the preview, but it does show up locally. So I guess it's fine.

hwilib/devices/jade.py
achow101: What is the "blind pin server"?

JamieDriver: Ok, well in a real-world situation the keys on a Jade hww cannot be decrypted without this interaction with a remote server.
( It can work offline once 'unlocked' - but the initial step to unlock the key material needs this handshake. )

instagibbs:
```
> What is "server enforced PIN protection"?

The mnemonic when generated is immediately encrypted using AES256 using a secret key that is computed through an encrypted and authenticated channel with a [remote] blind oracle server, the authentication involves ephemeral elliptic curve Diffie Hellman exchange with a known server key. The encrypted mnemonic is then stored on the encrypted flash of the Jade and protected by secure boot [v2].

The blind server is not aware of the actual PIN nor can brute force it. The blind server also is not associated with the users account (it is not associated with the per user derivation paths)

If the PIN is entered incorrectly 3 times the server and Jade both delete the secret requiring a from scratch mnemonic restore on Jade.

The companion app nor any ISP/WiFi AP in the middle can inspect the data passed between Jade and the oracle, including being able to tell if the PIN is correct or wrong (other than potential timing attacks). The oracle is available via Tor and open source and in the future users will be able to run their own and point their Jade to it if they prefer.
```

from https://www.reddit.com/r/Bitcoin/comments/kqgehd/were_the_blockstream_team_and_we_just_announced/gi3tvq0?utm_source=share&utm_medium=web2x&context=3

hwilib/devices/jadepy/jade.py
```diff
@@ -0,0 +1,819 @@
...
+#
+# Either:
+#  a) use via JadeInterface.create_ble() (see JadeInterface)
+# (recommended)
+# or:
+#  b) use JadeBleImpl() directly, and call connect() before
+#     using, and disconnect() when finished,
+# (caveat cranium)
+#
+class JadeBleImpl:
```
instagibbs, line 622: I'm guessing we wouldn't be using this?

JamieDriver: Right - at the top of the file I commented out the dependencies it needs to try to reduce transitive dependency creep (and didn't include them in setup/project files).  
We could remove all the code entirely - but it'll make refreshing the lib from jade's tree that much more involved.

prusnak: You could split the source code into different python modules - `jade_bluetooth.py`, `jade_serial.py`, then import these two into `jade.py`.

Then the matter of refreshing the lib would be reduced to not copying `jade_bluetooth.py` and removing the `from jade_bluetooth import JadeBleImpl` from `jade.py`.

JamieDriver: Updated Jade repo as suggested, and copied here.

wtogami:
Needs rebase

JamieDriver:
Maintainers - 
So I've got back to setting up the Jade emulator for the tests.  
Am I able to use 'docker' commands in the tests/setup ?

The Jade emulator runs a vaguely normal debug/headless build of the firmware in a version of the qemu emulator (https://www.qemu.org/).  ie. we don't build jade for x86, rather we build jade as normal (ie. for esp32) and then run it on x86 in an esp32 platform emulator.

* The esp-idf toolchain and dependencies are required to build the jade fw.  Rather than downloading and building these, we have the environment installed and available in a docker image.
* Rather than build the esp32 version of the qemu emulator, we have it built and installed in the docker image.

What I have trialled locally is:

`setup_environment.sh`: 
pulls our docker image as above
clones the jade source at a given release/tag
builds the jade fw inside the docker container and creates the qemu boot image
'commits' the container changes to a local image

`jade_test.py`:
Based on the other wallet's tests
Runs up the qemu (ie. jade) emulator in a new container (based on the image created in the setup step), exposing the jade tcp endpoint.
Runs the tests as normal
Stops the docker container

Does this sound like a workable/acceptable strategy ?

It's a bit rough-and-ready locally, but seems to be working.
If it sounds acceptable to you I'll polish it up and add it as a separate commit to this PR.
Thanks.

achow101:
@JamieDriver Sorry for taking so long to get back to this.

> Am I able to use 'docker' commands in the tests/setup ?

Unfortunately no. Cirrus CI uses docker containers to run everything under the hood, so it is not possible to have our test scripts pull and run additional docker containers.

JamieDriver:
No worries.  Thanks for the feedback - just not sure now how we can organise a CI.  I'll have a chat to the team.

pyproject.toml
```diff
@@ -25,6 +25,9 @@ typing-extensions = "^3.7"
 libusb1 = "^1.7"
 pyside2 = { version = "^5.14.0", optional = true }
 bitbox02 = ">=5.3.0,<6.0.0"
+cbor = "^1.0.0"
+requests = "^2.25.1"
```
achow101, line 29: `requests` is a disallowed dependency.

It is my understanding that this implementation will forego the PIN server and instead require the user to enter the mnemonic each time?

JamieDriver: Will make 'requests' an optional dependency in the api code and will remove from the setup/project files.

Georgantas:
Hi, I noticed some unexpected (to me) behavior using HWI with Jade. Not sure if it's intended:

I type this command using HWI: `./hwi.py --debug -t jade displayaddress --desc "wsh(sortedmulti(2,[xxxxxxxx/48'/0'/0'/2']xpubXXX/0/0,[zzzzzzzz/48'/0'/0'/2']xpubZZZ/0/0,[yyyyyyyy/48'/0'/0'/2']xpubYYY/0/0))" --addr-type wit` (notice xpubYYY < xpubZZZ lexicographically)

This triggers multisig registration on the jade. Using the `--debug` flag in the above command, I see this is sent to the device: `INFO:jade:Sending: {'method': 'register_multisig', 'id': '999999', 'params': {'network': 'mainnet', 'multisig_name': 'hwiXXXXXX', 'descriptor': {'variant': 'wsh(multi(k))', 'threshold': 2, 'signers': ...` Notice that `variant=wsh(multi(k))`, instead of `wsh(sortedmulti(k))` This is fine because I see under the `'signers'` sections, the xPubs are reordered lexicographically before being sent to the device for registration.

However, I then notice that the receive address displayed on the device is not the same as what I get in Specter or Sparrow. I retry the HWI command, using `0/1` this time for each of the signers, but still they're not matching up. I try again with `0/4`, and this time I see the address displayed on the device matches Sparrow/Specter.

I believe this mismatch is because Sparrow/Specter reorders the keys after applying `0/0`, while the Jade does not reorder them past the initial registration. The Jade is registered with `wsh(multi(k))`, so it wouldn't know that it needs to reorder them before generating the bech address. When I do `0/4`, I happen to land on a combination that is already ordered lexicographically.

JamieDriver:
> It seems sortedmulti is not being fully supported?

As it stands this is correct.  
If we register xpubX, xpubY, xpubZ and confirm on hw, then later we pass in paths [ [0,1], [0,2], [0,3] ] (say), we expect those to map to the xpubs in the order given, and the script/address generated will retain that order of the signers.

The immediate 'fix' may be to reject 'sortedmulti' descriptors...

JamieDriver:
> The immediate 'fix' may be to reject 'sortedmulti' descriptors...

sorted-multi should now be rejected at 'generate address'.  (Note: we can generate a sig for them, we just can't generate addresses (which also means we can't auto-validate change for them [which works by generating the address for those paths and checking it matches]).

JamieDriver:
> @JamieDriver Sorry for taking so long to get back to this.
> 
> > Am I able to use 'docker' commands in the tests/setup ?
> 
> Unfortunately no. Cirrus CI uses docker containers to run everything under the hood, so it is not possible to have our test scripts pull and run additional docker containers.

I will try to replicate what's in our Dockerfile into the 'setup_environment.sh' for Jade.
(If preferred I could move it into ci/cirrus.Dockerfile - which will obv save time per run, but will bloat the image.  
Unless you say otherwise I'll stick with 'setup_environment.sh'.)

JamieDriver:
> Is there a simulator or emulator available so that we can run the automated tests?

Simulator and tests for Jade added (finally!) - I believe consistent with the standard pattern.  
Moved the PR out of 'WIP'.
Cheers.

JamieDriver:
Branch rebased against upstream master, tests implemnted and passing against simulator as requested (the py3.8+ledger tests failing in this branch also appear to be failing in master).  
Is there anything else I need to do to progress this PR ?  
Cheers.

test/setup_environment.sh
```diff
@@ -258,6 +263,86 @@ if [[ -n ${build_ledger} ]]; then
     cd ..
 fi
 
+if [[ -n ${build_jade} ]]; then
+    # Testing with Jade tag 0.1.30
+    JADE_RELEASE_TAG="0.1.30"
```
achow101, line 268: In 67793345a98b95fdb045ee733aac3af0c34a3261 "Add tests for Jade hww using qemu emulator"

We generally prefer to test on the development branch at whatever the latest commit is. This allows us to be aware of any breaking changes before a release is made. It also means that this doesn't need to be updated every time a new firmware release is made.

hwilib/devices/jade.py
```diff
@@ -124,20 +128,25 @@ def __init__(self, path: str, password: str = '', expert: bool = False, timeout:
...
-                # User must use 'Emergency Restore' feature to enter mnemonic on Jade hw
-                raise DeviceNotReadyError("Use 'Emergency Restore' feature on Jade hw to enter wallet mnemonic")
-
-            # Authenticate the user - this may require a PIN and pinserver interaction
-            # (if we have required networking dependencies)
-            authenticated = self.jade.auth_user(self._network())
+        # Authenticate on the hw if required
+        if path == SIMULATOR_PATH:
+            # Push a test seed into the simulator
+            self.jade.set_seed(TEST_SEED)
```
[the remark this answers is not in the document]
JamieDriver: Fair - will address.

test/setup_environment.sh
JamieDriver: Bold!  Ok, I may have to make a change in the Jade repo so we can easily extract the relevant commits of the esp-idf toolkit and the qemu emulator that are relevant to that jade commit.  Leave it with me.

hwilib/devices/jade.py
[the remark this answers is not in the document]
JamieDriver: Done as suggested - tests start emulator and push in seed at that point.  Client does not set seed, but errors if wallet not initialised/seed not set (consistent with actual hw).

JamieDriver:
Hi,
'test' works for me, producing tpub's as expected:
```
(HWI) jamie@athens:~/blockstream/HWI$ hwi --chain test --fingerprint 1273da33 getdescriptors
WARNING:jade:No module named 'requests'
WARNING:jade:Default _http_requests() function will not be available
{"receive": ["pkh([1273da33/44h/1h/0h]tpubDCgUYU13ZZ2ES5mTZNfR93i2hGLKyCkGcbJtUJ2U1Lt9qQdNTSG7kQ4r6WK3mY7HVVKxyVpkbK6Hrdo1FLRHFRL4RP68eLmoySecZLLX5tW/0/*)#ey00p7p0", "sh(wpkh([1273da33/49h/1h/0h]tpubDCrYfPCx9xPgzLbhbqHYH5sac9XbJMT5CX63eFwKTVhvey15YXQTYwYmMoWoGq8EUBewX5Eo4xkpCWKL7Rq3XHfj8vgYoHuubRXDWqsbBJk/0/*))#jqhgsxj2", "wpkh([1273da33/84h/1h/0h]tpubDCtxUcxAGj2BY9SpLZcuaKPYykrX4Z85ZuYQU9jVsbKWeqrBqNx3qzpoyNcB9xCVenJzEQZXdzTLEszxnXQXKm2FgsgqcBHfsyeDoU5fH8m/0/*)#g2xqtj6u"], "internal": ["pkh([1273da33/44h/1h/0h]tpubDCgUYU13ZZ2ES5mTZNfR93i2hGLKyCkGcbJtUJ2U1Lt9qQdNTSG7kQ4r6WK3mY7HVVKxyVpkbK6Hrdo1FLRHFRL4RP68eLmoySecZLLX5tW/1/*)#gs2wut3h", "sh(wpkh([1273da33/49h/1h/0h]tpubDCrYfPCx9xPgzLbhbqHYH5sac9XbJMT5CX63eFwKTVhvey15YXQTYwYmMoWoGq8EUBewX5Eo4xkpCWKL7Rq3XHfj8vgYoHuubRXDWqsbBJk/1/*))#8pe7ge84", "wpkh([1273da33/84h/1h/0h]tpubDCtxUcxAGj2BY9SpLZcuaKPYykrX4Z85ZuYQU9jVsbKWeqrBqNx3qzpoyNcB9xCVenJzEQZXdzTLEszxnXQXKm2FgsgqcBHfsyeDoU5fH8m/1/*)#e7rpk82y"]}
```

Ah  - the issue may be that once your Jade hw has been setup for mainnet use, it won't allow that same wallet to be used on testnet/regtest etc - so is probably erroring there.

JamieDriver:
Good spot re: signet - Jade isn't aware of signet and is defaulting to mainnet:
Q:
a) does signet have the same version bytes/prefixes etc as testnet ?  if so I can set jade to produce 'testnet' type xpubs and addresses ... is that valid ?
b) rather than 'failing over' to mainnet, I should error if we see an unknown 'chain'.  Agree ?

Sjors:
> once your Jade hw has been setup for mainnet use, it won't allow that same wallet to be used on testnet/regtest etc - so is probably erroring there.

That might be the issue then. I went half-way through the GreenAddress setup flow (the pin selection entry went bezirk, perhaps because it was connected to laptop USB and iPhone via BlueTooth?) and aborted that. I then recovered the mnemonic and used the HWI flow. But with HWI I used signet first, not mainnet.


Signet is testnet for address prefixes, so you should be able to just do `test || signet` everywhere.

< b) rather than 'failing over' to mainnet, I should error if we see an unknown 'chain'. Agree ?

It's not an unknown chain, so better to give a specific warning (though I very much like to be able to try things on Signet on a device before mainnet, but that's a firmware discussion).

JamieDriver:
Many thanks - so if I tell jade that to all intents and purposes 'signet' xpubs and addresses etc are essentially 'testnet', then signet should work fine I think.  
Sorry, I guess I meant 'unhandled' or 'unexpected' chain.  ('unknown' as far as Jade is concerned - which is unaware of signet atm).
I'll make these changes now.

Btw you can press the button on Jade when the splash-screen is up, and it will offer the 'factory reset' option which should wipe whatever was partially setup before (if anything).

JamieDriver:
Mapping hwi Chain.SIGNET to Jade's 'testnet' yields:
```
(HWI) jamie@athens:~/blockstream/HWI$ hwi --chain signet --fingerprint 1273da33 getdescriptors
WARNING:jade:No module named 'requests'
WARNING:jade:Default _http_requests() function will not be available
{"receive": ["pkh([1273da33/44h/1h/0h]tpubDCgUYU13ZZ2ES5mTZNfR93i2hGLKyCkGcbJtUJ2U1Lt9qQdNTSG7kQ4r6WK3mY7HVVKxyVpkbK6Hrdo1FLRHFRL4RP68eLmoySecZLLX5tW/0/*)#ey00p7p0", "sh(wpkh([1273da33/49h/1h/0h]tpubDCrYfPCx9xPgzLbhbqHYH5sac9XbJMT5CX63eFwKTVhvey15YXQTYwYmMoWoGq8EUBewX5Eo4xkpCWKL7Rq3XHfj8vgYoHuubRXDWqsbBJk/0/*))#jqhgsxj2", "wpkh([1273da33/84h/1h/0h]tpubDCtxUcxAGj2BY9SpLZcuaKPYykrX4Z85ZuYQU9jVsbKWeqrBqNx3qzpoyNcB9xCVenJzEQZXdzTLEszxnXQXKm2FgsgqcBHfsyeDoU5fH8m/0/*)#g2xqtj6u"], "internal": ["pkh([1273da33/44h/1h/0h]tpubDCgUYU13ZZ2ES5mTZNfR93i2hGLKyCkGcbJtUJ2U1Lt9qQdNTSG7kQ4r6WK3mY7HVVKxyVpkbK6Hrdo1FLRHFRL4RP68eLmoySecZLLX5tW/1/*)#gs2wut3h", "sh(wpkh([1273da33/49h/1h/0h]tpubDCrYfPCx9xPgzLbhbqHYH5sac9XbJMT5CX63eFwKTVhvey15YXQTYwYmMoWoGq8EUBewX5Eo4xkpCWKL7Rq3XHfj8vgYoHuubRXDWqsbBJk/1/*))#8pe7ge84", "wpkh([1273da33/84h/1h/0h]tpubDCtxUcxAGj2BY9SpLZcuaKPYykrX4Z85ZuYQU9jVsbKWeqrBqNx3qzpoyNcB9xCVenJzEQZXdzTLEszxnXQXKm2FgsgqcBHfsyeDoU5fH8m/1/*)#e7rpk82y"]}
```

JamieDriver:
> Hi, I noticed some unexpected (to me) behavior using HWI with Jade. Not sure if it's intended:
<snipped>
> It seems `sortedmulti` is not being fully supported?

I will add support for `sortedmulti` in the next Jade fw release (which is due soon), and will then remove that restriction here.

achow101:
ACK db17fed8ff1109910ed65a36ddd20ba60bf90ee0

JamieDriver:
Many thanks Andrew.  

Once the next Jade fw is released I'll open a small PR to unblock 'sortedmulti' - and I should add a 'minimum supported fw version' check also.

PR #479, opened 2021-03-06
bitbox02: fix singlesig signing regression

benma:
`script_configs[0].script_config.multisig` is always True - a protobuf
pitfall. The way to check which variant it is is by using `WhichOneof`.


benma:
cc @ben-kaufman, who discovered it during testing of RC2.

instagibbs:
Will test.

instagibbs:
tACK 308763d

master signing fails with the vague "path" message, this PR branch signs correctly. I did not test multisig to make sure that wasn't broken.

PR #488, opened 2021-03-10
Properly handle and test sortedmulti descriptors

### LINKED ISSUE #487: Use of sortedmulti in displayaddress descriptor produces inconsistent results
According to [descriptors.md](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md), the use of `sortedmulti` should mean that "the order of public keys in the resulting witnessScripts is determined by the lexicographic order of the public keys at that index" (this for a `wsh` descriptor).

This suggests that using `sortedmulti` should mean calling `displayaddress` with the provided xpub/childpubkeypaths out of BIP67 ordering should result in the correct address being displayed according to BIP67. However, there are inconsistent results:
- Bitbox02 displays the correct address regardless of xpub ordering
- Trezor sometimes displays an incorrect address (possibly depending on whether the keys are already sorted?)
- Coldcard fails every time with `{"error": "Coldcard Error: BIP67 violation", "code": -7}` - even if the xpubs are in the correct order. Changing `sortedmulti` to `multi` fixes this.

achow101:
Implements better handling of `sortedmulti` descriptors. This is an API breaking change as `display_multisig_address` is changed to take the underlying `MultisigDescriptor` directly instead of the parameters stored within it.

For BitBox02 and Coldcard, since they only use BIP 67, they now require `sortedmulti` descriptors only.

For Trezors, `sortedmulti` is now being handled correctly. If a `sortedmulti` is provided, then the keys will be sorted appropriately prior to being sent to the Trezor for handling.

Lastly, tests are added to test both `multi` and `sortedmulti`, as well as both hex pubkeys and xpubs.

Fixes #487

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -457,7 +457,7 @@ def displayaddress(
                     addr_type = AddressType.SH_WIT
                 elif not is_sh and is_wsh:
                     addr_type = AddressType.WIT
-                return {"address": client.display_multisig_address(descriptor.thresh, descriptor.pubkeys, addr_type)}
+                return {"address": client.display_multisig_address(addr_type, descriptor)}
         is_wpkh = isinstance(descriptor, WPKHDescriptor)
         if isinstance(descriptor, PKHDescriptor) or is_wpkh:
             pubkey = descriptor.pubkeys[0]
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -306,7 +306,8 @@ class MultisigDescriptor(Descriptor):
         """
         super().__init__(pubkeys, None, "sortedmulti" if is_sorted else "multi")
         self.thresh = thresh
-        if is_sorted:
+        self.is_sorted = is_sorted
+        if self.is_sorted:
             self.pubkeys.sort()
 
     def to_string_no_checksum(self) -> str:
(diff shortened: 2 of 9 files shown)


instagibbs:
typecheck fail

thanks for tackling this

achow101:
Fixed the type check error.

instagibbs:
utACK https://github.com/bitcoin-core/HWI/pull/488/commits/3fad6631cc300c123015defa41618951afbcfb5f

moar tests :+1: 

@craigraw mind testing this to see if any more issues pop up?

craigraw:
Confirmed `sortedmulti` is working on Trezor, Coldcard and Bitbox02 using `hwi.py`. Using `multi` works on Trezor, and fails on Coldcard and Bitbox02 with an appropriate error message.

Oddly enough I can't seem to get a fully working binary from this branch - when communicating with the BitBox02 (for example with `enumerate`) the binary from `contrib/build_bin.sh` returns
```
  File "hwi.py", line 7, in <module>
  File "hwilib/_cli.py", line 305, in main
  File "hwilib/_cli.py", line 269, in process_commands
  File "hwilib/_cli.py", line 62, in enumerate_handler
  File "hwilib/commands.py", line 112, in enumerate
  File "importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "PyInstaller/loader/pyimod03_importers.py", line 627, in exec_module
  File "hwilib/devices/bitbox02.py", line 63, in <module>
.....
ValueError: source code string cannot contain null bytes
[17741] Failed to execute script hwi
```

PR #492, opened 2021-03-12
Use shutil.which to discover udevadm, groupadd and usermod binaries

### LINKED ISSUE #491: Command installudevrules fails when udevadm is not in /sbin
Ubuntu 20.04.1 places `udevadm` in `/usr/bin/` rather than `/sbin`, causing `installudevrules` to fail with
```
{"error": "installudevrules failed: [Errno 2] No such file or directory: '/sbin/udevadm'", "code": -13}
```

Ideally, hwi would detect the location with `which udevadm` before calling it (and `groupadd` and `usermod` as well).

prusnak:
Fixes #491

[`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) is available since python 3.3

### DIFF
--- a/hwilib/udevinstaller.py
+++ b/hwilib/udevinstaller.py
@@ -8,7 +8,7 @@ Classes and utilities for installing device udev rules.
 from .errors import NeedsRootError
 
 from subprocess import check_call, CalledProcessError, DEVNULL
-from shutil import copy
+from shutil import copy, which
 from os import path, listdir, getlogin, geteuid
 
 class UDevInstaller(object):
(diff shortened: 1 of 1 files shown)


prusnak:
> Type check error. This diff should fix it:

Amended in 09a747a61cac52b19280c7b7963b6407c130d7f7

prusnak:
> Type check error. This diff should fix it:

Amended in 09a747a61cac52b19280c7b7963b6407c130d7f7

prusnak:
> Type check error. This diff should fix it:

Amended in 09a747a61cac52b19280c7b7963b6407c130d7f7

prusnak:
Since the glitch did not recover I created a new PR here: https://github.com/bitcoin-core/HWI/pull/493

PR #494, opened 2021-03-16
Bump Pyenv's Python to 3.9

Sjors:
Since the release notes mention support for Python 3.9. Or do we also want to guarantee backwards compatibility?

Using 3.9.1 for PyEnv (latest available on macOS) and 3.9.2 for the Windows installer.


Sjors:
Seeing some timeouts.

achow101:
We want to guarantee backwards compatibility so everything should still use the oldest python version supported. The release note is just mentioning that people using Python 3.9 in their projects will be able to use HWI.

Sjors:
Ok, in that case I think it's sufficient to bump `.python-version` for PyEnv users.

Sjors:
Some confusing CI failings.

achow101:
Now that we have CI for all current python minor versions, we can bump pyenv and the build python versions to something more recent. Can you please rebase this and bump to more recent python?

Sjors:
Done, let's see how 3.9.7 fares (3.9.8 isn't available on Homebrew yet, so Cirrus CI probably also won't have it)

Sjors:
Done, let's see how 3.9.7 fares (3.9.8 isn't available on Homebrew yet, so Cirrus CI probably also won't have it)

Sjors:
The three test failures seem to be due to a timeout.

achow101:
ACK 3117eb4ffde8b5fad0f8da0ba4c153e492ce9612

.cirrus.yml
```diff
@@ -214,6 +214,7 @@ dist_builder_task:
   container:
     dockerfile: contrib/build.Dockerfile
   build_script:
+    - pyenv install 3.9.7
```
achow101, line 217: `contrib/build.Dockerfile` should be changed instead of cirrus.

achow101:
`contrib/build_wine.sh` should also be updated to use 3.9.7

.cirrus.yml
Sjors: Should this line be dropped entirely then?

achow101: Yes

Sjors: Done

achow101:
ACK d1055df7ba60ff6ba2a182980dcab1799c239c7a

Issue #283, opened 2019-12-04
How to prompt for passphrase on Trezor T


NicolasDorier:
Trezor T users seems to enter the pin before the device it can get enumerated.

However, I would like them to enter the passphrase after I can enumerate it. How do I prompt them to enter the password on the device?

achow101:
Doing a `getxpub` (or `getmasterxpub`) will prompt the Trezor T to ask for the passphrase.

NicolasDorier:
thanks a lot!

stevenroose:
This seems to be broken in 2.0.0. I think it might be a firmware issue as well.

prusnak:
@stevenroose Are you on the latest firmware? If not, update and try again. If yes, please report the version number.

NicolasDorier:
There was something probably broken in trezor, but I don't think that's related to hwi 2.0

https://github.com/btcpayserver/BTCPayServer.Vault/issues/29

prusnak:
> This seems to be broken in 2.0.0. I think it might be a firmware issue as well.

Fix probably in https://github.com/bitcoin-core/HWI/pull/495

stevenroose:
Yeah I can confirm that #495 fixes this.

PR #495, opened 2021-03-18
trezor: Return PASSPHRASE_ON_DEVICE if passphrase entry on device is available

achow101:
Somewhere between firmware versions 2.3.0 and 2.3.6, the Trezor T started using a slightly different message which results in our passphrase handling incorrectly giving the passphrase from the host instead of forcing passphrase entry on device. We fix this issue by returning the correct object to indicate that passphrase entry should occur on device.

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -29,7 +29,7 @@ from ..errors import (
     common_err_msgs,
     handle_errors,
 )
-from .trezorlib.client import TrezorClient as Trezor
+from .trezorlib.client import TrezorClient as Trezor, PASSPHRASE_ON_DEVICE
 from .trezorlib.debuglink import TrezorClientDebugLink
 from .trezorlib.exceptions import Cancelled, TrezorFailure
 from .trezorlib.transport import (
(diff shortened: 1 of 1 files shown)


stevenroose:
I can confirm that this re-fixes https://github.com/bitcoin-core/HWI/issues/283 :)

adamPetho:
Does it work with the older of firmware? Is this backward compatible? Is it possible to release a hotfix so we can test and update it?

nopara73:
@achow101, @prusnak we're getting more and more support request regarding this. Could one of you release a compatibility fix?

- For HWI it'd need to be compatible to both old Trezor firmware and new one.
- For Trezor, it'd need to be compatible with current HWI release.

achow101:
@nopara73 I've released 2.0.1

prusnak:
@achow101 Thanks!

@nopara73 I am just a contributor to HWI, I cannot cut a new release. However, please let us know whether the 2.0.1 release fixed the issues you have. Thanks!

nopara73:
Thank you very much guys! The testing discussion will go under https://github.com/zkSNACKs/WalletWasabi/pull/5465

PR #496, opened 2021-03-19
Update Coldcard patch for latest firmware

achow101:
Coldcard recently changed the layout of their firmware repo which broke the patch we were applying to it. This updates that patch.


achow101:
Closing in favor of #523 for general CI fixes.

PR #502, opened 2021-04-02
support other libusb paths for Linux

secinthenet:
Some linux distros such as Arch/Manjaro use /usr/lib


hwi.spec
```diff
@@ -1,21 +1,29 @@
...
+        libusb_path = '/lib/x86_64-linux-gnu/libusb-1.0.so.0'
+    elif os.path.exists('/usr/lib/libusb-1.0.so.0'):
+        libusb_path = '/usr/lib/libusb-1.0.so.0'
+    else:
+        libusb_path = '/lib/libusb-1.0.so.0'
 elif platform.system() == 'Darwin':
     find_brew_libusb_proc = subprocess.Popen(['brew', '--prefix', 'libusb'], stdout=subprocess.PIPE)
-    libusb_path = find_brew_libusb_proc.communicate()[0]
-    binaries = [(libusb_path.rstrip().decode() + "/lib/libusb-1.0.dylib", ".")]
+    libusb_path = find_brew_libusb_proc.communicate()[0].rstrip().decode + "/lib/libusb-1.0.dylib"
```
prusnak, line 21:
```suggestion
    libusb_path = find_brew_libusb_proc.communicate()[0].rstrip().decode() + "/lib/libusb-1.0.dylib"
```

hwi.spec
```diff
@@ -1,21 +1,29 @@
...
 import subprocess
+import os
 
 block_cipher = None
 
 binaries = []
+
 if platform.system() == 'Windows':
-    binaries = [("c:/python3/libusb-1.0.dll", ".")]
+    libusb_path = 'c:/python3/libusb-1.0.dll'
```
achow101, line 11: nit: Use double quotes rather than single quotes

hwi.spec
```diff
@@ -1,21 +1,29 @@
...
 elif platform.system() == 'Linux':
-    binaries = [("/lib/x86_64-linux-gnu/libusb-1.0.so.0", ".")]
+    if os.path.exists('/lib/x86_64-linux-gnu/libusb-1.0.so.0'):
+        libusb_path = '/lib/x86_64-linux-gnu/libusb-1.0.so.0'
+    elif os.path.exists('/usr/lib/libusb-1.0.so.0'):
+        libusb_path = '/usr/lib/libusb-1.0.so.0'
+    else:
+        libusb_path = '/lib/libusb-1.0.so.0'
```
achow101, line 18: I wonder if there is some way to automatically discover the path to libusb similar to what we do for Mac? Maybe something that uses ld?

achow101 (commented):
Please squash

hwi.spec
prusnak: This seems relevant: https://github.com/albertz/system-tools/blob/master/bin/find-lib-in-path.py

hwi.spec
secinthenet: I used single quotes because most of the file uses it. Should I change the other parts as well?

hwi.spec
secinthenet: I can parse `/etc/ld.so.conf` like the code @prusnak referenced, but on my system it doesn't help and it adds some complexity. I also can't copy paste the it right now since it doesn't have a license.
Also note that it doesn't address the priority between these paths, and I'm not aware of a way to do it.

What do you think is best here?
Should I parse `/etc/ld.so.conf`, or keep it as is?

prusnak: You could at least try to refactor the code to something like this so it is more extensible in the future:

``` python
libusb_path = None
paths = ["/lib/x86_64-linux-gnu", "/usr/lib/x86_64-linux-gnu", "/lib64", "/usr/lib64", "/lib", "/usr/lib"]
for path in paths:
    lib = path + "/libusb-1.0.so.0" 
    if os.path.exists(lib):
        libusb_path = lib
        break
```

infokiller: Done

hwi.spec
achow101: Just the changes that you have made.

hwi.spec
achow101: I was thinking something more like run `ldconfig -p | grep "libusb-1.0.so.0 (libc6,x86-64)"` and take the path from that.

secinthenet: That relies on glibc and will break on musl (so alpine and others won't work)

achow101:
ACK 8a37eb956184302a168099d454440ea8105d7e3f

PR #503, opened 2021-04-03
set correct permissions for deployed udev files

secinthenet:
It was previously set to 700 on my machine.


achow101:
ACK 5452eacbe8e8467bfdb40304051810c961ba84c7

PR #504, opened 2021-04-04
Always use h instead of ' for hardened derivation path

achow101:
Changes KeyOriginInfo to use h instead of ' for hardened derivation paths. Also checks this in the tests.


wtogami:
tested ACK, fixed my problem using `getkeypool` output verbatim into `importdescriptors`.

PR #505, opened 2021-04-06
bitbox02: handle multiple connected bitbox02s correctly

benma:
If there were more than one BitBox02 connected, enumerate would
fail. This commit fixes a small mistake in the initialization,
allowing HWI to handle multiple BitBox02s at once.


achow101:
ACK 15d8eca6daad6f425a03cf643bd14f610d086341

Issue #506, opened 2021-04-22
Is Ubuntu 16.04 supported by HWI 2.0+?


kiminuo:
Hi,

when `hwi` (2.0.1) is run on Ubuntu 16.04 it reports:

```bash
$ ./hwi
[11765] Error loading Python lib '/tmp/_MEImoo770/libpython3.6m.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by /tmp/_MEImoo770/libpython3.6m.so.1.0)
```

I'm not sure whether HWI 2.0+ is supposed to be run on newer Ubuntu versions or whether this is a bug.

Thanks!

PS: Ubuntu 18.04 reports:

```
[12221] Error loading Python lib '/tmp/_MEIJQ9F0E/libpython3.6m.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/_MEIJQ9F0E/libpython3.6m.so.1.0)
```

achow101:
Do you see this issue with 2.0.0 or earlier?

If not, I think the cause was #500. Python built in a newer version probably requires a newer glibc that is not present on older versions of Ubuntu.

kiminuo:
> Do you see this issue with 2.0.0 or earlier?

I tested it right now and I can see the error on 2.0.1 but NOT on 2.0.0 on my Ubuntu 18.04.

molnard:
@achow101 can this be fixed or are you planning to? Otherwise, older Linux support has to be dropped. Similarly, you did with macOS, like building the binaries with older dependencies.

achow101:
@molnard I'll look into building with an older Linux for the next release. However HWI is currently low priority for me so I have not had the time to investigate.

kiminuo:
> If not, I think the cause was #500. Python built in a newer version probably requires a newer glibc that is not present on older versions of Ubuntu.

I think so too.

* https://packages.debian.org/source/stretch/glibc mentions `2.24`
* https://packages.debian.org/source/buster/glibc mentions `2.28`

My Ubuntu 18.04 reports `glibc` version `Version: 2.27-3ubuntu1.4` (using `sudo apt-cache show libc6`).

Szpoti:
Under Ubuntu 16.04.7 I get
[...]

PR #510, opened 2021-05-26
Fix installing Wine Development and staging builds on Debian Buster

gruve-p:
`libfaudio0` dependency is the missing for Wine staging and Wine development. 

This dependency is not directly available in the official WineHQ Ubuntu and Debian 10 repository because it's not part of the Wine project. It is mentioned (along with a link to a forum post) at the top of the WineHQ Ubuntu and Debian installation pages that users need to download and install this package separately, but many users ignore it / don't read that part. On the other hand, those upgrading from older Wine builds / those that already have the WineHQ repository added never get to the official WineHQ installation page, so they won't notice this very important detail.

It's worth noting that the missing `libfaudio0` dependency has been added to the Debian bullseye / sid and the upcoming Ubuntu 19.10 release, so this is only an issue for older releases.


achow101:
We're moving back to debian stretch because of #506.

Superseded by #512

PR #511, opened 2021-05-26
Fix libusb path for Darwin

gruve-p:
Currently it can't find libusb path on Darwin, this PR fixes it.

### DIFF
--- a/hwi-qt.spec
+++ b/hwi-qt.spec
@@ -11,7 +11,7 @@ def get_libusb_path():
     if platform.system() == 'Darwin':
         proc = subprocess.Popen(['brew', '--prefix', 'libusb'], stdout=subprocess.PIPE)
         prefix = proc.communicate()[0].rstrip().decode()
-        return os.path.join(prefix, "/lib/libusb-1.0.dylib")
+        return os.path.join(prefix + "/lib/libusb-1.0.dylib")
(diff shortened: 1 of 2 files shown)


hwi-qt.spec
```diff
@@ -11,7 +11,7 @@ def get_libusb_path():
     if platform.system() == 'Darwin':
         proc = subprocess.Popen(['brew', '--prefix', 'libusb'], stdout=subprocess.PIPE)
         prefix = proc.communicate()[0].rstrip().decode()
-        return os.path.join(prefix, "/lib/libusb-1.0.dylib")
+        return os.path.join(prefix + "/lib/libusb-1.0.dylib")
```
prusnak, line 14: The correct way of fixing this is:

```suggestion
        return os.path.join(prefix, "lib", "libusb-1.0.dylib")
```

gruve-p: I don't mind how it is fixed as long it is fixed because currently it is broken and can be replicated with a macbook installing libusb through brew.

achow101: I prefer this to be implemented using the correct join syntax as suggested by @prusnak.

achow101:
Superseded by #514

ItsYoutacRRS (commented):
###

PR #516, opened 2021-06-04
Return status indicating if we signed the PSBT

barrystyle:
Report back whether we were able to sign PSBT, rather than having to compare two long base64 strings by eye; which can be very frustrating during multiple debug sessions.


achow101:
Can you add some tests for this?

barrystyle:
> Can you add some tests for this?

no problem..

PR #520, opened 2021-08-06
Add .whitesource configuration file

Jsn2win:


achow101:
NACK

IDE config files should be kept locally to avoid polluting contributors source trees with unnecessary files.

Issue #517, opened 2021-06-09
Trezor emulator is seen as both trezor and keepkey


Sjors:
Using the 2.0.2 release with the latest Trezor emulator build:
```
hwi  --device-type trezor enumerate
```

```
[{"type": "trezor", "path": "udp:127.0.0.1:21324", "model": "trezor_t_simulator", "needs_pin_sent": false, "needs_passphrase_sent": false, "fingerprint": "a904f46a"}, {"type": "keepkey", "model": "keepkey", "path": "udp:127.0.0.1:21324", "error": "Could not open client or get fingerprint information: Unsupported device", "code": -13}]
```

prusnak:
I asked Keepkey to change their emulator port: https://github.com/keepkey/keepkey-firmware/issues/276

prusnak:
https://github.com/keepkey/keepkey-firmware/issues/276 has been closed - we can change the Keepkey's port from 21324 to 11044

prusnak:
Fix in https://github.com/bitcoin-core/HWI/pull/519

Talkless:
I see fix is merged, so I've checkouted master, but still got this issue:

```
$ git status | fgrep branch
On branch master
$ ./hwi.py enumerate
[{"type": "trezor", "path": "udp:127.0.0.1:21324", "model": "trezor_t_simulator", "needs_pin_sent": false, "needs_passphrase_sent": false, "fingerprint": "c664c2b7"}, {"type": "keepkey", "model": "keepkey", "path": "udp:127.0.0.1:21324", "error": "Could not open client or get fingerprint information: Unsupported device", "code": -13}]
```
Still "keepkey" ?

prusnak:
@Talkless Have you updated (and rebuilt) the keepkey emulator? Because it seems you are still running the old keepkey code.

josibake:
> Have you updated (and rebuilt) the keepkey emulator?

the original issue was about using the trezor emulator, and it seems as tho the issue still exists. i pulled the latest branch of HWI, updated and rebuilt the trezor emulator and i am still getting the same error as @Sjors 

looks like you might have fixed an issue relating to using the keepkey emulator, @prusnak , but i would suggest this issue be re-opened as it relates to the trezor emulator.

prusnak:
Ah, correct. Keepkey enumeration did not work as expected - fix in https://github.com/bitcoin-core/HWI/pull/521

PR #521, opened 2021-08-19
Fix keepkey enumeration so it does not appear as Trezor

prusnak:
Fixes https://github.com/bitcoin-core/HWI/issues/517


josibake:
ACK https://github.com/bitcoin-core/HWI/pull/521/commits/6a481134fb3954470f1e0ca713f52063f251894c

thanks picking this up so quickly, @prusnak ! i ran the tests locally, and also ran with trezor emulator. confirmed that now trezor is the only device seen:

```bash
./hwi.py enumerate
[{"type": "trezor", "path": "udp:127.0.0.1:21324", "model": "trezor_t_simulator", "needs_pin_sent": false, "needs_passphrase_sent": false, "fingerprint": "a83bd035"}]
```

note: i have not tested with the keepkey emulator. also not sure if the build failures are related; seems like they've been failing previously as well

josibake (approved):

PR #523, opened 2021-08-27
Repair CI and update tests

achow101:
* Cleanup of `.cirrus.yml`
* Change to use artifacts rather than caching to get built things to test tasks. This should make it less likely to fail due to race conditions around the caching.
* Fix wine_builder to download the actual key. Python.org removed the keys file we used previously.
* Install rustup and clang in `cirrus.Dockerfile`. Trezor needs this now.
* Fix Coldcard simulator patches and build
* Fix Speculos simulator build
* Patch and fix Keepkey emulator builds
* Fix keepkey transport finding.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -121,7 +121,7 @@ lint_task:
   test_script:
     - flake8
 
-basic_test_task:
+task:
   install_script:
     - poetry install
   matrix:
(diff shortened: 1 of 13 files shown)


achow101:
Merging this as CI is finally passing.

PR #536, opened 2021-10-24
test: Fix label tests

achow101:
These tests weren't working. Also fixed another keepkey test bug.


achow101:
Merging as CI passed

Issue #539, opened 2021-10-31
Trezor One: Trezor is not responsive at "access hidden wallet"


cjackie:
I have a Trezor One with pin and passphrase enabled. I used hwi 2.0.2. Here is how to reproduce the issue:

```python
from hwilib.devices.trezor import TrezorClient
c = TrezorClient("<path>", "<passphrase>")
c.prompt_pin()
c.send_pin("00000000")
```

Pin was verified successfully according to Trezor One, then it switched to "Access hidden wallet?" with two options "Cancel" and "Confirm". When I press either button, the Trezor One did not respond.

Here is some logs while running the code above.

```
client.py:_raw_write:DEBUG:sending message: GetFeatures
client.py:_raw_read:DEBUG:received message: Features
client.py:_raw_write:DEBUG:sending message: Initialize
client.py:_raw_read:DEBUG:received message: Features
client.py:_raw_write:DEBUG:sending message: GetPublicKey
client.py:_raw_read:DEBUG:received message: PinMatrixRequest
client.py:_raw_write:DEBUG:sending message: PinMatrixAck
client.py:_raw_read:DEBUG:received message: PassphraseRequest
client.py:_raw_write:DEBUG:sending message: PassphraseAck
client.py:_raw_read:DEBUG:received message: ButtonRequest
```

achow101:
Fixed by #542

PR #540, opened 2021-10-31
Fix Trezor One freezing at access hidden wallet

cjackie:
See https://github.com/bitcoin-core/HWI/issues/539 for details on the issue. I tested on my Trezor One.

PR #542, opened 2021-11-09
keepkey, trezor: Handle if ButtonRequest is sent during sendpin

### LINKED ISSUE #526: Keep Key HWI issue?
Seeing issues with the Keep Key in multiple software at the point of sending the passphrase.  This happens in Specter as well as Sparrow wallets almost identically.  The Sparrow developer pointed me here as he suspects the issue may be in HWI which both wallets use.  Electrum does not have this issue.

In Specter it the error looks like this:
image

And in Sparrow we see something similar:
image

This is with the latest KeepKey firmware (7.2.1).  And I have reproduced the issue on both Windows 10 and Arch Linux. Electrum does not have this issue with the KeepKey and works properly.  I do not know if it used HWI.  

Here are the respective posts for both of these wallets:
https://github.com/cryptoadvance/specter-desktop/issues/1219
https://github.com/sparrowwallet/sparrow/issues/219

(strange they both end in 219? ;) )

Anyway.  Is this the right place to report the issue?

achow101:
Keepkey and Trezor may ask the user for confirmation of their passphrase after the PIN is sent. This could cause some issues as `ButtonRequests` were not being handled by `sendpin`. Additionally, `TrezorClient.call` would not work because it needs access to `TrezorClient.features`, which is not set for the `sendpin` command. In order to resolve this, `TrezorClient.call` is modified to optionally skip the firmware check that requires `TrezorClient.features`, and `sendpin` uses `TrezorClient.call` instead of `TrezorClient.call_raw`.

Fixes #526 
Fixes $539

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -737,7 +737,7 @@ class TrezorClient(HardwareWalletClient):
                     raise DeviceAlreadyUnlockedError('The PIN has already been sent to this device')
             return False
         elif isinstance(resp, messages.PassphraseRequest):
-            pass_resp = self.client.call_raw(messages.PassphraseAck(passphrase=self.client.ui.get_passphrase(available_on_device=False), on_device=False))
+            pass_resp = self.client.call(messages.PassphraseAck(passphrase=self.client.ui.get_passphrase(available_on_device=False), on_device=False), check_fw=False)
             if isinstance(pass_resp, messages.Deprecated_PassphraseStateRequest):
                 self.client.call_raw(messages.Deprecated_PassphraseStateAck())
         return True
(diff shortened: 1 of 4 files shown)


prusnak:
Seems like this is the same issue that PR https://github.com/bitcoin-core/HWI/issues/539 is trying to fix

prusnak:
Seems like this is the same issue that PR https://github.com/bitcoin-core/HWI/pull/540 is trying to fix

In another word, this also fixes https://github.com/bitcoin-core/HWI/issues/539

@cjackie can you please confirm this works for you?

achow101:
Oh, interesting, I didn't realize that #539 was basically the same problem. I've updated this to actually handle all the callbacks correctly.

cjackie:
Hi @prusnak, I can verify that this patch works.

PR #544, opened 2021-11-17
trezor: Implement Taproot support

achow101:
Implements Taproot support for Trezor devices.

Depends on #545, #546, #547, and #548

No tests yet as they would need https://github.com/bitcoin/bitcoin/pull/22558 merged into Bitcoin Core.

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -761,13 +761,15 @@ class TrezorClient(HardwareWalletClient):
     @trezor_exception
     def can_sign_taproot(self) -> bool:
         """
-        Trezor T supports Taproot in firmware versions greater than (not including) 2.4.2.
-        Trezor One supports Taproot in firmware versions greater than (not including) 1.10.3.
-        However HWI does not implement Taproot support for any Trezor devices yet.
+        Trezor T supports Taproot since firmware version 2.4.3.
+        Trezor One supports Taproot since firmware version 1.10.4.
 
         :returns: False, always.
         """
-        return False
+        self._prepare_device()
+        if self.client.features.model == "T":
+            return bool(self.client.version >= (2, 4, 3))
+        return bool(self.client.version >= (1, 10, 4))
 
 
 def enumerate(password: str = "") -> List[Dict[str, Any]]:
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -361,6 +361,7 @@ class TrezorClient(HardwareWalletClient):
             # Prepare inputs
             inputs = []
             to_ignore = [] # Note down which inputs whose signatures we're going to ignore
+            has_tr = False
             for input_num, (psbt_in, txin) in py_enumerate(list(zip(tx.inputs, tx.tx.vin))):
                 txinputtype = messages.TxInputType(
                     prev_hash=ser_uint256(txin.prevout.hash)[::-1],
@@ -464,6 +465,7 @@ class TrezorClient(HardwareWalletClient):
                         for key, (leaf_hashes, origin) in psbt_in.tap_bip32_paths.items():
                             # TODO: Support script path signing
                             if key == psbt_in.tap_internal_key and origin.fingerprint == master_fp:
+                                has_tr = True
                                 txinputtype.address_n = origin.path
                                 found = True
                                 our_keys += 1
@@ -483,6 +485,10 @@ class TrezorClient(HardwareWalletClient):
 
                 # append to inputs
                 inputs.append(txinputtype)
+
+            # Cannot sign transactions that have external inputs (to_ignore is not empty) and would sign taproot inputs
+            if has_tr and len(to_ignore) > 0:
+                raise BadArgumentError("Trezor cannot sign taproot inputs when the transaction also has external inputs")
 
             # address version byte
             if self.chain != Chain.MAIN:
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -596,7 +596,9 @@ class TrezorClient(HardwareWalletClient):
         elif addr_type == AddressType.LEGACY:
             script_type = messages.InputScriptType.SPENDADDRESS
         elif addr_type == AddressType.TAP:
-            raise UnavailableActionError("Trezor does not support displaying Taproot addresses yet")
+            if not self.can_sign_taproot():
+                raise UnavailableActionError("This device does not support displaying Taproot addresses")
+            script_type = messages.InputScriptType.SPENDTAPROOT
         else:
             raise BadArgumentError("Unknown address type")
 
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -100,6 +100,16 @@ Use the numeric keypad to describe number positions. The layout is:
 
 Device = Union[hid.HidTransport, webusb.WebUsbTransport, udp.UdpTransport]
 
+ECDSA_SCRIPT_TYPES = [
+    messages.InputScriptType.SPENDADDRESS,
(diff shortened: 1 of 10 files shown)


Sjors:
Nice! It might be worth splitting the descriptor, bech32(m) and psbt handling code into their own packages. That takes up a big chunk of the PR and other projects might benefit from them.

The "Trezor: Update trezorlib" and KeepKey commit(s) could be their own PR, and could use instructions on how to verify.

achow101:
I've broken up this PR into a number of smaller PRs.

Sjors:
Did some light testing with a physical Model T (firmware 2.4.2) . I'm able to create a fresh wallet and import the descriptors, including `tr()`.

Calling `walletdisplayaddress` on a taproot address results in nothing on the screen (except briefly a spinner).

Signing a transaction fails with "Failed to sign: Signer fingerprint ...... does not match any of the inputs:" (even though it does). Maybe I screwed up in rebasing https://github.com/bitcoin/bitcoin/pull/22558 so I'll try that again later.

Should I wait for a new firmware release?

When fetching the descriptors from pre-taproot firmware, we should omit the `tr()` descriptor and probable emit a warning to that effect (that the user should upgrade their firmware and create a new wallet).

achow101:
> Should I wait for a new firmware release?

That or you need to compile it yourself.

> When fetching the descriptors from pre-taproot firmware, we should omit the `tr()` descriptor and probable emit a warning to that effect (that the user should upgrade their firmware and create a new wallet).

Seems like a good idea.

Sjors:
> That or you need to compile it yourself.

I'll try that. You may want to point that out in the PR description :-)

prusnak:
You can download the firmware from our CI here: 

https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/pipelines?page=1&scope=all&ref=master&status=success

(click on vertical three dots on the right side and select `core fw regular build:archive`)

You should get a file like `trezor-fw-regular-2.4.3-0fc3b32b.zip` containing `trezor-fw-regular-2.4.3-0fc3b32b.bin` which you can flash via `trezorctl firmware-update -f *.bin`.

**Be careful, this is a debug firmware, so the update process will erase the device storage if you are upgrading from production firmware!**

Sjors:
@prusnak thanks! I was just about to ask how to build the firmware _without Docker_, but for this testing the download is fine too.

prusnak:
> how to build the firmware _without Docker_

https://github.com/trezor/trezor-firmware/blob/master/docs/core/build/embedded.md

Sjors:
With the pre-release firmware displaying an address works like a charm. But even with the latest bitcoin/bitcoin#22558 I still get the "does not match any of the inputs" error when trying to spend. `analyzepsbt` says it needs an "updater".

achow101:
> With the pre-release firmware displaying an address works like a charm. But even with the latest [bitcoin/bitcoin#22558](https://github.com/bitcoin/bitcoin/pull/22558) I still get the "does not match any of the inputs" error when trying to spend. `analyzepsbt` says it needs an "updater".

This appears to be an issue in Core as bitcoin/bitcoin#22558 does not update the external signer stuff. As the BIP 32 derivation paths are in a new field for Taproot, it looks like the `ExternalSigner::SignTransaction` does not search the new derivation paths field and so it is unaware of the keys. The failure occurs prior to HWI being called.

If you use HWI directly, this should work.

Sjors:
I see. I didn't expect there to be anything about the external signer scriptPubKeyMan an update, but we'll figure that out later.

achow101:
I've added firmware version checks so that Taproot things won't be given out if the firmware does not support Taproot.

hwilib/devices/trezor.py
```diff
@@ -754,6 +787,20 @@ def toggle_passphrase(self) -> bool:
...
+        """
+        Trezor T supports Taproot in firmware versions greater than (not including) 2.4.2.
+        Trezor One supports Taproot in firmware versions greater than (not including) 1.10.3.
+
+        :returns: False, always.
+        """
+        self._prepare_device()
+        if self.client.features.model == "T":
+            return bool(self.client.version > (2, 4, 2))
+        return bool(self.client.version > (1, 10, 3))
```
prusnak, line 801: This is just semantics but this reads and feels much better:

```suggestion
            return bool(self.client.version >= (2, 4, 3))
        return bool(self.client.version >= (1, 10, 4))
```

hwilib/devices/trezor.py
```diff
@@ -754,6 +787,20 @@ def toggle_passphrase(self) -> bool:
                 print(PIN_MATRIX_DESCRIPTION, file=sys.stderr)
         return True
 
+    @trezor_exception
+    def can_sign_taproot(self) -> bool:
+        """
+        Trezor T supports Taproot in firmware versions greater than (not including) 2.4.2.
+        Trezor One supports Taproot in firmware versions greater than (not including) 1.10.3.
```
prusnak, line 794: This is just semantics but this reads and feels much better:

```suggestion
        Trezor T supports Taproot since firmware version 2.4.3.
        Trezor One supports Taproot since firmware version 1.10.4.
```

hwilib/devices/trezor.py
achow101: Done.

I wasn't sure what number the next versions are going to be, but if they are definitely 2.4.3 and 1.10.4, then this is fine.

hwilib/devices/trezor.py

hwilib/devices/keepkey.py
```diff
@@ -169,6 +169,14 @@ def can_sign_taproot(self) -> bool:
...
         return False
 
+    def can_sign_taproot(self) -> bool:
+        """
+        The KeepKey does not support Taproot yet.
+
+        :returns: False, always
+        """
+        return False
+
```
gruve-p, line 179: `can_sign_taproot` is already defined above and is a duplicate

achow101: Oops, fixed.

test/test_device.py
```diff
@@ -24,7 +24,7 @@
 SUPPORTS_MULTISIG = {'ledger', 'trezor_1', 'digitalbitbox', 'keepkey', 'coldcard', 'trezor_t', 'jade'}
 SUPPORTS_EXTERNAL = {'ledger', 'trezor_1', 'digitalbitbox', 'keepkey', 'coldcard', 'trezor_t', 'jade'}
 SUPPORTS_OP_RETURN = {'ledger', 'digitalbitbox', 'trezor_1', 'trezor_t', 'keepkey', 'jade'}
-SUPPORTS_TAPROOT = {}
+SUPPORTS_TAPROOT = {"trezor_1", "trezor_t"}
```
gruve-p, line 27:
```suggestion
SUPPORTS_TAPROOT = {'trezor_1', 'trezor_t'}
```

test/test_device.py
```diff
@@ -373,6 +377,9 @@ def _test_signtx(self, input_type, multisig, external, op_return: bool):
         sh_wpkh_addr = self.wrpc.getnewaddress('', 'p2sh-segwit')
         wpkh_addr = self.wrpc.getnewaddress('', 'bech32')
         pkh_addr = self.wrpc.getnewaddress('', 'legacy')
+        tr_addr = None
+        if self.full_type in SUPPORTS_TAPROOT:
+            tr_addr = self.wrpc.getnewaddress("", "bech32m")
```
gruve-p, line 382:
```suggestion
            tr_addr = self.wrpc.getnewaddress('', "bech32m")
```

test/test_device.py
achow101: I'm trying to move the style to using double quotes rather than single.

PR #545, opened 2021-11-19
Add tr() descriptors for Taproot output scripts

achow101:
Implements `tr()` descriptors

PR #547, opened 2021-11-19
Add AddressType.TAP and be able to fetch Taproot descriptors from getdescriptors and getkeypool

achow101:
Adds a `AddressType.TAP` which can be used as `tap` in `displayaddress` and anywhere else `--addr-type` is used. Implements the bech32m encoding scheme. Also adds `tr()` descriptors to `getdescriptors` and `getkeypool` output.

Depends on #545


Sjors:
utACK 062dc59 (for the new commits in this PR)

Sjors:
utACK 3f74cc75ecfc24eb773efd4ec3728a6ad5cfc8e9

PR #548, opened 2021-11-19
trezor: Update trezorlib Nov 2021

achow101:
Update Trezorlib. This update in particular pulls in necessary components for Taproot support.


Sjors:
Is there a handy incantation to compare the first commit to upstream? As for the second commit, maybe we can maintain a patch file that is then applied in a (scripted) diff?

achow101:
> Is there a handy incantation to compare the first commit to upstream?

No. I just `diff` each file.

> As for the second commit, maybe we can maintain a patch file that is then applied in a (scripted) diff?

I tried that once, but we do this update relatively infrequently and so by the time we are ready to update, trezorlib has changed enough that such a patch no longer applies.

Sjors:
> trezorlib has changed enough that such a patch no longer applies

Does it change so much that even modifying the patch is pointless?

Sjors:
utACK c740c3d3ffb33c2f3f00101d212bc8153b799e2e

Recipe used for ~verifying:
* checkout the trezor firmware commit mentioned in the first commit here
* from hwi repo:
  * `rm -rf hwilib/devices/trezorlib`
  * `cp -r ../trezor-firmware/python/src/trezorlib hwilib/devices`
  * `git status`: ignore untracked and deleted files, the only diff is in messages.py
    * it's slightly easier to review the changes to that file in this PR, which only adds functions

Since `messages.py` is generated (from what?) with pb2py, it might be easier moving forward to track the original thing?

Sjors:
re-utACK 16b2141a7f69875f607dfb2b594ce837068c4749

achow101:
A couple days ago, trezorlib merged https://github.com/trezor/trezor-firmware/pull/1967, so I'll probably redo this PR on that. It includes several of the modifications that we patch into trezorlib, so it should reduce our diff to upstream.

achow101:
superseded by #554

PR #550, opened 2021-11-23
ledger: Support Bitcoin App 2.0

achow101:
The 2.0 version of the Ledger Bitcoin app uses a completely new protocol which utilizes PSBTv2 and descriptors(ish). Additionally, this new version supports signing Taproot inputs.

This PR adds support for using the new protocol when the app major version number is detected to be greater than or equal to 2. Ledger has provided a python library in https://github.com/LedgerHQ/app-bitcoin-new/tree/master/bitcoin_client. Given that many components of this library are copied from HWI itself, it doesn't make sense to include the entire thing. Additionally, as we already have access to the device via btchip-python, there is no need to use any of the transport things. So instead, parts of the library are copied to `hwilib/devices/ledgerbitcoin2` and modifications are made to use relative imports to common HWI components. Our copy of btchip-python is slightly modified to work with the new version as well, particularly with returning the status code rather than consuming it when receiving APDU responses.

As the new protocol uses PSBTv2, this PR also requires #549 for PSBTv2 itself.

As v2.0.0 supports Taproot, this branch is based on #544 so that Taproot can be supported.


prusnak:
> which utilizes PSBTv2

What is the maximum size of PSBTv2 that can be signed?

Sjors:
I was able to display a taproot address. Haven't tried spending yet.

achow101:
> What is the maximum size of PSBTv2 that can be signed?

I don't think there is a size limit. They appear to be streaming things rather than processing the PSBT as a whole.

gruve-p:
cc @bigspider

bigspider:
Not sure @achow101 wants to refactor at this time given the amount of work he's already done (thanks a lot, by the way!), but I did some work on a standalone python library that also supports the legacy (<2) Ledger app, by integrating the HWI ledger.py code in it. Here is the branch:

https://github.com/LedgerHQ/app-bitcoin-new/tree/pybtc

Still needing a review and docs before release, but it should be pretty stable (main method: [createClient](https://github.com/LedgerHQ/app-bitcoin-new/blob/e9edb893e3d19ae515e449146ce7046221ced0ee/bitcoin_client/ledger_bitcoin/client.py#L222-L231)).
It might simplify a bit the implementation of `ledger.py`, since it handles the legacy support implicitly; therefore `ledger.py` should be able to care about the new protocol only.

As I mentioned to @achow101 on a separate discussion, the current app still supports the legacy APDUs, meaning an unupgraded library or the current version of HWI doesn't see any difference; the plan is to deprecate the legacy APDUs as soon as an updated HWI is out, and remove the support completely on a future release some time in 2022.

bigspider:
Update: library now merged to the [main repository](https://github.com/LedgerHQ/app-bitcoin-new/tree/master/bitcoin_client), and published on pypi as [ledger-bitcoin](https://pypi.org/project/ledger-bitcoin).

achow101:
I've updated this to use the new ledger_bitcoin library (pulled locally, with modifications as usual).

Still need to work out tests since speculos does not come with any 2.0 apps.

bigspider:
@achow101, speculos is now updated with the binaries for the 2.0.1 bitcoin and bitcoin-testnet app (on firmware 2.1).

achow101:
I've had to rebase this onto #565 in order to wrangle the test framework to be able to test all the different conditions that this change introduces (e.g. testing for legacy api, testing around the new restrictions on inputs.

achow101:
I think I've finally worked out the tests. This should be working now.

Sjors:
I tried to display a `wsh(multi(...))` address using its descriptor and got an error "The Ledger Nano S and X do not support P2SH address display". This happens on master too.

 When calling `signtx` on a PSBT from such a wallet it just echoes the PSBT without an error message. That's with a PSBT generated by Bitcoin Core master branch, I haven't tried with a psbt2 patched version yet, but I assume we want backwards compatibility for a while. But this _does_ work on master.

achow101:
> I tried to display a `wsh(multi(...))` address using its descriptor and got an error "The Ledger Nano S and X do not support P2SH address display". This happens on master too. With a simple p2wpk address it does work.

That is expected behavior.

> When calling `signtx` on a PSBT from such a wallet it just echoes the PSBT without an error message. That's with a PSBT generated by Bitcoin Core master branch, I haven't tried with a psbt2 patched version yet, but I assume we want backwards compatibility for a while. But this _does_ work on master.

Currently the 2.0 implementation does not support multisigs (yes that's a breaking change).

Sjors:
Then why did I get the error? Are we switching the protocol based on the firmware version? I guess we'd have to fall back to 1.0 in case of a multisig (if we can detect that)?

achow101:
> Are we switching the protocol based on the firmware version?

Yes

achow101:
I've got the multisig tests working. One of the limitations of the 2.0 App is that the signer can only be in a multisig once. However our test framework would put the device in multiple times. So there are a few commits to resolve that problem.

Sjors:
When you forget to open the Bitcoin app, `enumerate` normally just returns nothing (which sucks), but now it returns a cryptic error:  "Could not open client or get fingerprint information: unpack requires a buffer of 1 bytes"

achow101:
> When you forget to open the Bitcoin app, `enumerate` normally just returns nothing (which sucks), but now it returns a cryptic error: "Could not open client or get fingerprint information: unpack requires a buffer of 1 bytes". Maybe there's hope that we can intercept this and return something more useful?

It will now tell you that you are in a Bitcoin app.

> spending from a taproot address fails: "Failed to sign: Signer fingerprint .... does not match any of the inputs" (same failure on master)

Of course it doesn't. Master doesn't have taproot fields.

> verifying a multisig (SegWit) address: "The Ledger Nano S and X do not support P2SH address display" (as before, should probably say "P2SH and P2WSH")

Updated even though it's completely unrelated.

>  spending from multisig (SegWit) address (where the device has one of the keys): does not fail, but also doesn't sign (maybe I misunderstood what you meant by "I've implemented support for multisigs.")

> Or do you mean this PR is now conditional on [bitcoin/bitcoin#21283](https://github.com/bitcoin/bitcoin/pull/21283)? (that didn't fix it for me, but maybe it requires some tweaks to the extern scriptpubkeyman?)

No that PR is unrelated. The ledger requires the xpubs thus HWI requires the global xpubs in a PBST to be filled in. There is no way around this. You will need to use a software which does that, or put it in manually. Bitcoin Core is unable to do this currently, and there is no PR open for it. The tests deal with this by intercepting the PSBT and putting the xpubs in.

Sjors:
> > spending from a taproot address fails: "Failed to sign: Signer fingerprint .... does not match any of the inputs" (same failure on master)
> 
> Of course it doesn't. Bitcoin master doesn't have taproot fields.

That makes sense. I vaguely remember this working before, but I was probably using https://github.com/bitcoin/bitcoin/pull/22558 then. 

> > spending from multisig [...]

> No that PR is unrelated. The ledger requires the xpubs thus HWI requires the global xpubs in a PBST to be filled in. There is no way around this. You will need to use a software which does that, or put it in manually. Bitcoin Core is unable to do this currently, and there is no PR open for it. The tests deal with this by intercepting the PSBT and putting the xpubs in.

I see. Even if we had a PR on Bitcoin Core it wouldn't be in a release until v24 this fall. Keeping this change unmerged that long doesn't seem like a god idea. So then there's two possibilities:

1. Find a way to fall back to version 1 firmware behavior when there's a multisig.
2. Tell HWI users to stick to 2.0.2 for the rest of the year

In this I'm assuming Specter does add those xpubs, but I haven't been able to test that yet.

achow101:
> When there's a multisig, fall back to the behavior from before this PR

I've added a commit that will do that if a multisig that cannot be signed is detected. However this will not last forever - Ledger is going to remove the legacy protocol, and as I understand it, soon.

> Tell HWI multisig users to stick to 2.0.2 for the rest of the year

And be unable to use Taproot.

The fact of the matter is that Ledger has made a breaking change to their software so users will need to expect that things break if they upgrade. We can communicate that in the release notes, but we cannot try to force the new protocol to do everything the old one did as Ledger has simply removed the ability to do certain things.

Sjors:
I agree that as soon the firmware itself no longer supports the old protocol, there's not much we can do.

achow101:
> I've added a commit that will do that if a multisig that cannot be signed is detected.

Bah! This doesn't work if the psbt has taproot outputs. The legacy protocol doesn't recognize taproot, so it will abort when such an output is seen. This makes an absolute mess of the test suite, so I'm going to remove the test for that behavior.

Sjors:
We don't need the fallback for taproot multisig. That never worked before.

achow101:
> We don't need the fallback for taproot multisig. That never worked before.

Not Taproot multisig. I mean if any of the outputs of the transaction (not signing for them) are a Taproot output.

Sjors:
Mmm, did I really never try sending to taproot? It's possible.

Sjors:
@bigspider is your GPG public key available somewhere? (so I can TOFU it for the next time the library is updated)

Sjors (approved):
tACK b7b06827d79818e02a0346eca67da86e190bd885

Nothing that can't wait for a followup.

I mainly checked that we indeed import the upstream libraries. I only looked shallowly at af0726ced9aed4cd19561b82b15a578cb791a06a which is where we interact with the Ledger API. Perhaps @bigspider can take a closer look at that commit, or `sign_tx` in `ledger.py` in general.  I just hope we have enough test coverage.

bigspider:
> @bigspider is your GPG public key available somewhere? (so I can TOFU it for the next time the library is updated)

I uploaded it here: https://github.com/bigspider/bigspider.github.io/blob/master/pubkey.asc
Although I am not very familiar with practical GPG usage, feel free to drop me a pointer to what I should learn.

hwilib/devices/ledger.py
```diff
@@ -117,209 +144,260 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
 
-            scriptcode = b""
+        For application versions 1.x:
+
+        - Transactions containing both segwit and non-segwit inputs are not entirely supported; only the segwit inputs will be signed in this case.
+
+        For application versions 2.x:
+
+        - Transactions containing OP_RETURN outputs are not supported.
+        - Transacttions containing multisig inputs are currently not supported.
```
bigspider, line 198: Typo: "Transacttions"

Sjors:
@bigspider seems like you're doing fine. Your commits are signed and I can verify that with: `git verify-commit 4e82e44ecfe4ba358da9848087e7e597309abc53`

If you ever run into Bitcoin developers in person, you could ask them to sign your PGP key. It also helps to upload it to a key server, e.g. https://keys.openpgp.org (and maybe link to it on the about page of your website, that's where I first went looking). I also like to leave my PGP fingerprint behind all sorts of places, e.g. on the first slide of a presentation I give, so it ends up on Youtube

hwilib/devices/ledger_bitcoin/btchip/README.md
```diff
@@ -0,0 +1,11 @@
+# Ledger Nano S and Nano X Library
+
+This is a stripped down and modified version of the official [btchip-python](https://github.com/LedgerHQ/btchip-python) library.
+
+This stripped down version was made at commit [17f27c1996c75145b8eb5d16583bddcb6e2bf691](https://github.com/LedgerHQ/btchip-python/tree/17f27c1996c75145b8eb5d16583bddcb6e2bf691).
```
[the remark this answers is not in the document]
achow101: Yes

hwilib/devices/ledger_bitcoin/client.py
```diff
@@ -216,7 +216,11 @@ def createClient(comm_client: Optional[TransportClient] = None, chain: Chain = C
         comm_client = TransportClient("hid", debug=debug)
 
     base_client = Client(comm_client, chain)
-    _, app_version, _ = base_client.get_version()
+    app_name, app_version, _ = base_client.get_version()
+    
+    if not app_name.startswith("Bitcoin"):
+        raise NotSupportedError(0x6A82, None, "Not in a Bitcoin app")
```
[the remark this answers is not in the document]
achow101: That's a side effect of how enumerate works.

hwilib/devices/ledger_bitcoin/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This is a stripped down version of the client provided at https://github.com/LedgerHQ/app-bitcoin-new/tree/master/bitcoin_client.
+
+This stripped down version was made at commit [4e82e44ecfe4ba358da9848087e7e597309abc53](https://github.com/LedgerHQ/app-bitcoin-new/commit/4e82e44ecfe4ba358da9848087e7e597309abc53)
+
+## Changes
+
+As the library originally copied several functions and classes from HWI, the majority of changes were to use relative imports to access those things rather than copying them and duplicating code.
+
+* Relative imports to our `common.py`, `key.py`, `_serialize.py`, `psbt.py`, and `_script.py`.
```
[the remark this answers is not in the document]
achow101: Done

hwilib/devices/ledger_bitcoin/README.md
```diff
@@ -0,0 +1,17 @@
...
+* `write_varint` is an alias for `..._serialize.ser_compact_size` rather than a separate function.
+* Inline `serialize_str` function into `wallet.py`.
+* Inline `bip32_path_from_string` function into `command_builder.py`.
+* Inline `ByteStreamparser` class into `client_command.py`.
+* Removed `btchip/btchipComm.py` as it is not doing anything.
+* Remove usage of `typing.Literal` (needed for Python 3.6 support)
```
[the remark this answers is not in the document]
achow101: Added a mention of `sign_message`. I do not find the other functions in the original, and the tests, README, and license files were not in what I copied, so I don't think that needs mention.

hwilib/devices/ledger_bitcoin/ledgercomm/README.md
```diff
@@ -0,0 +1,9 @@
+# LedgerCOMM
+
+Modified version of the [ledgercomm library](https://github.com/LedgerHQ/ledgercomm).
```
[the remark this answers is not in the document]
achow101: Yes, added the hash.

hwilib/devices/ledger_bitcoin/ledgercomm/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+# Changes
+
+The changes are only to make it work on Python 3.6 as the official library requires 3.8+
```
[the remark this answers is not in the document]
achow101: Done

hwilib/key.py
```diff
@@ -384,3 +384,39 @@ def get_bip44_chain(chain: Chain) -> int:
         return 0
     else:
         return 1
+
+def get_addrtype_from_bip44_purpose(purpose: int) -> Optional[AddressType]:
```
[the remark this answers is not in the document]
achow101: Changed to index

hwilib/devices/ledger.py

hwilib/devices/ledger.py
```diff
@@ -117,209 +147,272 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
-                    assert(tx.inputs[i].non_witness_utxo is not None)
-                    self.app.startUntrustedTransaction(first_input, i, legacy_inputs, script_codes[i], c_tx.nVersion)
-                    self.app.finalizeInput(b"DUMMY", -1, -1, change_path, tx_bytes)
-                    tx.inputs[i].partial_sigs[signature_attempt[1]] = self.app.untrustedHashSign(signature_attempt[0], "", c_tx.nLockTime, 0x01)
-                    first_input = False
-
-        # Send PSBT back
+                self._switch_to_legacy_client()
+                return legacy_sign_tx()
+
```
bigspider, line 409: Wouldn't `_switch_to_legacy_client()` make it impossible to use any non-legacy feature from now on, since it permanently changes `self.client`?
I suppose self.client could be restored after `legacy_sign_tx` returns to avoid that.

hwilib/devices/ledger.py
```diff
@@ -117,209 +147,272 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
+                                break
+                        else:
+                            # No xpub, Ledger will not accept this multisig
+                            ok = False
+                    else:
+                        # No origin info, Ledger will not accept this multisig
+                        ok = False
+
+                if not ok:
+                    retry_legacy = False
```
bigspider, line 302: I think this was meant to be `retry_legacy = True`.

hwilib/devices/ledger.py
```diff
@@ -117,209 +147,272 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
+                    continue
+
+                if our_keys != 1:
+                    # Ledger does not allow having more than one key per multisig
+                    retry_legacy = False
+                    continue
```
bigspider, line 308: Same as above.

hwilib/devices/ledger.py
```diff
@@ -117,209 +147,272 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
 
-            scriptcode = b""
+        For application versions 1.x:
+
+        - Transactions containing both segwit and non-segwit inputs are not entirely supported; only the segwit inputs will be signed in this case.
+
+        For application versions 2.x:
+
+        - Transactions containing OP_RETURN outputs are not supported.
+        - Transactions containing multisig inputs are currently not supported.
```
bigspider, line 201: Anything missing from multisig, or should this line be removed?

hwilib/devices/ledger.py
```diff
@@ -117,209 +147,272 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
+                    continue
+
+                if our_keys != 1:
+                    # Ledger does not allow having more than one key per multisig
+                    retry_legacy = False
+                    continue
+
+                # Make and register the MultisigWallet
+                msw = MultisigWallet(f"{k} of {len(key_exprs)} Multisig", script_addrtype, k, key_exprs)
+                _, registered_hmac = self.client.register_wallet(msw)
```
bigspider, line 312: While the `wallets` variable is a set and therefore each added entry is different even if there are multiple inputs for the same wallet, the `register_wallet` call requires user's interaction. It would be nice to not repeat it for the same wallet, by checking if there is an entry in `wallets` with the same policy map as `msw` before registering.

bigspider (commented):
I went through `sign_tx`; looks generally good, only left some comments related to the handling of multisigs.

bigspider:
> I agree that as soon the firmware itself no longer supports the old protocol, there's not much we can do.

Just to clarify on the legacy (<2.0.0) protocol support: currently the 2.x.x app _also_ replies as expected if some software uses the legacy protocol (it basically still contains the legacy app inside). This is an interim solution to prevent current client software breaking suddenly, but clearly it's not sustainable in the long term.

The plan is to remove this in a future 2.x version later this year, so the new app will only respond to the new protocol.

After that, for anyone that needs to work with any software that is unaware of the new protocol, the simple solution is to make sure that the latest 1.x.x version of the app is kept available in the app store. I think that should be enough to make sure that nobody is locked out of their coins.

hwilib/devices/ledger.py
achow101: Oops fixed.

hwilib/devices/ledger.py
achow101: Fixed

hwilib/devices/ledger.py
achow101: Done.

Sjors:
> the simple solution is to make sure that the latest 1.x.x version of the app is kept available in the app store.

As a sperate app? That sounds like a good enough temporary solution.

bigspider:
> As a sperate app? That sounds like a good enough temporary solution. Once the legacy app becomes available, we could drop the workaround in [ba42b5b](https://github.com/bitcoin-core/HWI/commit/ba42b5be6f772e4b8fdcc7cfacd96e1a837571e3) 

Yes, as a separate app. It will be impossible to install both at the same time (only one app can claim "Bitcoin" as its name, and libraries − including HWI − rely on that for identification), but that should only be a slight annoyance.

Sjors:
We could modify HWI to also recognize bitcoin-legacy? Deleting and reinstalling apps depending on which wallet you're using is not the end of the world, but I imagine it (slightly) increases the risk of bricking the device?

bigspider:
> We could modify HWI to also recognize bitcoin-legacy? Deleting and reinstalling apps depending on which wallet you're using is not the end of the world, but I imagine it (slightly) increases the risk of bricking the device?

I don't think there is any recorded case of bricking the device while installing/uninstalling apps − but indeed no harm is done in relaxing thelibrary's check to leave this option open. I would suggest changing [these two lines](https://github.com/bitcoin-core/HWI/blob/ba42b5be6f772e4b8fdcc7cfacd96e1a837571e3/hwilib/devices/ledger_bitcoin/client_legacy.py#L79-L80) to:
```python
        app_name = self.app.getAppName()
        if app_name != "app" and not (isinstance(app_name, str) and app_name.startswith("Bitcoin")):
            raise ValueError("Ledger is not in either the Bitcoin or Bitcoin Testnet app")
```
(for context, `"app"` is the default name when running within the speculos emulator)

What do you think, @achow101?

Sjors:
Wouldn't that also match "Bitcoin Pizza"? Might be safer to pick a specific name. We can do that in a later release anyway, once Ledger is sure what name to use. HWI can be released much quicker than Bitcoin Core.

achow101:
We can add the name later, once a name is decided.

However, I've just realized the app name detection for the new client doesn't match the legacy client, so I've changed those to match, and removed the redundant check in LegacyClient.

Sjors:
re-ACK daeec1dc3ce6134dbf253bdd95fb83821f4d9bd0

PR #561, opened 2022-01-13
add build scripts w/o gui

KayBeSee:
Adds scripts to build HWI without the QT GUI.

Still needs to remove `{ include = "hwi-qt.py" },` from L14 and `hwi-qt = 'hwilib._gui:main'` from L46 in `pyproject.toml`


achow101:
Instead of a new script, I would prefer if an option were just added to the existing script(s) that would disable the qt build.

PR #565, opened 2022-02-05
tests: Put device specific parameters in Emulator classes

achow101:
Currently we have a bunch of device specific parameters passed into the tests when they are created. However the parameters are almost always the same across test cases, so passing them in individually results in long mostly duplicated lines. Since most tests have a `Emulator` class which manages the emulator, we can instead pack those parameters into the `Emulator` object. This makes the code easier to read, and it allows us to also adjust those parameters at runtime when there are different modes for the emulators. This is particularly important as we implement Taproot as the different combinations of things to test gets much more complicated than can be handled with lookups to global dictionaries. Instead of querying a dictionary and doing comparisons of device type strings, we can directly query the emulator.

There were a few test parameters which are test specific. The way that test parameters are passed in have been updated to allow for test specific parameters instead of having those parameters be available to all tests.

The Coldcard, Jade, and Ledger have also been updated to use a per-test emulator.


achow101:
I've added a commit to put bitcoind management in a class, and another commit to use free ports for the p2p and rpc ports.

achow101:
Added a few more commits to parameterize the multisig tests too. Instead of generating all address types for multisigs, generate the ones we want to test.

PR #580, opened 2022-02-23
tests, trezor: Assert scriptPubKey failure for Trezor in test_big_tx

grdddj:
Soon there will be a security tightening in `Trezor` making sure that device will refuse to sign a transaction trying to spend inputs that do not belong to the wallet (the `scriptPubKey` of the input does not correspond to private keys in the wallet).

Connected PR, that will (most probably) merge these changes into `master` - https://github.com/trezor/trezor-firmware/pull/2081.

Making appropriate changes in `test_big_tx` testcase, that is generating inputs that are not spendable by Trezor's private keys. (Seems that could be the same case as with `coldcard`?) 

To still test Trezor's capability of signing big transaction, we might create inputs corresponding to [slip-14](https://github.com/satoshilabs/slips/blob/master/slip-0014.md) standard, which the wallet (inputted with `all all all...` seed) will be able to sign. That is up to discussion (we could supply the test vectors from our tests).

### DIFF
--- a/test/test_device.py
+++ b/test/test_device.py
@@ -535,6 +535,10 @@ class TestSignTx(DeviceTestCase):
                 self.fail('Big tx did not cause CLI to error')
             if self.emulator.type == 'coldcard':
                 self.assertEqual(result['code'], -7)
+            elif self.emulator.type in ('trezor_1', 'trezor_t'):
+                # Trezor rejects to sign as inputs do not correspond to the wallet's private keys
+                self.assertEqual(result['code'], -13)
+                self.assertIn('Input does not match scriptPubKey', result['error'])
             else:
                 self.assertNotIn('code', result)
                 self.assertNotIn('error', result)


achow101:
That would actually affect `test_sign_tx` where we do external inputs (and not with the expected way to be doing them).

grdddj:
> That would actually affect `test_sign_tx` where we do external inputs (and not with the expected way to be doing them).

It really seems to be affecting them, some cases from `test_sign_tx` are currently failing in our `CI`, where we do run the `HWI` tests - https://gitlab.com/satoshilabs/trezor/trezor-firmware/-/jobs/2123051935

There are some errors like 
```python
self.assertTrue(first_sign_res["signed"]) ... KeyError: 'signed'
```
which are not explicitly mentioning the `scriptPubKey` issue, but are probably caused by it.

I am not sure what is the best resolution, we might discuss it with @matejcik and @mmilata

achow101:
> There are some errors like
> 
> ```python
> self.assertTrue(first_sign_res["signed"]) ... KeyError: 'signed'
> ```
> 
> which are not explicitly mentioning the `scriptPubKey` issue, but are probably caused by it.
> 
> I am not sure what is the best resolution, we might discuss it with @matejcik and @mmilata

I think the only solution is to disable external inputs for trezors entirely. The supported method for external inputs requires proofs which we do not currently support, and I am not sure of the correct way to implement those here.

matejcik:
Trezor will also sign all `EXTERNAL` inputs that already have a valid signature. So you can sign anything if Trezor is the last signer. (in case of two Trezor signer, you'd indeed have to exchange ownership proofs)

achow101:
Superceded by #581

PR #581, opened 2022-02-23
trezor: Remove support for external inputs

achow101:
Trezor will be closing the loophole that we use to be able to sign external inputs. So we will need to tell users that they will not be able to sign such transactions if they are using those firmware versions.


grdddj:
I can confirm that the updated `trezor` and `test` code works well with the `ScriptPubKey`-checking Trezors.

All the tests passed when running this branch locally with new (unreleased) Trezor version.

Thanks for the cooperation!

andrewkozlik:
> Trezor will be closing the loophole that we use to be able to sign external inputs.

We decided to allow this loophole if the user disables "safety checks" on their Trezor. Then they will just need to confirm a warning that the scriptPubKey doesn't match the provided path.

andrewkozlik:
For the record, as @matejcik already [mentioned](https://github.com/bitcoin-core/HWI/pull/580#issuecomment-1049037608), Trezor T will accept all `EXTERNAL` inputs that already have a valid signature and `EXTERNAL` inputs that come with a SLIP-19 ownership proof. See also the [documentation](https://github.com/trezor/trezor-firmware/blob/master/docs/common/communication/bitcoin-signing.md#external-inputs) on signing external inputs. Trezor 1 doesn't support `EXTERNAL` inputs at all.

Note that I proposed a PSBT extension for SLIP-19 ownership proofs in https://github.com/satoshilabs/slips/blob/master/slip-0019.md#psbt-bip-174-extension, but we haven't implemented it anywhere yet.

achow101:
> For the record, as @matejcik already [mentioned](https://github.com/bitcoin-core/HWI/pull/580#issuecomment-1049037608), Trezor T will accept all `EXTERNAL` inputs that already have a valid signature

Signature or scriptSig? AFAICT, there's only a scriptSig field. As HWI does not contain a finalizer, this is not necessarily useful for us.

PR #582, opened 2022-02-25
fix minor spelling in ledger comments

instagibbs:


achow101:
These typos ended up being fixed in #550

PR #586, opened 2022-03-04
Add support for Ledger Nano S Plus

gruve-p:
https://shop.ledger.com/products/ledger-nano-s-plus
https://github.com/LedgerHQ/ledgerjs/blob/master/packages/devices/src/index.ts#L55-L64
https://github.com/LedgerHQ/udev-rules/blob/master/20-hw1.rules#L13-L14


prusnak:
Do we need another column in `docs/devices/index.rst`? Maybe just replace "Ledger Nano S" with "Ledger Nano S [Plus]" in the column header?

PR #593, opened 2022-04-04
jade: Pass utxo satoshi value to Jade on signing

JamieDriver:
Pass the utxo value, along with the prior tx (non-witness-utxo) when it
is available.  Most of the time Jade requires the prior tx but there are
cases when it can be omitted if the sats value is passed.
Pass both whenever possible - the precise logic of which is required can
remain contained in the Jade fw.
(Improves compatibility with PSBT which does not always contain
'non-witness-utxo' for segwit inputs.)


achow101:
Please rebase now that CI is fixed.

JamieDriver:
Rebased - cheers.

achow101:
ACK 90a12a45e498d2677ab94fbc49bc6c1dd3c9e457

PR #598, opened 2022-04-21
Tapsigner

scgbckbone:
* add Tapsigner to HWI
* add Tapsigner simulator 
* testing
* adjust README.md in tests

### DIFF
--- a/hwilib/devices/tapsigner.py
+++ b/hwilib/devices/tapsigner.py
@@ -1,0 +1,309 @@
+"""
+Tapsigner
+********
+"""
+import os
+import pyaes
+import getpass
+import datetime
+from base64 import b64encode
+from functools import wraps
+from typing import List, Dict, Any, Union, Callable
+from cktap.transport import find_cards, CKTapUnixTransport, CKTapCard, CardRuntimeError  # type: ignore
+from cktap.utils import ser_compact_size, pick_nonce, path2str  # type: ignore
+from cktap.bip32 import InvalidKeyError  # type: ignore
+from ..common import Chain, AddressType, hash256, hash160
+from ..key import ExtendedKey
+from ..psbt import PSBT
+from ..descriptor import MultisigDescriptor
+from ..errors import UnavailableActionError, BadArgumentError, NO_PASSWORD, NoPasswordError, UnknownDeviceError
+from ..hwwclient import HardwareWalletClient
+from .._serialize import ser_sig_der
+
+
+def tapsigner_exception(f: Callable[..., Any]) -> Callable[..., Any]:
+    @wraps(f)
+    def func(*args: Any, **kwargs: Any) -> Any:
+        try:
+            return f(*args, **kwargs)
+        except (ValueError, InvalidKeyError, AssertionError) as e:
+            raise BadArgumentError(str(e))
+        except (RuntimeError, CardRuntimeError) as e:
+            raise UnknownDeviceError(str(e))
+    return func
+
+
+def is_cvc_length_range(cvc: str) -> bool:
+    """CVC has to be at least 6 characters long and at most 32 characters long"""
+    return 6 <= len(cvc) <= 32
+
+
+def find_card_by_path(path: str) -> CKTapCard:
+    # path is actually an ident
+    for card in find_cards():
+        if path in card.card_ident:
+            return card
+    raise RuntimeError("Card not found! Is it on the reader?")
+
+
+class TapsignerClient(HardwareWalletClient):
+    # In path variable we store card ident
+    # In password variable we store card CVC
+    def __init__(self, path: str, password: str = "", expert: bool = False, chain: Chain = Chain.MAIN) -> None:
+        super(TapsignerClient, self).__init__(path, password, expert, chain)
+        self.device = find_card_by_path(path=path)
+        if not password:
+            raise NoPasswordError('Password (cvc) must be supplied for Tapsigner')
+
+    @tapsigner_exception
+    def get_master_fingerprint(self) -> bytes:
+        """
+        m fingerprint
+        """
+        extended_key_m = self.get_pubkey_at_path("m")
+        fingerprint = hash160(extended_key_m.pubkey)[:4]
+        return fingerprint
+
+    @tapsigner_exception
+    def get_pubkey_at_path(self, bip32_path: str) -> ExtendedKey:
+        xpub = self.device.derive_xpub_at_path(self.password, fullpath=bip32_path)
+        return ExtendedKey.deserialize(xpub)
+
+    @tapsigner_exception
+    def sign_tx(self, tx: PSBT) -> PSBT:
+        """
+        Tapsigner blind signing.
+
+        Tapsigner can only sign paths that have no more than 8 hardened components. If path contains
+        non-hardened components, they must be after hardened ones (cannot have hardened component after non-hardened).
+        Length of non-hardened components is limited to 2.
+        """
+        # Get the master key fingerprint
+        master_fp = self.get_master_fingerprint()
+
+        sighash_tuples = tx.get_sighash_tuples(master_fp=master_fp)
+        if sighash_tuples is None:
+            return tx
+
+        for sighash, keypath, i_num, pubkey in sighash_tuples:
+            int_path = keypath.path
+            rec_sig = self.device.sign_digest(cvc=self.password, slot=0, digest=sighash, fullpath=path2str(int_path))
+            assert len(rec_sig) == 65  # recoverable signature
+            # ignore header byte
+            der_sig = ser_sig_der(rec_sig[1:33], rec_sig[33:65])
+            # add sigs to tx
+            tx.inputs[i_num].partial_sigs[pubkey] = der_sig
+
+        return tx
+
+    @tapsigner_exception
+    def sign_message(self, message: Union[str, bytes], bip32_path: str) -> str:
+        message = message.encode('ascii') if not isinstance(message, bytes) else message
+        xmsg = b'\x18Bitcoin Signed Message:\n' + ser_compact_size(len(message)) + message
+        md = hash256(xmsg)
+        rec_sig = self.device.sign_digest(cvc=self.password, slot=0, digest=md, fullpath=bip32_path)
+        sig = str(b64encode(rec_sig), 'ascii').replace('\n', '')
+        return sig
+
+    def display_singlesig_address(self, bip32_path: str, addr_type: AddressType) -> str:
+        """
+        The Tapsigner does not have a screen to display addresses on.
+
+        :raises UnavailableActionError: Always, this function is unavailable
+        """
+        raise UnavailableActionError('The Tapsigner does not have a screen to display addresses on')
+
+    def display_multisig_address(self, addr_type: AddressType, multisig: MultisigDescriptor) -> str:
+        """
+        The Tapsigner does not have a screen to display addresses on.
+
+        :raises UnavailableActionError: Always, this function is unavailable
+        """
+        raise UnavailableActionError('The Tapsigner does not have a screen to display addresses on')
+
+    def wipe_device(self) -> bool:
+        """
+        The Tapsigner does not support wiping via software.
+
+        :raises UnavailableActionError: Always, this function is unavailable
+        """
+        raise UnavailableActionError('The Tapsigner does not support wiping via software')
+
+    @tapsigner_exception
+    def setup_device(self, label: str = "", passphrase: str = "") -> bool:
+        """
+        Setup the device.
+
+        :param label: Value is ignored
+        :param passphrase: Value is ignored
+        """
+        args: Dict[str, Union[int, bytes]] = dict(slot=0)
+        args['chain_code'] = hash256(os.urandom(128))
+        ses_key, resp = self.device.send_auth('new', self.password, **args)
+        if "error" in resp:
+            raise RuntimeError(resp["error"])
+        return True
+
+    @tapsigner_exception
+    def restore_device(self, label: str = "", word_count: int = 24) -> bool:
+        """
+        The Tapsigner does not support restoring via software. This method is used to
+        decrypt previously created backup with 'backup_device' command. All data it needs
+        are asked via command prompt. CAUTION: Secret data are dumped to stdout, so
+        make sure you know what you're doing.
+
+        :param label: Value is ignored
+        :param word_count: Value is ignored
+        """
+        warn = input("This operation will dump secret data (xprv) to command line. "
+                     "Are you sure you want to continue?[y/n]")
+        if warn.lower() == "y":
+            print(f"Possible backup files in cwd: {[i for i in os.listdir() if i.endswith('.aes')]}")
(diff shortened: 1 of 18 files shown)


prusnak:
Please rebase on top of current master to remove the merge commit.

scgbckbone:
> Please rebase on top of current master to remove the merge commit.

will do squash and rebase after this is reviewed and ready to go - I assume this will need some more commits (build is failing + implementation of review comments)

scgbckbone:
Few notes here. We use password field for CVC, which is something like pin code but you have to provide it for every auth command. Therefore it is not suitable for us to use PIN (with sendpin/promptpin) as PIN does not unlock the card for some amount of time as with Coldcard or Trezor, instead we would need to use sendpin before every command which is not very practical to do.

* added togglepassphrase command which is interactive way how to change CVC. (keep in mind passphrase is CVC for Tapsigner and there is no way how to add bip39 passphrase to the card so I think it is ok)
* added restore command which will not restore the card as it is not possible, but is the way how to decrypt backup file with card a dump xprv to console (unsafe but I have a warning there)
* I have tested this with core v23.0 with all supported script types, you can check [here](https://blockstream.info/testnet/tx/02500d17a97c25a0153d457319ab2b676f4641e7529ce150cede90e4f6adea20) and [here](https://blockstream.info/testnet/tx/90bb23cb2420aae15f66f9c4f979207b394bfff8466cd8a59fe8b462ad1b835d) (signer is tapsigner simulator)
* with core you need to specify signer with CVC(password) on command line like this `-signer='./hwi.py -p 123456'` which definitely has some security implications for the user (but you still need to tap with the physical card).
* I have squashed and rebased on master as @prusnak proposed

Any review is greatly appreciated.

achow101:
Please rebase now that CI is fixed.

scgbckbone:
rebased on current master

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,474 @@
...
+    # In password variable we store card CVC
+    def __init__(self, path: str, password: str = "", expert: bool = False, chain: Chain = Chain.MAIN) -> None:
+        super(TapsignerClient, self).__init__(path, password, expert, chain)
+        self.device = find_card_by_path(path=path)
+
+    def interactive_get_cvc(self) -> None:
+        if not self.password:
+            cvc = getpass.getpass("Please provide CVC:\n")
+            if cvc:
+                self.password = cvc
```
achow101, line 83: We prefer to validate that the passphrase was provided in `__init__` rather than during the individual functions and interactively retrieving it. Having this optional interactivity breaks our model of non-interactivity for most functions.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,474 @@
...
+        if not non_hardened:
+            self.device.set_derivation(path=bip32_path, cvc=self.password)
+            xp = self.device.get_xpub(self.password)
+            return ExtendedKey.deserialize(xp)
+        else:
+            str_path = path2str(hardened)
+            self.device.set_derivation(path=str_path, cvc=self.password)
+            xp = self.device.get_xpub(self.password)
+            ext_key = ExtendedKey.deserialize(xp)
+            return ext_key.derive_pub_path(non_hardened)
```
achow101, line 119: Does the Tapsigner only do hardened derivation? I don't generally like to have HWI do path validation and would much rather the device do it and let it give us an error.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,474 @@
...
+            continue
+        if not card.is_tapsigner:
+            continue
+        d_data: Dict[str, Any] = {}
+        d_data['path'] = card.card_ident
+        d_data['type'] = 'tapsigner'
+        d_data['model'] = 'tapsigner'
+        d_data['label'] = None
+        d_data['needs_pin_sent'] = False
+        d_data['needs_passphrase_sent'] = False
```
achow101, line 459: Since `-p` must be specified, this should be `True`.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,474 @@
...
+
+        if isinstance(card.tr, CKTapUnixTransport):
+            d_data['model'] += '_simulator'
+        if password:
+            # this is kind of unfortunate that we need password (a.k.a CVC) to get master fp
+            # because when using it with core, one has to provide password on the command line
+            # -signer='/home/more/PycharmProjects/HWI/venv/lib/python3.8/site-packages/hwi.py -p 123456'
+            d_data['fingerprint'] = card.get_xfp(password).hex().lower()
+        else:
+            d_data['fingerprint'] = None
```
achow101, line 469: If the password is not provided, this should set an `error` and `code` and omit `fingerprint`, as we do for the digitalbitbox and trezors which have a similar issue.

hwilib/devices/tapsigner.py
scgbckbone: check for password in init - raise if not provided; remove interactive_get_cvc

hwilib/devices/tapsigner.py
scgbckbone: yes, tapsigner only does hardened derivation... I moved the code to `coinkite-tap-protocol` https://github.com/coinkite/coinkite-tap-proto/pull/24 - after it is merged I will update `pyproject.toml` and `setup.py` with current version (for now there is a workaround)

scgbckbone:
https://github.com/coinkite/coinkite-tap-proto/pull/24 merged

scgbckbone:
updated `poetry.lock` with  `poetry update && poetry lock` after addition of `coinkite-tap-protocol`

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,408 @@
...
+        fingerprint = hash160(extended_key_m.pubkey)[:4]
+        return fingerprint
+
+    def get_pubkey_at_path(self, bip32_path: str) -> ExtendedKey:
+        """
+        Get the public key at the BIP 32 derivation path.
+
+        :param bip32_path: The BIP 32 derivation path
+        :return: The extended public key
+        """
```
achow101, line 73: Docs don't need to be duplicated.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,408 @@
...
+        xpub = self.device.derive_xpub_at_path(self.password, fullpath=bip32_path)
+        return ExtendedKey.deserialize(xpub)
+
+    def sign_tx(self, tx: PSBT) -> PSBT:
+        # completely stolen and adjusted from digitalbox (blind signing)
```
achow101, line 78: Would it be possible for you to refactor this so that the sighashing code is shared rather than copy-pasting.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,408 @@
...
+                    # Create tuples and add to List
+                    tup = (sighash, keypath.path, i_num, pubkey)
+                    sighash_tuples.append(tup)
+
+        # Return early if nothing to do
+        if len(sighash_tuples) == 0:
+            return tx
+
+        for sighash, int_path, i_num, pubkey in sighash_tuples:
+            rec_sig = self.device.sign_digest(cvc=self.password, slot=0, digest=sighash, fullpath=path2str(int_path))
```
achow101, line 184: What kind of exceptions will this throw? We need to be able to handle device exceptions and convert them into HWI exceptions so that the correct error codes come out. Usually this is done through decorators, see the other devices for an example.

achow101:
Also, rebase for CI to be not as dead.

hwilib/devices/tapsigner.py
```diff
@@ -0,0 +1,408 @@
...
+        if not password:
+            raise NoPasswordError('Password (cvc) must be supplied for Tapsigner')
+
+    def get_master_xpub(self, addrtype: AddressType = AddressType.WIT, account: int = 0) -> ExtendedKey:
+        bip44_pupose = get_bip44_purpose(addrtype)
+        bip44_chain = get_bip44_chain(self.chain)
+        path = f"m/{bip44_pupose}h/{bip44_chain}h/{account}h"
+        return self.get_pubkey_at_path(path)
+
+    def get_master_fingerprint(self) -> bytes:
```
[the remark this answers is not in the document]
scgbckbone: this has to stay, otherwise `0xffffffff`

hwilib/devices/tapsigner.py
scgbckbone: added `get_sighash_tuples` as a member of `PSBT` class

hwilib/devices/tapsigner.py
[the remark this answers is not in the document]
scgbckbone: limitation of the Coinkite cards: XFP's are not calculated during derivation. `0xffffffff` as a placeholder

scgbckbone:
not sure about CI --> seems to be full of random failures

hwilib/devices/tapsigner.py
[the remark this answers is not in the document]
scgbckbone: real card placeholder is `0x00000000` simulator had `0xffffffff` --> now fixed and both have real thing and simulator has `0x00000000`

achow101:
Please squash/adjust so that each commit passes. Refactors should be in their own commit at the beginning.

achow101:
I'm a bit concerned about the extra dependencies that this introduces. Up to know, HWI has only interacted with devices over USB, and the USB dependency stack is relatively well understood and under control. However this now introduces a completely new dependency stack to deal with NFC readers. Does this require us to ship new driver dependencies such as libusb? I'm highly hesitant to introduce a new device that is the only one that is going to use new driver dependencies.

scgbckbone:
[52aa685f95b01a9480c0fb07acb7e433dc98ef2a](https://github.com/bitcoin-core/HWI/commit/52aa685f95b01a9480c0fb07acb7e433dc98ef2a) Moved Tapsigner and its dependencies to extras. This aims to partly alleviate your concerns with regards to `sdist` and `wheel` distributions. In sdist and wheel, only references to dependencies are stored. With normal `pip install ...` it won't install Tapsigner and its dependencies - so user has to specify `.[tapsigner]` or `-E tpsigner` to actually install/use it.

In this revision, however I still kept Tapsigner in binary distributions as I was not sure how "bit concerned" you are. I can remove Tapsigner from `build_{bin,wine}.sh` meaning that it would only ship by wheel and sdist not binaries. Or make separate binaries with tapsigner support? not sure what is your opinion on this @achow101 ?

scgbckbone:
> [52aa685f95b01a9480c0fb07acb7e433dc98ef2a](https://github.com/bitcoin-core/HWI/commit/52aa685f95b01a9480c0fb07acb7e433dc98ef2a) Moved Tapsigner and its dependencies to extras. This aims to partly alleviate your concerns with regards to `sdist` and `wheel` distributions. In sdist and wheel, only references to dependencies are stored. With normal `pip install ...` it won't install Tapsigner and its dependencies - so user has to specify `.[tapsigner]` or `-E tpsigner` to actually install/use it.
> 
> In this revision, however I still kept Tapsigner in binary distributions as I was not sure how "bit concerned" you are. I can remove Tapsigner from `build_{bin,wine}.sh` meaning that it would only ship by wheel and sdist not binaries. Or make separate binaries with tapsigner support? not sure what is your opinion on this @achow101 ?

@prusnak @Sjors any opinions on this ?

Sjors:
I don't understand Python dependencies that well. But if a lot a extra stuff is needed for NFS, can we (initially) have the user install those seperately and not fail if they're missing?

scgbckbone:
> I don't understand Python dependencies that well. But if a lot a extra stuff is needed for NFC, can we (initially) have the user install those separately and not fail if they're missing?

This is already done with regards to python dependencies (tapsigner dependencies are part of extras). Namely `coinkite-tap-protocol` and `pyscard`. Not installed by default.

My question is whether you would accept additional dependencies (`swig` `libpcsclite-dev`) in binaries. Currently these two additional dependencies are needed for tapsigner to work. Other option is to not provide tapsigner support in binaries at all. User would need to install `swig` and `libpcsclite-dev` on their own and can only use `sdist` and `wheel` distributions with tapsigner explicitly specified as extra during installation (see above)

scgbckbone:
Fair points. I will remove NFC dependencies from binaries and therefore `tapsigner` support will be purely optional and only available via `sdist` and `wheel`. Need to make tests work somehow. Thanks for your input @Sjors

scgbckbone:
@Sjors I use HID Omnikey 5022 CL (if you want to play with coinkite NFC cards on desktop)

Sjors:
I ordered the ACR122U because [a guy Reddit said so](https://www.reddit.com/r/linuxquestions/comments/kgil9c/comment/gggg5th/?utm_source=reddit&utm_medium=web2x&context=3).

scgbckbone:
his reditt post is for `libnfc` which isn't what we use to talk to the reader (`sclite`) ... so doesn't apply

Seems like the one you choose coinkite lists as "not recommended" here https://github.com/coinkite/coinkite-tap-proto/blob/master/README.md#requirements 

Check coinkite store https://store.coinkite.com/store/category/accessories correct readers are available and there is also 25% discount today

Sjors:
Oh man, why are there multiple NFC libraries for Linux?

prusnak:
> Check coinkite store

"IMPORTANT: Incompatible with Mk4 NFC features! Works only with NFC cards."

Wow, so Coldcard and Tapsigner use incompatible NFC standards?

scgbckbone:
Tapsigner is now completely optional and not included in binary distributions. Tests are handled to not fail if extras are not installed.

achow101:
I've spent a while thinking on this and I've come to the conclusion that NFC cards don't really fit into HWI's expected workflow. With USB devices,there is a persistent connection between the device and the computer by virtue of having to plug it in. However with NFC, the connection is really ephemeral with the user expecting to be prompted to tap the card to perform an action. Using a NFC card with HWI would require having to but the card on a reader as if it were plugged in, but this is not how NFC is generally used. HWI is often used in a way where the type of device being used is unknown at the time of command invocation. Users either call `enumerate` to figure that out before issuing the command, or they use the fingerprint and let HWI do it automatically. Both of these require that the device is connected prior to the command being invoked. With NFC cards, this would require the caller to prompt the user to place their card on the reader and to not move it for some time. I don't think users expect to need to do that, nor for callers to know the type of device in order to make such a prompt.

Additionally, I question whether NFC cards would be used with HWI. HWI is a software targeted at desktop and laptop computers. Unless the user purchases a NFC reader, desktops and laptops are unlikely to be able to even interface with the card. ISTM the vast majority of users will only be able to use NFC from their phones, but HWI isn't expected to work on those devices. So I'm not sure that including Tapsigner support is worth the additional maintenance burden it will incur.

tl;dr I don't think Tapsigner is a good fit for HWI.

nvk:
> > > Check coinkite store
> 
> > "IMPORTANT: Incompatible with Mk4 NFC features! Works only with NFC cards."
> 
> Wow, so Coldcard and Tapsigner use incompatible NFC standards?

Yes, was unavoidable. The hardware capabilities and the software stack are too far apart.

PR #600, opened 2022-04-29
normalize master fp to lowercase

scgbckbone:


hwilib/commands.py
```diff
@@ -114,7 +114,13 @@ def enumerate(password: str = "") -> List[Dict[str, Any]]:
     for module in all_devs:
         try:
             imported_dev = importlib.import_module('.devices.' + module, __package__)
-            result.extend(imported_dev.enumerate(password)) # type: ignore
+            partial_res = imported_dev.enumerate(password)  # type: ignore
+            # normalize master fingerprint to lowercase
+            for device_dict in partial_res:
+                # if None or other type than string - skip
+                if device_dict['fingerprint'] and isinstance(device_dict['fingerprint'], str):
```
scgbckbone, line 121: not sure if it should raise `ValueError` if master fp is not string

scgbckbone: or even attempt some further normalization? , i.e. `if isinstance(device_dict['fingerprint'], bytes):  device_dict['fingerprint'] = device_dict['fingerprint'].hex()`

scgbckbone:
HWI seems to lack some kind of validation layer - probably on purpose left to device developers

achow101:
IMO `commands.py` is the wrong place to put this. As each device implementation is ostensibly part of the public API, each device's `enumerate` should do the normalization.

However, I don't see which device would return an uppercase fingerprint. They all use `get_master_fingerprint` which always returns bytes, and that is always converted to lowercase hex with the python `bytes.hex()` builtin.

achow101:
Concept NACK. Fundamentally, we shouldn't care about the case-ness and everything should be a comparison of the bytes that are represented by the string.

scgbckbone:
I updated [PR](https://github.com/bitcoin/bitcoin/pull/25019) in core

scgbckbone:
closing this as it is meh, does not seem as good idea as yesterday - core should be able to handle both upper/lower case

PR #606, opened 2022-05-21
Ledger: Use ' instead of h for hardened character in wallet policies

### LINKED ISSUE #605: Ledger app 2.0.4 fails to sign
Several users are reporting signing failures when using Ledgers with Bitcoin app version 2.0.4. The error is:

```
{"error": "('0x6a80', 'Error in <BitcoinInsType.SIGN_PSBT: 4> command', '')", "code": -13}
```

achow101:
Ledgers appear to not like `h` as the hardened character in wallet policies, so change it to use `'`. To make this happen, all of the descriptor `to_string` functions which eventually call `KeyOriginInfo.to_string()` will now take a `hardened_char` argument which defaults to `h`. This allows us to change the hardened char as needed while retaining a default of `h`.

Fixes #605


craigraw:
Confirmed working with a Ledger Nano X on the latest Bitcoin app 2.0.4. Tested importing the xpub, displaying an address, signing a PSBT and signing a message. As previously noted the issue appears to be caused by a breaking change present on the latest Bitcoin and Bitcoin Testnet apps on the Ledger.

Thanks for the quick work on this. Given the popularity of the Ledger, is another HWI release imminent?

achow101:
Yes, there will be a release soon.

PR #625, opened 2022-07-29
trezor: Set p2sh multisig output script type correctly

### LINKED ISSUE #624: Unusual P2SH derivation path causes signing on Trezor to fail
Unchained Capital uses two unusual derivation paths when creating their standard P2SH 2-of-3 multisig wallets: `m/45'/0/0/0` and `m/45'/0'/0'/0`. These derivation paths cause HWI to return an error when attempting to sign with a Trezor. It appears this is new behaviour as of HWI 2.1.0 (it did not occur in previous versions).

To reproduce this, I created a multisig P2SH wallet with a Trezor as one of the signers, wallet descriptor as follows:
```
sh(sortedmulti(2,[c3591455/45h/0/0/0]tpubDErVNgpDZVndkCfPCANKaba95X4cBy4ieJTD7KmrmwbWUpmoSnyXBbw8yMnPySSLmuSSQDvKnaxeawu5Z7ay8EX6uewFnP26KfafTZ2rDBf/<0;1>/*,[ca9a2b19/45h/0/0/0]tpubDFU47kwMQqhh5GZmo5eP7CA68TaxtymGFsrgHNSfr3frVGcFKhcLxVnWV5XeHTbBXoUr3C2Eb9kNNqHPuXekPjcTTvdoqFrLtod4Z5N2apX/<0;1>/*))#stpamxtd
```

Attempting to sign with the Trezor using HWI results in the following error: 
`DataError: Multisig field provided but not expected.` 

I see this error text is generated by the Trezor firmware. The PSBT I was attempting to sign as follows:
```
cHNidP8BAHICAAAAAebA8z/RmQPLAgsN7zQUNCfMDCY0HBfGS5Da22CvGnfCAQAAAAD9////AsERAAAAAAAAFgAUZ+/SZu0/MnKKrqvmNsu5woaKAmK+DgAAAAAAABepFOUJ5ZfKAcojUmEo+pkDu8WeU9Dkh/NBIwBPAQQ1h88E1FWvhwAAAABx093CJKJO2rIKBJGi0OnHSFK0ZiB7D/VLOqLF842JIQL4MkiTlFF8JBb2/dyC2Q6Xns8Z6E4ZQsIXP63CI9tRkRTKmisZLQAAgAAAAAAAAAAAAAAAAE8BBDWHzwSA6SQ7AAAAABC7em5gh2yW9nmKt9rIr51/qKo1hcskuRaZxl8d3X5aAkmF9ugKiObVRw+MIpJw0tg7RLwh+9N35342dLhO3GAgFMNZFFUtAACAAAAAAAAAAAAAAAAAAAEAcgIAAAAB84o3aqcz7Qyy36z30GGG60N+0KvZz7yuSmZbsHrGZEwBAAAAAP7///8CW+QPAAAAAAAWABS0Xmkd4ROVjWRaLN43vC3zFZV3PhAnAAAAAAAAF6kUOVOqrknZCk+RdAp3h4Zwj0Pm60KH8EEjAAEDBAEAAAABBEdSIQLXZNH/vR5DTH5bNp5aLjlkTWmngdqaEsw7BJooqZRCByEDPCSTkW5TI37C6PLv9LHsNIlowndzc4h+mbhGO0Mh+kNSriIGAzwkk5FuUyN+wujy7/Sx7DSJaMJ3c3OIfpm4RjtDIfpDHMqaKxktAACAAAAAAAAAAAAAAAAAAAAAAAAAAAAiBgLXZNH/vR5DTH5bNp5aLjlkTWmngdqaEsw7BJooqZRCBxzDWRRVLQAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAEdSIQMBIXKdDxFTwBSYrNsWKLR1hNYVVcZff75xXJH9cAtqqCEDNu00Oxx4Ta7EnWpxDoqg7SX1hMm1Wq+4JAI60ekWtsdSriICAwEhcp0PEVPAFJis2xYotHWE1hVVxl9/vnFckf1wC2qoHMqaKxktAACAAAAAAAAAAAAAAAAAAQAAAAAAAAAiAgM27TQ7HHhNrsSdanEOiqDtJfWEybVar7gkAjrR6Ra2xxzDWRRVLQAAgAAAAAAAAAAAAAAAAAEAAAAAAAAAAA==
```

I confirmed that the same Trezor signs without problems when using the standard BIP45 P2SH derivation path of `m/45'`.

achow101:
P2SH outputs that are multisigs that belong to the wallet should be PAYTOMULTISIG

Fixes  #624

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -576,6 +576,7 @@ class TrezorClient(HardwareWalletClient):
                         tx.xpub, psbt_out)
                     if is_ms:
                         txoutput.multisig = multisig
+                        txoutput.script_type = messages.OutputScriptType.PAYTOMULTISIG
 
                 # append to outputs
                 outputs.append(txoutput)


craigraw:
Results:

Signing with an `m/45'` derivation still works fine.

Signing with a `m/45'/0/0/0` or `m/45'/0'/0'/0` derivation is now different - the normal confirmation process on the Trezor starts, with the amounts being presented all the way to the confirmation of the locktime. No change detection is in place however, so all amounts must be confirmed. On acceptance of the locktime, HWI returns the error `DataError: Forbidden key path` and signing fails.

In addition, displaying the address using the `--desc` descriptor for all derivations still results in the error `No path supplied matched device keys`.

matejcik:
it is surprising that "forbidden key path" happens after confirmation. It should either die at start, or not at all. This might be a Trezor bug.

Which Trezor model are you using? And is this still the same PSBT as in #624 ?

craigraw:
Yes, same PSBT as before. This is signing with a Trezor One.

Interestingly I just updated to from firmware 1.10.5 to 1.11.1, and now I get `DataError: Forbidden key path` immediately on signing (no confirmations are displayed).

matejcik:
ah, so a fixed bug :)
Trezor now strongly enforces standard bip32 patterns. You will need to lower the safety checks setting to be able to use the nonstandard path.

craigraw:
Thanks for the tip! After applying
```
trezorctl set safety-checks prompt
```

I was able to sign with `m/45'`, `m/45'/0/0/0` and `m/45'/0'/0'/0` paths.

PR #629, opened 2022-09-15
Update Jade api to 0.1.38, fix CI, fix #630

### LINKED ISSUE #630: enumerate function hangs
I am trying to use the BitBox2 wallet with Sparrow in Windows10 but the enumerate function hangs.
I have debugged the code and the problem appears calling 
`jade.get_version_info`
this ends up calling to
`read_cbor_message`
that tries to read from a TCP socket
`# Throws EOFError on end of stream/timeout/lost-connection etc.`
`            message = cbor.load(self)`
and the recv method hangs forever, the EOFError is not thrown
`return self.tcp_sock.recv(n)`
If I remove Jade from the list of devices to be checked, the bitbox2 is correctly enumerated
Do you know why this happens? Does the socket need to be blocking?
The problem is solved by setting the socket to not blocking, but don't know if that breaks the setup for Jade wallet
`self.tcp_sock.setblocking(False)`
in the `connect` method from jade_tcp.py

JamieDriver:
Update jade client library, thanks.

I have updated to the 0.1.38 release, as this includes the changes to address #630 (ie. including the change tested under #631)

### DIFF
--- a/hwilib/devices/jadepy/README.md
+++ b/hwilib/devices/jadepy/README.md
@@ -2,9 +2,8 @@
 
 This is a slightly stripped down version of the official [Jade](https://github.com/Blockstream/Jade) python library.
 
-This stripped down version was made from tag [0.1.32](https://github.com/Blockstream/Jade/releases/tag/0.1.32)
+This stripped down version was made from tag [0.1.38](https://github.com/Blockstream/Jade/releases/tag/0.1.38)
 
 ## Changes
 
 - Removed BLE module, reducing transitive dependencies
-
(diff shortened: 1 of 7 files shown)


JamieDriver:
Moving to WIP - if https://github.com/bitcoin-core/HWI/pull/631 is seen to be good I'll delay this update to the next jade api version which includes that change.

JamieDriver:
CI has some issues (as does master), but looks good from a Jade perspective, I believe.

JamieDriver:
Many thanks.

PR #631, opened 2022-09-19
WIP - jade: use non-blocking socket for the jade simulator

JamieDriver:
When enumerating and looking for the Jade simulator, set the socket to non-blocking mode, as some platforms (Windows?) will wait forever for a response if the simulator is not running.

For #630


JamieDriver:
Successfully tested - see #630.  
Changes sill be included in #629, so closing this item.

PR #634, opened 2022-09-28
Improve gui

giacomocaironi:
This PR tries to improve the GUI look and usability:
* Improve SignMessageDialog, GetXpubDialog, SignPsbtDialog and MainWindow widget size and spacing
* Add buttons to import a psbt from a file and to export it after signing

### DIFF
--- a/hwilib/ui/getxpubdialog.ui
+++ b/hwilib/ui/getxpubdialog.ui
@@ -6,109 +6,111 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1205</width>
-    <height>158</height>
+    <width>1050</width>
+    <height>125</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Dialog</string>
   </property>
-  <widget class="QLabel" name="path_label">
-   <property name="geometry">
-    <rect>
-     <x>320</x>
-     <y>20</y>
-     <width>101</width>
-     <height>31</height>
-    </rect>
+  <layout class="QGridLayout" name="gridLayout">
+   <property name="leftMargin">
+    <number>10</number>
    </property>
-   <property name="text">
-    <string>Derivation Path</string>
+   <property name="topMargin">
+    <number>10</number>
    </property>
-  </widget>
-  <widget class="QLineEdit" name="path_lineedit">
-   <property name="geometry">
-    <rect>
-     <x>430</x>
-     <y>20</y>
-     <width>401</width>
-     <height>32</height>
-    </rect>
+   <property name="rightMargin">
+    <number>10</number>
    </property>
-  </widget>
-  <widget class="QDialogButtonBox" name="buttonBox">
-   <property name="geometry">
-    <rect>
-     <x>1100</x>
-     <y>110</y>
-     <width>91</width>
-     <height>34</height>
-    </rect>
+   <property name="bottomMargin">
+    <number>10</number>
    </property>
-   <property name="focusPolicy">
-    <enum>Qt::NoFocus</enum>
+   <property name="spacing">
+    <number>10</number>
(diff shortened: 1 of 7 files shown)


.python-version
```diff
@@ -1 +0,0 @@
-3.9.7
```
achow101, line 1: This file shouldn't be removed, it's for pyenv to use a specific python version.

giacomocaironi: I would like to use a virtual environment with pyenv, with this file commited I can't. 

In general .python-version should be commited only for an application, not for a library: https://github.com/github/gitignore/blob/e5323759e387ba347a9d50f8b0ddd16502eb71d4/Python.gitignore#L85-L88. I don't think that devs using pyenv should be locked to a specific python version, if HWI itself supports many

giacomocaironi: Anyway if it is a problem I can revert the change, it was only a nice touch (in my opinion) not essential for the PR

giacomocaironi:
@achow101 Any updates?

.gitignore
```diff
@@ -25,3 +25,5 @@ env.bak/
 venv.bak/
 
 .vscode
+
+.python-version
```
achow101: Please put this change in its own commit.

hwilib/_gui.py
```diff
@@ -125,8 +125,25 @@ def __init__(self, client):
...
+                self.ui.psbt_in_textedit.setPlainText(f.read())
+
+    @Slot()
+    def export_psbt_button_clicked(self):
+        filename, _ = QFileDialog.getSaveFileName(self, 'Save file')
+        if filename:
+            with open(filename, 'w') as f:
+                f.write(self.ui.psbt_out_textedit.toPlainText())
+
+
```
achow101: Introducing new buttons should be in a separate commit.

hwilib/_gui.py
```diff
@@ -303,12 +320,12 @@ def __init__(self, passphrase='', chain=Chain.MAIN):
...
-        self.ui.signmsg_button.setEnabled(False)
-        self.ui.display_addr_button.setEnabled(False)
-        self.ui.getkeypool_opts_button.setEnabled(False)
-        self.ui.toggle_passphrase_button.setEnabled(False)
+        self.ui.getxpub_button.setEnabled(True)
+        self.ui.signtx_button.setEnabled(True)
+        self.ui.signmsg_button.setEnabled(True)
+        self.ui.display_addr_button.setEnabled(True)
+        self.ui.getkeypool_opts_button.setEnabled(True)
+        self.ui.toggle_passphrase_button.setEnabled(True)
```
achow101, line 327: I don't think this is correct. We don't want users to be able use these dialogs if no device is selected yet.

hwilib/ui/getxpubdialog.ui
```diff
@@ -6,109 +6,97 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1205</width>
-    <height>158</height>
+    <width>430</width>
+    <height>140</height>
```
achow101, line 10: xpubs are pretty long, so shrinking this makes the xpub display kind of unergonomic.

hwilib/ui/signmessagedialog.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>957</width>
-    <height>350</height>
+    <width>660</width>
```
achow101, line 9: This has caused the close button to be out of the dialog bounds.

giacomocaironi:
The new commit should have fixed everything

achow101:
ACK fbb34a3bcba2873ef60281a48078b81d07a720e1

PR #640, opened 2022-11-07
trezor: optimize signing speed by not serializing transaction in trezor

prusnak:
[ opening as draft PR as it still needs testing , will remove the draft status once it has been tested on all supported devices ]

Since HWI is not using TxRequestSerializedType.serialized_tx we might ask the device not to serialize transactions by setting SignTx.serialize=False

This field is introduced in trezorlib 0.13.4. This PR only backports this field to local copy of trezorlib and nothing else.


achow101:
ACK 021c7de159fb3752b1a1986e116415273d2ed5be

PR #641, opened 2022-11-07
ci: trezor build requires rust nightly

prusnak:
I just noticed that CI fails while building Trezor from master since the build requires Rust Nightly since August 2022 - https://github.com/trezor/trezor-firmware/pull/2461

@matejcik can you double check my change please?


matejcik:
LGTM :+1:

achow101:
Noticed this last week, have a fix in https://github.com/achow101/hardware-wallet-interface/tree/fix-ci-2022-10 which I'm hoping to open a PR for this week to fix all of the current CI issues. Still have a few more to work through though.

***

Trezor T Sim Builder is still failing.

achow101:
Superseded by #642

PR #642, opened 2022-11-08
ci: Various Fixes Nov 2022

achow101:
Periodic CI fixing

* Remove unneeded `type: ignore` that is causing the type checker to fail
* Update Trezor T Sim Builder to use rust nightly
* Update Coldcard patches to apply cleanly to latest Coldcard firmare
* Ensure a recent protoc is being used since there has been a breaking change in python's protobuf implementation
* Add additional things for Ledger Speculos automation to automatically accept

Note that this does not fix CI entirely. There is still a failure in Ledger's `test_sign_msg` that appears to actually be an issue with our implementation and the new message signing implementation for Ledgers. This will be fixed in a followup PR.

### DIFF
--- a/ci/cirrus.Dockerfile
+++ b/ci/cirrus.Dockerfile
@@ -35,7 +35,6 @@ RUN apt-get install -y \
     libusb-1.0-0-dev \
     ninja-build \
     pkg-config \
-    protobuf-compiler \
     qemu-user-static \
     swig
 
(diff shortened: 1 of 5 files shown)


achow101:
There is currently a failure in the Jade Sim Builder. It's unclear to me whether that's caused by an unclean cache or something else. cc @JamieDriver

JamieDriver:
It looks like when using the cached env, it doesn't pick up the virtualenv python environment, but instead uses the system one, which is not right.  
I will try to reproduce locally and see what gives ...

achow101:
> It looks like when using the cached env, it doesn't pick up the virtualenv python environment, but instead uses the system one, which is not right.

There seems to have been a previous run that timed out: https://cirrus-ci.com/task/5608215019257856. I'm guessing that although it timed out, this failure gets cached which causes the subsequent run to have unfinished setup cached.

bigspider:
> Note that this does not fix CI entirely. There is still a failure in Ledger's `test_sign_msg` that appears to actually be an issue with our implementation and the new message signing implementation for Ledgers. This will be fixed in a followup PR.

That's among the things addressed in #637; note that that will only break in the new app release (v2.1.0); any previous version will still correctly respond to the legacy protocol.

JamieDriver:
@achow101 Thanks for trying that, but the next build that used the work cache exhibited the same issue.

I have I think resolved it in #643 (I made sure to run a CI that got the issue, then a couple more after making the change and these seemed to work...)  - feel free to review and merge that PR, or cherry-pick that commit into this PR, as you prefer.

PR #644, opened 2022-11-14
Use Optional for passphrase

### LINKED ISSUE #639: Trezor: it should be possible to use an empty passphrase
In case a user enters empty passphrase on Trezor, it should work and a passphrase-less wallet should be open (the same as on a Trezor with passphrase protection off).

It seems that HWI disallows empty passphrases leading to confusions such as:
* https://github.com/cryptoadvance/specter-desktop/issues/1946
* https://github.com/sparrowwallet/sparrow/issues/240

achow101:
We should be distinguishing from the empty passphrase and no passphrase provided, so instead of defaulting to the empty passphrase, change passphrase (aka password) to an `Optional`. This way, if it is not provided, it is `None` and empty passphrases can be handled correctly.

Fixes #639

### DIFF
--- a/hwilib/devices/trezor.py
+++ b/hwilib/devices/trezor.py
@@ -289,6 +289,8 @@ class TrezorClient(HardwareWalletClient):
         sim_path: str = SIMULATOR_PATH,
         model: Optional[TrezorModel] = None
     ) -> None:
+        if password is None:
+            password = ""
         super(TrezorClient, self).__init__(path, password, expert, chain)
         self.simulator = False
         transport = get_path_transport(path, hid_ids, webusb_ids, sim_path)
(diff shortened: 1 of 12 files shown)


achow101:
Updated to unify Trezor's enumerate and other commands behavior. Both will require a passphrase to be provided if passphrase protection is on.

hwilib/devices/trezor.py
```diff
@@ -326,6 +326,8 @@ def _check_unlocked(self) -> None:
             self.client.ui.disallow_passphrase()
         if self.client.features.pin_protection and not self.client.features.unlocked:
             raise DeviceNotReadyError('{} is locked. Unlock by using \'promptpin\' and then \'sendpin\'.'.format(self.type))
+        if self.client.features.passphrase_protection and self.password is None:
+            raise NoPasswordError("Passphrase protection is enabled, passphrase must be provided")
```
moneymanolis, line 330: The import for `NoPasswordError` is missing.

hwilib/devices/trezor.py
```diff
@@ -326,6 +326,8 @@ def _check_unlocked(self) -> None:
             self.client.ui.disallow_passphrase()
         if self.client.features.pin_protection and not self.client.features.unlocked:
             raise DeviceNotReadyError('{} is locked. Unlock by using \'promptpin\' and then \'sendpin\'.'.format(self.type))
+        if self.client.features.passphrase_protection and self.password is None:
```
moneymanolis, line 329:
```suggestion
        if self.client.features.passphrase_protection and not self.client.features.unlocked:
```

Otherwise, `./hwi.py -t "trezor" -d "webusb:020:2:4" signtx ... ` doesn't work! This would also break the signing flow in Specter Desktop with a Model T right now.

achow101:
https://github.com/bitcoin-core/HWI/issues/639#issuecomment-1320085824 recommends that we default to the empty string passphrase if no passphrase is provided, so I've updated this to do that. `TrezorClient` will use `""` if no passphrase is given, and `enumerate` will now only emit a warning if no passphrase is provided rather than an error.

moneymanolis:
Just fyi: Works fine with Specter now.

moneymanolis:
@achow101 just checking in, when are you planning a new release with that PR included? The last release was in May, so might perhaps be about time anyway?

achow101:
@moneymanolis I plan on making a release at the beginning of the new year.

moneymanolis:
Great, thanks!

PR #648, opened 2022-11-18
Add buttons to psbt dialog

giacomocaironi:
Add buttons to import/export a psbt from/to a file

### DIFF
--- a/hwilib/_gui.py
+++ b/hwilib/_gui.py
@@ -1,5 +1,6 @@
 #! /usr/bin/env python3
 
+import base64
 import json
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


achow101:
Files containing PSBTs can be both text files that contain the PSBT in base64, or the binary data of the PSBT itself. So "Open file" needs to be able to handle both, and "Save file" should present an option to the user to choose which type to create. For saving, I think we should default to saving as a binary file with a `.psbt` extension as the BIP specifies.

giacomocaironi:
With the new commit the user has the option to decide in which format to load/save a transaction. So there is not a default and the user has to specify the format each time. Is that ok?

hwilib/_gui.py
```diff
@@ -304,7 +351,7 @@ def __init__(self, passphrase='', chain=Chain.MAIN):
 
     def clear_info(self):
         self.ui.getxpub_button.setEnabled(False)
-        self.ui.signtx_button.setEnabled(False)
+        self.ui.signtx_button.setEnabled(True)
```
achow101, line 354: I don't think this is correct.

giacomocaironi: Sorry, I missed it. Now it is fixed

achow101:
ACK a98ac31a4f4b67a4fcdcac33f42ccd3ed97b3a62

PR #649, opened 2022-11-22
Fix Jade CI build pollution

JamieDriver:
It appears that the Jade CI is being polluted by the output of prior runs.  This needs removing, otherwise we're not getting a true clean build for this commit using the current tools, but rather some cached values (now out of date and not applicable) can interfere with the current build.

### DIFF
--- a/test/setup_environment.sh
+++ b/test/setup_environment.sh
@@ -392,8 +392,8 @@ if [[ -n ${build_jade} ]]; then
     # always seem to pick up the locally installed python virtualenv, and instead uses
     # the system python/no-virtualenv which fails ...)
     # Only install the tools we need (ie. esp32)
+    rm -fr "${IDF_TOOLS_PATH}"
     cd esp-idf
-    rm -fr ${IDF_TOOLS_PATH}
     ./install.sh esp32
     cd ..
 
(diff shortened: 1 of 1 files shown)


JamieDriver:
The CI here is fixed by forcing the cmake files to be rebuilt.  


NOTE: the master build is currently failing with:
```
CMake Error at /tmp/cirrus-ci-build/test/work/jade/esp-idf/tools/cmake/project.cmake:296 (__project):
  The CMAKE_CXX_COMPILER:
    /tmp/cirrus-ci-build/test/work/jade/esp-idf-tools/tools/xtensa-esp32-elf/esp-2021r2-patch3-8.4.0/xtensa-esp32-elf/bin/xtensa-esp32-elf-g++
```
Note earlier in the build we see:
```
Installing xtensa-esp32-elf@esp-2021r2-patch5-8.4.0
...
Extracting /tmp/cirrus-ci-build/test/work/jade/esp-idf-tools/dist/xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch5-linux-amd64.tar.gz to /tmp/cirrus-ci-build/test/work/jade/esp-idf-tools/tools/xtensa-esp32-elf/esp-2021r2-patch5-8.4.0
```

ie. this build installs 'esp-2021r2-**patch5**-8.4.0', but because of old/cached cmake files, is trying to use 'esp-2021r2-**patch3**-8.4.0' - which fails because this is not installed.

I can only assume this is happening because these builds are not proper clean builds, but are essentially incrementals (due to the 'work_cache' being copied in) - this is not something we've seen in our own CI.

The fix is to do a 'clean' when building the jade fw image (which is the smallest part of the build anyway, compared to fetching the toolchain repos and building them - which is helped by using the work_cache since we only update them once every few months - the jade 'master' code is usually updated every few days).  In any case the Jade C code is entirely rebuilt anyway - this extra 'clean' will just ensure the cmake files are recreated - minimal additional overhead.

PR #651, opened 2022-12-01
Pass '--expert' and '--chain' cmd-line arguments to enumerate() functions

JamieDriver:
These arguments are currently discarded/ignored for the enumerate command.  Pass them through to the per-hw enumerate() calls.

In the Jade client, use the 'chain' parameter to enable testnet-configured Jade units to enumerate successfully.

### DIFF
--- a/hwilib/_cli.py
+++ b/hwilib/_cli.py
@@ -60,7 +60,7 @@ def displayaddress_handler(args: argparse.Namespace, client: HardwareWalletClien
     return displayaddress(client, desc=args.desc, path=args.path, addr_type=args.addr_type)
 
 def enumerate_handler(args: argparse.Namespace) -> List[Dict[str, Any]]:
-    return enumerate(password=args.password)
+    return enumerate(password=args.password, expert=args.expert, chain=args.chain)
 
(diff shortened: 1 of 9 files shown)


strangebit:
Thanks for this. For the maintainer, this came about because I was trying to set up Jade to use with Sparrow on testnet/signet, but having trouble with the Jade returning `JadeError: -32003 - Network type inconsistent with prior usage (Data: None)`. That was the case even though the Jade was properly previously initalized for testnet using the Blockstream Green Desktop client first.

We had a long discussion about it here: https://github.com/Blockstream/Jade/issues/13#issuecomment-1334089833. You can safely ignore anything above this part of the thread as not being relevant. Then the breakthrough realization here about why Sparrow was finding an error when enumerating devices after clicking 'Scan': https://github.com/Blockstream/Jade/issues/13#issuecomment-1334380743.

JamieDriver:
* We could then only probe for 'simulators' if '--expert' was passed, and make sure the test framework passes that option.

achow101:
ACK f4091f47d92ead716b0b0ce53ed61b567a8fc22e

PR #655, opened 2022-12-09
`--without-gui` for building without GUI support

kiminuo:
Alternative to #561

Addresses https://github.com/bitcoin-core/HWI/pull/561#issuecomment-1027110481:

> Instead of a new script, I would prefer if an option were just added to the existing script(s) that would disable the qt build.

### DIFF
--- a/contrib/build_bin.sh
+++ b/contrib/build_bin.sh
@@ -1,5 +1,6 @@
 #! /bin/bash
 # Script for building standalone binary releases deterministically
+# Usage: First script parameter can be `--without-gui` to build without UI support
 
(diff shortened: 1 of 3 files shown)


KayBeSee:
I looked at this a few weeks ago but was having issues building on my raspberry pi. I think it was an issue installing dependencies, etc and not these code changes. IIRC, I was able to build it fine on my mac.

I will take another look at it this week, but I am really excited you took the time to add this flag!

achow101:
ACK f6d246abdd99a575bccad32a809033d442e97d76

PR #662, opened 2023-02-12
Remove upper bound pin on typing extensions

### LINKED ISSUE #572: update typing-extensions dependency
In setup.py we have `'typing-extensions>=3.7,<4.0'`, but recently version 4.x has been released. We should test whether 4.x version works too and if it does we can relax the dependency to `'typing-extensions>=3.7,<5.0'`

patrick91:
Upper bounds should be done in user land :)

Should fix this issue: https://github.com/strawberry-graphql/strawberry/issues/2533

Closes #572

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ install_requires = \
  'mnemonic>=0,<1',
  'pyaes>=1.6,<2.0',
  'pyserial>=3.5,<4.0',
- 'typing-extensions>=3.7,<4.0']
+ 'typing-extensions>=3.7']
 
 extras_require = \
 {':python_version >= "3.6" and python_version < "3.7"': ['dataclasses>=0.8,<0.9'],


k9ert:
Discussed also in https://github.com/bitcoin-core/HWI/issues/572

achow101:
`setup.py` is automatically generated, our dependencies are managed with poetry in `pyproject.toml`. Please update in there.

> Upper bounds should be done in user land :)

They use semver, so that introduces the possibility of a breaking change in newer major versions, so we should upper bound on the current major version and evaluate whether new major versions are incompatible.

k9ert:
> setup.py is automatically generated, our dependencies are managed with poetry in
> pyproject.toml. Please update in there.

I've created https://github.com/bitcoin-core/HWI/pull/666
Thanks @patrick91 for your effort.
@achow101 i guess you can close this PR.

PR #677, opened 2023-05-30
bitbox02: update to 6.2.0

benma:
6.2.0 was released to put a minimum requirement on hidapi 0.14.0, which includes the fix for this issue:

https://github.com/libusb/hidapi/issues/531

That bug caused hidapi on macOS 13.3 to report 0 as the interface number for all hid devices, which led to the bitbox02 being listed twice instead of once. This breaks some downstream integrations like Sparrow: https://github.com/sparrowwallet/sparrow/issues/972


benma:
Bump @achow101 please merge this and release a new version of HWI :pray:

benma:
@achow101 I don't see why not, as long as the vendored code is not modified and can be updated to newer versions without a hassle. Could you make a PR (or update this PR) to vendor it as you see fit?

achow101:
Superseded by #683

PR #679, opened 2023-06-07
jade: update jade simulator build in line with current jade master

JamieDriver:
The Jade qumu simulator has been updated to qemu 8, which requires changes to the configure options when building from source.


achow101:
ACK d53da52cf48cbbe69be6e0d79a7eda73638e0541

PR #683, opened 2023-07-06
Vendor bitbox02 library

### LINKED ISSUE #675: bitbox02 should not be a hard dependency
There is a hard install-time dependency on bitbox02 currently,

https://github.com/bitcoin-core/HWI/blob/master/pyproject.toml#L27

And, if it's not installed, `hwi-qt` will fail to run, with an error like,

```
  File "/usr/bin/hwi-qt", line 33, in <module>
    sys.exit(load_entry_point('hwi==2.2.1', 'console_scripts', 'hwi-qt')())
  File "/usr/bin/hwi-qt", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 171, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/hwilib/_gui.py", line 32, in <module>
    import bitbox02.util
  File "/usr/lib/python3.10/site-packages/shiboken2/files.dir/shibokensupport/feature.py", line 139, in _import
    return original_import(name, *args, **kwargs)
ModuleNotFoundError: No module named 'bitbox02'
```

But, you actually can use `hwi` without bitbox02, you just need to ignore a warning

```
% hwi enumerate
WARNING:root:No module named 'bitbox02', required for bitbox02. Ignore if you do not want this device.
WARNING:root:No module named 'cbor', required for jade. Ignore if you do not want this device.
   .. proceeds to work ...
```

For a user that does not have a bitbox device, this is an extra dependency and extra code that is not needed.

I don't believe this package has any other direct dependencies which exist solely to support one specific vendor.

For these reasons, I would like to see the `bitbox02` dependency removed, and changed to runtime detection. That is, if you want to use HWI with bitbox, you install `bitbox02` separately, and HWI notices that at runtime and uses it. Otherwise, you can use HWI with other devices without installing `bitbox02`.

Any support for this idea?

achow101:
This brings the bitbox02 implementation in line with the implementations of the other hardware wallets as we include copies of their respective libraries in our source tree rather than just including them as a dependency. Alternative to #677

Also an issue with bitbox02 in the GUI is fixed.

Fixes #675


achow101:
cc @benma

benma:
Thanks. Looking at this now, I see you need to add the deps of the bitbox02 deps manually. This is an obvious disadvantage - if the bitbox02 lib gets new deps or bumps its dep versions to fix issues, this would have to be done here as well.

What is the problem exactly with using a regular dep? The linked issue is not convincing to me:

> For a user that does not have a bitbox device, this is an extra dependency and extra code that is not needed.

But the code is still there, just vendored. 

In the end I think a regular dep is preferable for the above reason, and should also be easier to manage. If you have other reasons to vendor, please let me know and proceed anyway. In that case it would be helpful to add the commands you used to vendor in the commit msg (or in the README) so that updating it to newer versions is easier.

achow101:
The vendoring of device libraries is done for two reasons: 1) to remove unnecessary dependencies and code, and 2) to lock in the exact API that we are using. We have had issues in the past where differing library versions on a user's system have resulted in incompatibiliies and unexpected behavior. Vendoring the device library lets us ensure that the version of the library that ends up being used is the one that we have tested against, and it gives us better control over dependencies.

We mainly want to avoid letting external dependencies pull in even more dependencies that completely blow up our dependency stack which makes it hard to actually know what code is being run and if any of it is malicious.

benma:
Sounds reasonable :+1: Thanks for elaborating. 

If it helps: noiseprotocol is used to encrypt the communication (see http://www.noiseprotocol.org/), protobuf is used to encode/decode the API messages to/from the BitBox02. semver is used to enable/disable support for features depending on the firmware version.

PR #691, opened 2023-07-25
qt: switch to PySide6

achow101:
PySide2 uses Qt5 which is no longer maintained. PySide6 uses Qt6 and is the current version of Qt, so we should switch to that. This also lets us build the GUI for Apple Silicon macs.

TODOs:
* Binary builds (debian oldoldstable doesn't have qt6)

### DIFF
--- a/contrib/generate-ui.sh
+++ b/contrib/generate-ui.sh
@@ -6,7 +6,6 @@ pushd hwilib/ui
 for file in *.ui
 do
     gen_file=ui_`echo $file| cut -d. -f1`.py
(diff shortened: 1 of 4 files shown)


brunoerg:
Tested it on Mac M1 and worked fine.

brunoerg:
You missed to change line 242:
```py
self.ui.path_lineedit.setValidator(QRegularExpressionValidator(QRegExp(r"m(/[0-9]+['Hh]?)+/\*"), None))
```

PR #695, opened 2023-07-26
build: Use more recent Python 3.9 for *nix builds

achow101:
MacOS on Apple Silicon requires a more recent python version, so just bump it to the latest security release of 3.9, and let any 3.9 version when developing in pyenv.


Sjors:
I guess this is fine as long as the user doesn't have multiple patch versions on their system.

PR #696, opened 2023-08-02
use warning instead of deprecated warn

fametrano:
logging.warn has been deprecated since Python 3.3 and one should use logging.warning

see also https://github.com/Blockstream/Jade/pull/77

### DIFF
--- a/hwilib/commands.py
+++ b/hwilib/commands.py
@@ -118,7 +118,7 @@ def enumerate(password: Optional[str] = None, expert: bool = False, chain: Chain
         except ImportError as e:
             # Warn for ImportErrors, but largely ignore them to allow users not install
             # all device dependencies if only one or some devices are wanted.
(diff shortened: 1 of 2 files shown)


hwilib/devices/trezorlib/btc.py
```diff
@@ -261,7 +261,7 @@ def sign_tx(
         prev_txes = {}
 
     if details is not None:
-        warnings.warning(
+        warnings.warn(
```
achow101, line 264: In a5a68961b3bd625c4ab6849f58003a7707d7b55c "restored"

What's the purpose of this commit?

fametrano: It should be logging.warning as logging.warn is deprecated, but for warnings it is warnings.warn.
So I restored the correct one. Please squash and merge as the individual commits have no value

achow101:
ACK b49c14e81494174f59712c5235228d37e0267584

PR #698, opened 2023-08-02
do not initialize Decimal from float

fametrano:
this avoid warnings and prefer explicit proper initialization

See https://github.com/LedgerHQ/app-bitcoin-new/pull/191


achow101:
ACK 2692f7fa32d9acf6df917c97ec066b7a44333af4

PR #701, opened 2023-08-08
fix base58.decode_check()

### LINKED ISSUE #700: Base58 decode_check always returns invalid checksum
As per title. Because the calculated checksum in https://github.com/bitcoin-core/HWI/blob/17d5643b9db240d4fbee3f1614e7e1aeade4e9ac/hwilib/_base58.py#L96-L98 uses the full result of the double sha256 and not just the first four bytes, the comparison to the received checksum always fails.

In https://github.com/bitcoin-core/HWI/commit/674cb08d6ffcecbf2502a05dd4371c1cfc8baa93 HWI switched to using it's own base58 implementation for the Bitbox02. This bug is now causing all attempts to sign a multisig PSBT to fail on the Bitbox02 using HWI 2.3.0 due to use of this function in https://github.com/bitcoin-core/HWI/blob/17d5643b9db240d4fbee3f1614e7e1aeade4e9ac/hwilib/devices/bitbox02.py#L425

cc @benma

benma:
The function used the full sha256 and not just the first four bytes to verify the checksum. As a result, the checksum check always fails.

This breaks the BitBox02 multisig registration, which uses this function to decode xpubs. This broke when it switched to using the internal base58 library over the `base58` dependency in 674cb08d6ffcecbf2502a05dd4371c1cfc8baa93.

Fixes #700.

### DIFF
--- a/hwilib/_base58.py
+++ b/hwilib/_base58.py
@@ -95,7 +95,7 @@ def decode_check(s: str) -> bytes:
     payload = data[:-4]
     checksum = data[-4:]
     calc_checksum = hash256(payload)
-    if checksum != calc_checksum:
+    if checksum != calc_checksum[:4]:
         raise ValueError("Invalid checksum")
     return payload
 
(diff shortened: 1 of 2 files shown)


benma:
ping @achow101. please take a look and merge/release as soon as possible. thanks!

achow101:
ACK c31eed944b742f8145d985c9d11535ee41e06532

PR #709, opened 2023-09-17
add onekey device support

somebodyLi:


hwilib/devices/onekey.py
```diff
@@ -0,0 +1,362 @@
...
+        self.se_ver = se_ver
+        self.backup_only = backup_only
+        self.onekey_version = onekey_version
+        self.onekey_serial = onekey_serial
+        self.bootloader_version = bootloader_version
+        self.serial_no = serial_no
+        self.boardloader_version = boardloader_version
+
+
+DEFAULT_MAPPING.register(OnekeyFeatures)
```
achow101, line 227: I don't think this is correct. I believe this ends up changing `DEFAULT_MAPPING` everywhere once this module is imported and that would change the Features message for Trezors.

The way that this was done for keepkey was to make this modification to the `default_mapping` within each `TrezorModel` in the constructor for the client.

hwilib/devices/onekey.py
```diff
@@ -0,0 +1,362 @@
...
+class OnekeyClient(TrezorClient):
+    def __init__(
+        self,
+        path: str,
+        password: Optional[str] = None,
+        expert: bool = False,
+        chain: Chain = Chain.MAIN,
+    ) -> None:
+        super().__init__(path, password, expert, chain, webusb_ids=USB_IDS, sim_path=ONEKEY_EMULATOR_PATH)
+        self.client._refresh_features = MethodType(_refresh_features, self.client)
```
achow101, line 295: The correct `model` can be passed to the superclass constructor which should remove the need to override `_refresh_features`.

hwilib/devices/onekey.py
```diff
@@ -0,0 +1,362 @@
...
+        if not isinstance(self.client.ui, debuglink.DebugUI):
+            self.client.ui.button_request = MethodType(button_request, self.client.ui)
```
achow101, line 297: What is the reason for overriding `button_request`? It seems the only difference is the string, in which case the original string can instead be generalized rather than duplicating code.

test/README.md
```diff
@@ -73,6 +76,44 @@ $ pipenv install
 $ pipenv run script/cibuild
 ```
 
+## Onekey emulator
+
+### Dependencies
+
+In order to build the Onekey emulator, the [Nix](https://nixos.org) will need to be installed:
```
achow101, line 83: Why is Nix necessary?

424778940z: The Nix is there to ensure build environment is reproducible, thus also ensures firmware build by user and official release matches (the binary part, not the header which contain signatures). As side effect, emulator has to be built with in Nix environment.

Nix also used in Trezor build process ([here](https://github.com/trezor/trezor-firmware/blob/main/ci/shell.nix))

brunoerg: So, is Nix a must? Aren't there other ways to do it?

424778940z: HWI CI test no longer require Nix

424778940z:
Hello @achow101, we are still waiting to merge this PR, we have already rebased to latest commit and addressed questions and issues above. Could you kindly reveiw again base on our latest commit?

brunoerg:
Can you please rebase it?

achow101:
Please squash your commits.

prusnak:
Looking at the code, it seems this PR is not adding anything substantial.

The Onekey devices are just clones of Trezor One and Trezor Model T. They even copied the internal names (T1B1 and T2T1). I am not sure it is really worth complicating the codebase and risking rendering of Trezor support unusable (this PR modifies trezorlib too).

Since Onekey tries to be compatible with Trezor ([their FAQ](https://shop.onekey.so/pages/faq?) mentions "support all third-party clients supported by Trezor"), maybe we should not change HWI at all and let Onekey be detected as Trezor via HWI?

424778940z:
> Looking at the code, it seems this PR is not adding anything substantial.
> 
> The Onekey devices are just clones of Trezor One and Trezor Model T. They even copied the internal names (T1B1 and T2T1). I am not sure it is really worth complicating the codebase and risking rendering of Trezor (and all its clones) support unusable (this PR modifies trezorlib too).
> 
> Since Onekey tries to be compatible with Trezor ([their FAQ](https://shop.onekey.so/pages/faq?) mentions "support all third-party clients supported by Trezor"), maybe we should not change HWI at all and let Onekey be detected as Trezor via HWI?

OneKey Touch/Pro are using different MCU and SE than Trezor T/One, which means hardware and low-level codes are totally different. We also designed and built our own GUI since the screen resolution is different, and we have no interest to have GUI looks the same.

OneKey firmware was forked from Trezor, but I don't think building on top of a fork should count as a counterfeit, that's not how open-source projects work. Many great projects were based on a fork.

Due to the nature of initially forked from Trezor, our protocol is compatible with Trezor. However, it does not work the other way around, since OneKey has many things Trezor does not support. This is why we need a separate implementation to fully support our device, and adding OneKey support to HWI is the first step.

Introducing fewer changes possible was suggested by @achow101. Ideally, we want a separate library path.

rayston92:
> Looking at the code, it seems this PR is not adding anything substantial.
> 
> The Onekey devices are just clones of Trezor One and Trezor Model T. They even copied the internal names (T1B1 and T2T1). I am not sure it is really worth complicating the codebase and risking rendering of Trezor (and all its clones) support unusable (this PR modifies trezorlib too).
> 
> Since Onekey tries to be compatible with Trezor ([their FAQ](https://shop.onekey.so/pages/faq?) mentions "support all third-party clients supported by Trezor"), maybe we should not change HWI at all and let Onekey be detected as Trezor via HWI?

Hi @prusnak 

I totally understand your worries.

I'm a big fan of Trezor—my very first hardware wallet was a Trezor back in 2014.

Yes, a lot of OneKey's firmware was indeed developed on the shoulders of Trezor. We strictly followed Trezor's open-source license and had no intentions beyond that.

It’s worth mentioning that all of OneKey’s app code is completely open-source. We do have plans to fully rebuild our firmware using the RTOS architecture, but that's going to take some time.

To draw an analogy, if Trezor is like Google, then OneKey would be like Xiaomi. Sundar Pichai doesn't mind other manufacturers using Android's code in an open-source manner, because it actually helps Google solidify its position and market share in mobile operating systems. Similarly, I believe OneKey genuinely helps promote Trezor’s communication standards, which is beneficial for Trezor.

And let’s be honest—there's a big, annoying elephant in the room. That’s the one we really need to aim to surpass, right?

achow101:
Closing as part of sunsetting described in #850

PR #715, opened 2023-12-12
add support for Python 3.12

toxeus:
Fedora 39 ships with Python 3.12. To achieve a better out-of-the-box experience, support for Python 3.12 is added. This required `pyinstaller` and `hidapi` to be updated.

`pyinstaller` has dropped support for Python 3.7 because it reached EOL on on 2023-06-27 [1]. The `pyinstaller` update forces us to also drop support for Python 3.7. Considering that HWI is a security-sensitive package it makes sense to stop using EOL software components.

~~`hidapi` is missing a release that works with Python 3.12 [2] so we depend on a git version for now.~~
`hidapi` is missing a release that works on Alpine Linux [2] but this is an orthogonal issue.

[1] https://github.com/pyinstaller/pyinstaller/commit/da33dbf1681afa7541124034a8aa40227353dcf2
[2] https://github.com/trezor/cython-hidapi/issues/168

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11' ]
+        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12' ]
         device:
           - 'trezor-1'
           - 'trezor-t'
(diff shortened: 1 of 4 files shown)


prusnak:
Please revert the hidapi change and just update the dependency in the poetry.lock.

Meanwhile I made the wheels for 3.12 that are now part of the PyPI release - see https://pypi.org/project/hidapi/0.14.0/#files

toxeus:
@prusnak done. I've tested the hidapi cp312 wheels in a Alpine Linux container before where it failed and so I concluded that the git version is required. But I can confirm now that when using the hidapi cp312 wheels in a Debian Bookworm container HWI builds fine.

prusnak (approved):
utACK

achow101:
ACK b1912994f365696064a002b68d498d2b2d82ee62

PR #718, opened 2023-12-26
Improve error handle for `Ledger` devices

brunoerg:
Discovered it while reviewing [#bitcoin/24313](https://github.com/bitcoin/bitcoin/pull/24313). The `ledger_exception` function was not dealing with `ApduException` exceptions. It's basically similar to `BTChipException` and we can use its "error code" to return friendly error messages. 

e.g for `displayaddress`: 

**before**:
`{"error": "Exception: invalid status 0x6985", "code": -13}`
**after**:
`{"error": "display_singlesig_address canceled", "code": -14}`

### DIFF
--- a/hwilib/devices/ledger.py
+++ b/hwilib/devices/ledger.py
@@ -39,6 +39,7 @@ from .ledger_bitcoin.client import (
     LegacyClient,
     TransportClient,
 )
+from .ledger_bitcoin.client_base import ApduException
 from .ledger_bitcoin.exception import NotSupportedError
 from .ledger_bitcoin.wallet import (
     MultisigWallet,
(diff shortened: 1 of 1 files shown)


brunoerg:
cc: @Sjors

Sjors:
@bigspider is there a complete list of (Bitcoin app) error codes published somewhere?

bigspider:
> @bigspider is there a complete list of (Bitcoin app) error codes published somewhere?

[This mapping](https://github.com/LedgerHQ/app-bitcoin-new/blob/d6b3f56bb0957c55a127c250074028d1d4d125c7/bitcoin_client/ledger_bitcoin/exception/device_exception.py#L9-L21) is currently as good as it gets. The `IncorrectDataError` (SW `0x6A80`) is the most common error and it's returned for way too many distinct error situations; hopefully I find some time to improve it with more fine-grained error results.

achow101:
ACK 0d5412a7ea6a2f4043d2bccfd4e0957caffba65f

bigspider:
> 0x5515 is what I get when the app is in screensaver mode (Ledger Nano X, BTC app 2.1.3)

Indeed, that situation used to be managed by the app directly, while now it's handled in the app sdk, so the error code changed. I'll add the error code to the client libraries, thanks!

PR #719, opened 2023-12-26
gui: add chain in the title

brunoerg:
It allows to identify the chain (main, test, etc)
when using the GUI.


achow101:
ACK 7f4aca10ba56349ff5927cf43ddaade3147b1581

PR #722, opened 2024-01-18
Fix typos

cristiantroy:


brunoerg:
No strong opinion but perhaps it is better to fix typos when touching the code in fact?

achow101:
Thank you for your contribution. While this stylistic change makes sense on its own, we prefer to make such changes when the related code is touched for other reasons.

PR #725, opened 2024-02-13
`COLDCARD` edge taproot

scgbckbone:
* enable taproot for EDGE releases
* enumerate result now contains label `edge`
* bump ckcc-protocol to current master HEAD 

- [ ] optimize `git clone` that is run twice (maybe with `git worktree`)
- [ ] fix taproot multisig tests


Sjors:
#849 includes this change (partially)

PR #729, opened 2024-03-26
Add --emulators option to explicitly choose to detect emulators in enumeration and device auto detection

### LINKED ISSUE #653: Don't enumerate emulators unless `--emulator` flag is provided
See https://github.com/bitcoin-core/HWI/issues/652#issuecomment-1340149342 for motivation.

And https://github.com/bitcoin-core/HWI/issues/628.

I think there are other issues reported here which are caused by stalled enumerations where there is something else than an emulator running on a scanned port.

achow101:
This changes the default behavior to ignore emulators unless specified with `--emulators`. Since this is a backwards incompatible change, the major version number is bumped to 3.

Closes #653


achow101:
cc @prusnak @craigraw

craigraw:
LGTM.

I tested:
- Singlesig and multisig signing with a couple of hwws
- With `--emulators` and without while separately opening a server on localhost:9999. With `--emulators` HWI hangs on enumeration, without it does not.

PR #733, opened 2024-04-17
build: Use debian buster for releases

### LINKED ISSUE #732: Build with ubuntu-20.04 rather than ubuntu-latest
The currently released 3.0.0 Linux binaries fail on Ubuntu 20.04 with `version 'GLIBC_2.35' not found`. 

There seems to be no particular reason to require Ubuntu 22.04 or later, and this could be fixed by changing the relevant `runs-on` parameter in `workflow.yml` from `ubuntu-latest` to `ubuntu-20.04`.

achow101:
Fixes #732

### DIFF
--- a/contrib/build.Dockerfile
+++ b/contrib/build.Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bookworm-slim
+FROM debian:buster-slim
 
 SHELL ["/bin/bash", "-c"]
 


molnard (commented):
Yes, thank you!

achow101:
Oops, completely forgot about this.

PR #741, opened 2024-06-05
Add BitBox02 Simulator

asi345:
Recently, a simulator for BitBox02 was implemented. Therefore, we also wanted to add this to HWI repository for automated tests and ease of reference in the future.


benma:
fyi @achow101, I reviewed this before @asi345 made the PR

test/test_device.py
```diff
@@ -130,11 +129,12 @@ def create(*args, **kwargs):
         return c
 
 class DeviceTestCase(unittest.TestCase):
-    def __init__(self, bitcoind, emulator=None, interface='library', methodName='runTest'):
+    def __init__(self, bitcoind, emulator=None, interface='library', methodName='runTest', supports_legacy=True):
```
achow101, line 132: `supports_legacy` should be a property of the emulator, not an argument to the test case. See `supports_taproot`.

test/test_device.py
```diff
@@ -583,21 +583,23 @@ def test_signtx(self):
     # Make a huge transaction which might cause some problems with different interfaces
     def test_big_tx(self):
         # make a huge transaction
-        keypool_desc = self.do_command(self.dev_args + ["getkeypool", "--account", "10", "--addr-type", "legacy", "0", "100"])
+        addr_type = "legacy" if self.supports_legacy else "sh_wit"
+        keypool_desc = self.do_command(self.dev_args + ["getkeypool", "--account", "10", "--addr-type", addr_type, "0", "100"])
```
achow101, line 587: It should be fine to just change the address type to `sh_wit` always instead of using `legacy` for the other devices.

test/setup_environment.sh
```diff
@@ -414,6 +419,35 @@ if [[ -n ${build_jade} ]]; then
...
+        elif [ $LOCAL = $BASE ]; then
+            git pull
+        fi
+    fi
+
+    # Build the simulator. This is cached, but it is also fast
+    docker pull shiftcrypto/firmware_v2:latest
+    # The safe.directory config is so that git commands work. even though the repo folder mounted in
+    # Docker is owned by root, which can be different from the owner on the host.
+    docker run -i --rm --volume ./:/bb02 shiftcrypto/firmware_v2 bash -c "git config --global --add safe.directory /bb02 && cd /bb02 && make -j simulator"
```
achow101, line 448: This script needs to work when run inside a docker container as that is how the tests are suggested to be run locally. IIRC Docker within docker doesn't really work, so I don't think this will.

test/test_device.py
asi345: Okay, we can use this convention. I'll updating that shortly, but then I will also add the relevant field to the other emulators.

test/test_device.py
asi345: Okay, changed the types to `sh_wit`

asi345:
Okay, we also came up with a solution for docker in docker problem for bitbox02 simulator.
Now, a separate `build_bitbox02.sh` script is ran before running hwi container, to build BitBox02 simulator.
See the comments in ci Dockerfile for reference.

test/README.md
```diff
@@ -329,6 +332,38 @@ You also have to install its python dependencies
...
+Clone the repository:
+
+```
+git clone --recursive https://github.com/BitBoxSwiss/bitbox02-firmware.git
+```
+
+Pull the BitBox02 firmware Docker image:
+
+```
+docker pull shiftcrypto/firmware_v2:latest
```
benma, line 356: Replace with 

```
cd bitbox02-firmware
make dockerpull
```

(remove the `cd` below then)

test/setup_environment.sh
```diff
@@ -414,6 +419,35 @@ if [[ -n ${build_jade} ]]; then
...
+
+        if [ $LOCAL = $REMOTE ]; then
+            echo "Up-to-date"
+        elif [ $LOCAL = $BASE ]; then
+            git pull
+        fi
+    fi
+
+    # Build the simulator. This is cached, but it is also fast
+    docker pull shiftcrypto/firmware_v2:latest
```
benma, line 445: use `$(cat .containerversion)` instead of `latest` here too

benma:
@achow101 please approve the workflow and take another look at the changes, would be really nice to have this merged soon. Thanks!

achow101:
Could you rebase so that CI runs the tests with the simulator?

test/README.md
```diff
@@ -329,6 +332,38 @@ You also have to install its python dependencies
...
 ```
 
+## BitBox02 Simulator
+
+### Dependencies
+
+In order to build the BitBox02 simulator, the following packages will need to be installed:
+
+```
+apt install docker.io
```
brunoerg, line 342: Isn't there a way to build it without docker?

benma: Not really, this is how we build and release the BB02 firmware. Replicating the Docker environment outside of it is bound to lead to tons of overhead in maintenance and CI breaks.

podman instead of docker would also work though.

brunoerg: Ok, anyway, I could build and run it. No problem.

test/test_bitbox02.py
```diff
@@ -0,0 +1,109 @@
...
+    return result.wasSuccessful()
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Test BitBox02 implementation')
+    parser.add_argument('simulator', help='Path to simulator binary')
+    parser.add_argument('bitcoind', help='Path to bitcoind binary')
+    parser.add_argument('--interface', help='Which interface to send commands over', choices=['library', 'cli', 'bindist'], default='library')
+    args = parser.parse_args()
+
+    sys.exit(not bitbox02_test_suite(args.simulator, None, None))
```
brunoerg, line 109: I had some problems with this test. It seems you should pass bitcoind and interface in `bitbox02_test_suite ` instead of None, no?

test/setup_environment.sh
```diff
@@ -414,6 +419,36 @@ if [[ -n ${build_jade} ]]; then
...
+            git pull
+        fi
+    fi
+
+    # Build the simulator. This is cached, but it is also fast
+    CONTAINER_VERSION=$(cat .containerversion)
+    docker pull shiftcrypto/firmware_v2:$CONTAINER_VERSION
+    # The safe.directory config is so that git commands work. even though the repo folder mounted in
+    # Docker is owned by root, which can be different from the owner on the host.
+    docker run -i --rm --volume ./:/bb02 shiftcrypto/firmware_v2:$CONTAINER_VERSION bash -c "git config --global --add safe.directory /bb02 && cd /bb02 && make -j simulator"
```
brunoerg, line 449: Got this error running `setup_environment.sh`:
```
+ docker run -i --rm --volume ./:/bb02 shiftcrypto/firmware_v2:42 bash -c 'git config --global --add safe.directory /bb02 && cd /bb02 && make -j simulator'
docker: Error response from daemon: create ./: "./" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```

I think you could use `"$(pwd):/bb02"` instead of `./:/bb02`.

test/test_bitbox02.py
asi345: Yes, you're right. Fixed that now.

test/setup_environment.sh
asi345: This was not an issue in CI, but in local maybe that created problems. Anyways, I changed to `$(pwd)` instead of `.` as you suggested. This should solve the error.

asi345:
Okay, I rebased this branch on top of master. CI is again waiting for your approval.

However, can you in any chance fix `bitcoind` build first? This commit breaks `bitcoind` build in `setup_environment.sh`, since `autogen.sh` file is not existent in bitcoin repo anymore : https://github.com/bitcoin/bitcoin/commit/d71ac768424333b65a6d88c9752cc9c7fdb276f3

brunoerg:
> However, can you in any chance fix bitcoind build first? This commit breaks bitcoind build in setup_environment.sh, since autogen.sh file is not existent in bitcoin repo anymore : https://github.com/bitcoin/bitcoin/commit/d71ac768424333b65a6d88c9752cc9c7fdb276f3

See https://github.com/bitcoin-core/HWI/pull/752

ci/build_bitbox02.sh
```diff
@@ -0,0 +1,9 @@
+docker volume rm bitbox02_volume || true
+docker volume create bitbox02_volume
+CONTAINER_VERSION=$(curl https://raw.githubusercontent.com/BitBoxSwiss/bitbox02-firmware/master/.containerversion)
+docker pull shiftcrypto/firmware_v2:$CONTAINER_VERSION
+docker run -i --rm -v bitbox02_volume:/bitbox02-firmware shiftcrypto/firmware_v2:$CONTAINER_VERSIONs bash -c \
```
brunoerg, line 5:
```suggestion
docker run -i --rm -v bitbox02_volume:/bitbox02-firmware shiftcrypto/firmware_v2:$CONTAINER_VERSION bash -c \
```

hwilib/devices/bitbox02.py
```diff
@@ -252,9 +258,27 @@ def func(*args, **kwargs):  # type: ignore
...
+        except ValueError as e:
+            raise BadArgumentError(str(e))
 
     return cast(T, func)
 
+class BitBox02Simulator():
+    def __init__(self) -> None:
+        self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        ip, port = SIMULATOR_PATH.split(":")
+        self.client_socket.connect((ip, int(port)))
```
brunoerg, line 270: It throw`ConnectionRefusedError` if it wasn't able to connect to the BitBox02 simulator, so we won't be able to use any other simulator.

asi345: As far as I saw, Trezor or Keepkey also call `connect` on their simulators without handling any exceptions or `connect_ex`. Do you have a suggestion how to handle this exactly?

brunoerg: > As far as I saw, Trezor or Keepkey also call connect on their simulators without handling any exceptions or connect_ex. 

See:
```py
def enumerate(password: Optional[str] = None, expert: bool = False, chain: Chain = Chain.MAIN, allow_emulators: bool = False) -> List[Dict[str, Any]]:
    results = []
    devs = hid.HidTransport.enumerate()
    devs.extend(webusb.WebUsbTransport.enumerate())
    if allow_emulators:
        devs.extend(udp.UdpTransport.enumerate())
    for dev in devs:
```

When calling `enumerate` with `--simulators`, Trezor will call `enumerate` from `UdpTransport`. This function will return nothing if it wasn't able to find a device (simulator) so `devs` is empty.

Maybe you could do something like:
```diff
diff --git a/hwilib/devices/bitbox02.py b/hwilib/devices/bitbox02.py
index 53497c6..7cd6929 100644
--- a/hwilib/devices/bitbox02.py
+++ b/hwilib/devices/bitbox02.py
@@ -186,6 +186,8 @@ def enumerate(password: Optional[str] = None, expert: bool = False, chain: Chain
         devs.append(SIMULATOR_PATH)
     for path in devs:
         client = Bitbox02Client(path=path)
+        if allow_emulators and not client.transport:
+            continue
         if path != SIMULATOR_PATH:
             client.set_noise_config(SilentNoiseConfig())
         d_data: Dict[str, object] = {}
@@ -267,7 +269,11 @@ class BitBox02Simulator():
     def __init__(self) -> None:
         self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         ip, port = SIMULATOR_PATH.split(":")
-        self.client_socket.connect((ip, int(port)))
+        self.connected = True
+        try:
+            self.client_socket.connect((ip, int(port)))
+        except:
+            self.connected = False
 
     def write(self, data: bytes) -> None:
         # Messages from client are always prefixed with HID report ID(0x00), which is not expected by the simulator.
@@ -301,7 +307,10 @@ class Bitbox02Client(HardwareWalletClient):
             self.noise_config = CLINoiseConfig()
         else:
             simulator = BitBox02Simulator()
-            self.transport = u2fhid.U2FHid(simulator)
+            if simulator.connected:
+                self.transport = u2fhid.U2FHid(simulator)
+            else:
+                self.transport = None
         self.device_path = path
 
         # use self.init() to access self.bb02.
```

achow101:
The current implementation is causing all other tests to fail.

hwilib/devices/bitbox02.py
asi345: Okay, I did a fix based on this(not the same since this does not pass type checks). Now, in my tests, other devices are also passing the tests.

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

benma:
> The current implementation is causing all other tests to fail.

Please take a look again, @asi345 updated the PR and all checks have passed.

achow101:
ACK 3c3a185200fb92281128d0d84e0697da21f7dc09

PR #746, opened 2024-07-15
Update Ledger udev rules and model IDs

bigspider:
Updated as per https://github.com/LedgerHQ/udev-rules/blob/f474382e370c9fa2a2207e6e675b9b364441aed7/20-hw1.rules, with more generic rules and including support for Ledger Stax and Flex.

Also added the corresponding device IDs for `hwi enumerate`.


craigraw:
@bigspider It's probably a different PR, but would there need to be an additions to the map of `LEDGER_MODEL_IDS` https://github.com/bitcoin-core/HWI/blob/9fbe6beb63d9a3be68aec084fd9ac0bef4756119/hwilib/devices/ledger.py#L76-L80 for Stax (and now Flex) support?

bigspider:
> @bigspider It's probably a different PR, but would there need to be an additions to the map of `LEDGER_MODEL_IDS`
> 
> https://github.com/bitcoin-core/HWI/blob/9fbe6beb63d9a3be68aec084fd9ac0bef4756119/hwilib/devices/ledger.py#L76-L80
> 
> for Stax (and now Flex) support?

Good catch, thank you! I think there's no reason to have two separate PRs: I added the IDs for Stax and Flex in b6e0d334c7240c45f2cb13dfde0f3b18d11e12d3. I tested locally that `hwi enumerate` correctly lists them.

Theschorpioen (approved):

achow101:
ACK b6e0d334c7240c45f2cb13dfde0f3b18d11e12d3

Theschorpioen (approved):

Theschorpioen (approved):

PR #753, opened 2024-09-03
Jade native psbt

JamieDriver:
1. latest jade api to recognise more recently added hw (now available as its own PR, #768)
2. remove code which walked psbt and mapped it into Jade's legacy 'sign_tx' json format - instead pass PSBT directly to Jade for signing, since this has been supported in fw now for well over a year and seems to be working well.

### DIFF
--- a/hwilib/devices/jade.py
+++ b/hwilib/devices/jade.py
@@ -52,6 +52,7 @@ from .._script import (
     parse_multisig
 )
 
+import base64
 import logging
 import semver
 import os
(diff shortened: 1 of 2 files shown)


brunoerg:
Concept ACK

JamieDriver:
Thanks - any idea what's up with the CI ?  Always feel more confident when CI looks good ...

achow101:
> Thanks - any idea what's up with the CI ? Always feel more confident when CI looks good ...

You need to rebase.

JamieDriver:
Probably best superseded by #763

achow101:
We prefer to support as old firmware as possible. This bumps the minimum supported firmware version, but it looks like it would be trivial to keep all of the old code while still using native PSBTs for newer firmware versions.

JamieDriver:
Rebased onto #768 (so this PR implicitly includes that PR), and retained support for legacy signing for old firmwares.
Thanks.

achow101:
Rebase for #768 and also there's an actual CI failure.

```
hwilib/devices/jade.py:380: error: Returning Any from function declared to return "PSBT"  [no-any-return]
```

achow101:
ACK c66dd1afe81cc89c01075bd6325c6b5a784ba06f

PR #757, opened 2024-09-10
allow "hwi signtx -"

8go:
Enhanced `signtx` to accept special character `-` to indicate to read the base64-encoded PSBT from the stdin instead of from the command line. This allows running commands like
`xclip -o -selection clipboard | hwi -t <device> signtx -`.

### DIFF
--- a/hwilib/_cli.py
+++ b/hwilib/_cli.py
@@ -160,7 +160,7 @@ def get_parser() -> HWIArgumentParser:
     getmasterxpub_parser.set_defaults(func=getmasterxpub_handler)
 
     signtx_parser = subparsers.add_parser('signtx', help='Sign a PSBT')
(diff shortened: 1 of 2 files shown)


achow101:
`-stdin` already exists to do that.

8go:
Ok, I did not realize that. Now I do. 

Before I close this PR, a question: 

How about an option `--clipin` to read the PSBT from the clipboard? And a `--clipout` to write the resulting signed PSBT back onto the clipboard. As in:

```
# unsigned PSBT is on the clipboard
hwi -t $DEVICE --clipin --clipout signtx 
# signed PSBT is on the clipboard
```

Would that be a worthwhile PR for me to make?

MatinGhahremani1379 (approved):

PR #759, opened 2024-09-12
build: Use Python 3.9.19

achow101:
3.9.17 wasn't building in the linux arm64 container.


brunoerg (approved):
utACK c19090e38c12a6dd69ebe12ef6949fd9b3766ebf

contrib/build_wine.sh
```diff
@@ -3,7 +3,7 @@
 
 set -ex
 
-PYTHON_VERSION=3.9.7
+PYTHON_VERSION=3.9.19
```
gruve-p, line 6: https://www.python.org/ftp/python/3.9.19/amd64/ does not exists.
Latest on 3.9 would be https://www.python.org/ftp/python/3.9.13/amd64/

achow101: Huh, I thought it worked when I did the rc, but maybe the result I used was from a run before this change was made.

PR #762, opened 2024-09-24
2 new commandline arguments: --clipin and --clipout

8go:
Added 2 new features, implemented as 2 new command line arguments:

- `--clipin`: similar to --stdin, reads arguments from clipboard
- `--clipout`: for selected commands writes output additionally to clipboard
- example use case: `hwi -t $DEVICE --clipin --clipout signtx # works well in combination with e.g. Bitcoin Core wallet`

### DIFF
--- a/hwilib/_cli.py
+++ b/hwilib/_cli.py
@@ -26,6 +26,7 @@ from .common import (
 )
 from .errors import (
     handle_errors,
+    BAD_ARGUMENT,
     DEVICE_CONN_ERROR,
     HELP_TEXT,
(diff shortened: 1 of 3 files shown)


Thompson1985 (approved):

achow101:
Tend towards NACK

AFAIK, there's no guaranteed cross platform way to read the clipboard, and I would rather not have to have some dependency that tries to guess for every possible platform. If you want to use from clipboard, you can use whatever clipboard commands are relevant for your system and pipe the output into HWI.

PR #763, opened 2024-10-11
Jade pip api and native psbt

JamieDriver:
As of `1.0.32`, the jade-client (jadepy) is available on PyPi and so can be `pip install`-ed, and we can remove the local copy of the code.

If preferred, this can supersede #753


JamieDriver:
Hi,
Are you waiting for anything further from me before you can consider this PR ?
I've rebased it onto current master (3.1.0).
Thanks,
J.

achow101:
NACK

We prefer to pull in vendor libraries to have better control over dependencies and to be able to remove things that we do not wish to support. In this particular case, that would be the bluetooth stuff.

JamieDriver:
The pip install does not include the 'Bluetooth stuff' !

But ok, if you prefer to include a copy of the lib code, I'll revisit #753.

PR #767, opened 2024-12-03
add Trezor Safe 3 and Safe 5 to device support matrix table

prusnak:
Trezor Safe 3 and Safe 5 devices are supported but not listed in the device support matrix table

this PR fixes this


achow101:
ACK cd90146c10715076cdb5d4ff6734f3a34341278a

Issue #785, opened 2025-06-04
BIP-388 policy support


Sjors:
We could add a `--bip388-policy` (or just `--policy`) option to the `signtx` and `displayaddress` commands.

For devices that support it, e.g. Ledger, these commands would first register the policy if needed.

Ledger returns an HMAC after registration. Afaik this isn't documented in any BIP. I also don't know if other devices do this. We could add a `--ledger-hmac` argument for the time being and replace it with something more generic later?

bigspider:
In the typical multisig wallet workflow, 'wallet registration' and 'address visualization' are two different UX flows that happen at different times. Clearly separating them would make the API of HWI match common practice. Therefore, I would strongly recommend having an explicit registration command and API.

Not sure if other devices use the proof-of-registration, but most (all?) the devices that support multisig or more need a registration step, as it's not possible to get the same security level without. I believe at least BitBox showed interest in having an explicit registration API (cc @benma).

Ledger's proof-of-registration is not covered by a BIP, but for the purpose HWI's API, I think it would already be a great improvement to consider it a binary blob and leave it vendor-specific what the meaning is, or whether it's used at all.

Sjors:
See also https://github.com/bitcoin-core/HWI/pull/647

achow101:
No device does registration the same way. This seems to me to be more device specific stuff that we are specifically trying to avoid. Given that we can do on-the-fly registration, I don't think this is at all necessary; sure the UX sucks, but until the hardware vendors can standardize and agree on something, I don't want to implement any device specific policy registration things.

Sjors:
The standard could be as simple as: given a BIP388 policy, return up to 255 bytes to re-authenticate later?

PR #647, opened 2022-11-16
[PoC] Wallet policies (and miniscript support)

bigspider:
Hello,

this in an exploratory PR as a proof of concept of how wallet policies could look like in HWI.

Wallet policies [were presented](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-May/020423.html) to the bitcoin-dev mailing list back in May.
A BIP proposal is in preparation [here](https://github.com/bitcoin/bips/pull/1389).

The goals of this PR are to:
- show that adding wallet policies to hwi is relatively straightforward and can be done with non-breaking changes
- argue that wallet policies naturally and conveniently represent the "accounts" of a typical single-signature, multisignature or miniscript-based wallet accounts;
- demo how a typical wallet-policy workflow could look like with hwi;
- showcase specifically the implementation and UX in the Ledger bitcoin application.

Wallet policies are designed to be vendor-agnostic.
In fact, they are ready for miniscript wallets, and they might in my opinion be a convenient, opinionated way of adding miniscript support to HWI.

They are implemented in the Ledger bitcoin app since version 2.1.0, which uses them to support miniscript (on segwit only, at this time).

# Implementation details

Not production-ready!

In this PR, I attempted to add wallet policies as a first-class object, while not breaking compatibility with existing code. 

The three commit add:
- The wallet-policy-related methods in `HardwareWalletClient`: `can_register_wallet_policies`, `register_wallet_policy`, `display_wallet_policy_address`, `sign_tx_with_wallet_policy`.
- Support in CLI within `displayaddress` and `signtx`; added `registerpolicy`.
- The specific implementation of wallet policies for the Ledger client in `ledger.py`.

For subclasses not implementing wallet policies, a compatibility layer guarantees that if the policy falls back into the supported cases (single-signature, or standard multisignature), calls to functions using wallet policies are automatically converted to the corresponding single-sig or multi-sig function to show addresses or sign transactions.

With this approach, software wallets adopting wallet policies, therefore, could use them directly without using any other `HardwareWalletClient` function without wallet policies.

# Demo

In the following, I'd like to walk you through an end-to-end demo of wallet policies that can be tested from this PR's branch using the [Ledger bitcoin testnet app ](https://github.com/LedgerHQ/app-bitcoin-new) version 2.1.0 in the speculos emulator; the app will be released in production within a few weeks.

In all the examples with multiple keys, internal keys are the ones with fingerprint `f5acc2fd`, which is the fingerprint for the default seed used by the speculos emulator.

[...]

PR #791, opened 2025-07-18
Add register command for BIP388 policies

Sjors:
First step towards #785. See also #647.

This adds a generic `register` command for devices that support BIP388 wallet policy registration.

The command takes:

- A policy name.
- A combined multipath output descriptor using expressions such as `/<0;1>/*`.

HWI converts the descriptor into a BIP388 descriptor template and key vector, validates the resulting policy, and fails if the descriptor cannot be represented as a valid BIP388 policy.

Example:

```bash
hwi --device-type ledger register \
  --name "2-of-2 multisig" \
  --desc "wsh(sortedmulti(2,[00000001/48'/0'/0'/2']xpub.../<0;1>/*,[00000002/48'/0'/0'/2']xpub.../<0;1>/*))"
```

Ledger returns an opaque registration value containing the device registration and policy name as separately length-prefixed fields. This lets `signtx` in #792 recover the policy name without a separate `--policy-name` argument:

```json
{"registration": "00..."}
```

BitBox02 stores the policy on the device and therefore returns no registration value:

```json
{"registration": null}
```

The policy conversion tests cover all BIP388 test vectors. Simulator device tests cover registration on both Ledger and BitBox02.

### DIFF
--- a/.github/workflows/type-check.yml
+++ b/.github/workflows/type-check.yml
@@ -40,4 +40,5 @@ jobs:
           hwilib/hwwclient.py
           hwilib/__init__.py
           hwilib/key.py
+          hwilib/policy.py
           hwilib/udevinstaller.py
--- a/.github/workflows/type-check.yml
+++ b/.github/workflows/type-check.yml
@@ -32,6 +32,7 @@ jobs:
           hwilib/devices/__init__.py
           hwilib/devices/keepkey.py
           hwilib/devices/ledger.py
+          hwilib/devices/ledger_registration.py
           hwilib/devices/trezor.py
           hwilib/errors.py
           hwilib/_script.py
--- a/hwilib/_cli.py
+++ b/hwilib/_cli.py
@@ -12,6 +12,7 @@ from .commands import (
     getdescriptors,
     prompt_pin,
     toggle_passphrase,
+    register,
     restore_device,
     send_pin,
     setup_device,
@@ -24,6 +25,7 @@ from .common import (
     AddressType,
     Chain,
 )
+from .policy import BIP388Policy
 from .errors import (
     handle_errors,
     DEVICE_CONN_ERROR,
@@ -58,6 +60,10 @@ def backup_device_handler(args: argparse.Namespace, client: HardwareWalletClient
 
 def displayaddress_handler(args: argparse.Namespace, client: HardwareWalletClient) -> Dict[str, str]:
     return displayaddress(client, desc=args.desc, path=args.path, addr_type=args.addr_type)
+
+def register_handler(args: argparse.Namespace, client: HardwareWalletClient) -> Dict[str, Optional[str]]:
+    policy = BIP388Policy.from_descriptor(name=args.name, descriptor=args.desc)
+    return register(client, bip388_policy=policy)
 
 def enumerate_handler(args: argparse.Namespace) -> List[Dict[str, Any]]:
     return enumerate(password=args.password, expert=args.expert, chain=args.chain, allow_emulators=args.allow_emulators)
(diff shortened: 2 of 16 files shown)


Sjors:
I'll give this a rebase after #836.

I need to sync this up with the latest #794 (e2a7288, c3815ba), and consider cherry-picking 562fe54.

achow101:
It seems like BitBox02 also has policy registration, so I would prefer if this could also be implemented for that as well. We should not be adding commands for one device only.

I strongly prefer for this to be generic and standardized. As of now, there is no standard for policy registration. Policies can be registered on BitBox02 but that seems to return nothing back to the user as they have enough storage to remember that registration. Conversely, Ledger returns the hmac. Since clearly not every device that can take policies will return the same thing, I think it would be better to say that the result value is some opaque binary result that has no meaning to the user. Rather, the interpretation of that blob occurs on a per-device basis, which users, and HWI for the most part, don't need to care about what that value is. We could also invent some encoding for it that can wrap the ledger's hmac, or signal that there is no value for the bitbox.

I also don't really like strictly following BIP 388 here in taking all of the fields directly from BIP 388 as separate positional arguments. Fundamentally, the policy is just a bastardized descriptor, and it should be straightforward to go from a descriptor to a policy. We have to massage the data into the device specific protocol stuff anyways. So instead of taking a policy, this could take a full descriptor, extract the keys and put in the substitutes, and verify that the rest of the descriptor meets the BIP 388 requirements. Then it can package up all of that to send to the device. This would reduce the number of parameters required and remove the requirement of the caller having to remember the correct key order.

Lastly, in addition to returning the policy registration result, I think the result should also include the descriptor/policy as part of the opaque string. Then the whole policy and proof can be supplied to `signtx` as a single argument rather than also requiring the user to know specifically which pieces to pass in. Both the policy and proof can be encoded into one big string that is visually meaningless to the user (e.g. base64 encode it) that can be passed in at a later time.

Sjors:
> We have to massage the data into the device specific protocol stuff anyways. So instead of taking a policy, this could take a full descriptor, extract the keys and put in the substitutes, and verify that the rest of the descriptor meets the BIP 388 requirements.

Having HWI convert from descriptors to BIP388 will make things easier on the Bitcoin Core side too, so that's nice.

In https://github.com/bitcoin-core/HWI/issues/785#issuecomment-5140521971 I suggested that we simply store a max 255 byte blob from the device. A null blob means we don't store anything. That should be easy to make device independent.

> I think the result should also include the descriptor/policy as part of the opaque string.

Oh but then it needs to be bigger. And the wallet still needs to know what blob to pass for which descriptor, so not sure if this helps at all.

cc @bigspider

Sjors:
> It seems like BitBox02 also has policy registration, so I would prefer if this could also be implemented for that as well. 

Last time I checked BitBox didn't support MuSig2, but that's good reason to add BIP388 support independent of MuSig2.

Sjors:
Pushed the new approach, it's fully vibe coded though, will polish it later if needed.

The first commit adds (multi-path) descriptor -> BIP388 conversion, with BIP test vectors.

These will be shared with #792 so I'll keep that draft.

The next commits add ledger and bitbox02, with only the former returning an hmac.

bigspider:
> In [#785 (comment)](https://github.com/bitcoin-core/HWI/issues/785#issuecomment-5140521971) I suggested that we simply store a max 255 byte blob from the device. A null blob means we don't store anything. That should be easy to make device independent.

This is in line with the approach I drafted in [my old PoC](https://github.com/bitcoin-core/HWI/pull/647), and seems likely to remain compatible with all devices.

Sjors:
Fixed the linter, updated PR description.

Sjors:
> > I think the result should also include the descriptor/policy as part of the opaque string.
> 
> Oh but then it needs to be bigger. And the wallet still needs to know what blob to pass for which descriptor, so not sure if this helps at all.

I changed `register` to echo the policy _name_ for Ledger. That way #792 doesn't need a `--policy-name` argument. The `signtx` command still takes a descriptor argument.

We could also shove the descriptor into the registration blob, but then we're stuck with it forever. I'd rather see this be added to PSBT spec, which seems useful in general. In the mean time a `--descriptor` argument is easy enough to support, and something we already do for other methods.

We could add a (proprietary) field for authentication, but the Bitcoin Core wallet needs to be aware of this mechanism one way or the other, and having a `--registration` argument seems simple enough.

---

I briefly investigated how this works with ColdCard. IIUC it's able to figure out which policy is being used by looking at the PSBT input derivations. But when connected via USB, you can help it by passing the policy name. So for registration we should return `00 <name-length> <ASCII name>` (empty hmac + policy name). We could even support offline registration by generating the necessary json file and putting it in a tmp folder.

achow101:
Come on, we already have an entire descriptor parser with classes. Why does this include a wholly separate descriptor parser, that uses regex insanity?

This is too sloppy, I've opened #842 to do this the right way.

Sjors:
I said:

> it's fully vibe coded though, will polish it later if needed.

Of course much better if you do it :-) I'll review #842.

PR #792, opened 2025-07-21
Optionally pass BIP388 policy to signtx

Sjors:
Another step toward #785.

This adds optional registered BIP388 descriptor policy information to
`signtx`. Existing `signtx` usage remains unchanged when no registration
is supplied.

`signtx` accepts `--registration`, containing the serialized
`RegisteredDescriptor` returned by the `registerdescriptor` command from
#842. The registration contains the policy name, descriptor, device type,
and any device-specific registration data, so separate `--policy-desc`
and `--policy-name` arguments are not needed.

Example:

```bash
hwi --device-type ledger signtx \
  --registration "cmRlc2M..." \
  "cHNidP..."

### CODE REVIEW

**Comment by Sjors** [comment]
I'll update this after #836, taking the latest changes in #794 into account.

**Comment by Sjors** [comment]
Rebased after #836.

**Comment by Sjors** [comment]
This now uses the same approach as #791, with the first two commits shared, so I'll keep this draft.

**Comment by Sjors** [comment]
Rebased and now using `RegisteredDescriptor` from https://github.com/bitcoin-core/HWI/pull/842. I still need to polish the commits. This also has some overlap with #841 so I suggest reviewing that first.

**Comment by Sjors** [comment]
ColdCard's multisig simulator patch, needed a patch 12ccec2f3022e51e1014be618cdb7e9864c8220b. This belongs in #842 though.

**Comment by Sjors** [comment]
Tidied up the commits a bit. I noticed that our BitBox02 code only allows one occurrence per policy. That's a problem with more complicated recovery paths, but that's perhaps better left for a followup to improve.

**Comment by Sjors** [comment]
Rebased after #842. Ready for review, though I suggest prioritizing #841.

**Comment by Sjors** [comment]
The two early ledger legacy failures look spurious.

File: hwilib/devices/ledger.py [hunk 1]
```diff
@@ -267,6 +281,16 @@ def legacy_sign_tx() -> PSBT:
...
                         continue
 
+            if registered_wallet is not None:
+                assert registered_descriptor is not None
+                wallets[registered_wallet.id] = (
+                    signing_priority[script_addrtype],
+                    script_addrtype,
+                    registered_wallet,
+                    registered_descriptor.registration,
+                )
```
achow101: In 07833327384f92e3fad68a5f1f187f7e25ee4bb9 "ledger: support BIP388 policy signing"

This should live outside of the loop, there's only one possible wallet policy, it does not need to have the same data set for every input in this loop. Furthermore, if the transaction were to involve both a provided policy and an inferred policy, the inferred would not be signed.

hwilib/devices/ledger.py
```diff
@@ -188,7 +188,11 @@ def get_pubkey_at_path(self, path: str) -> ExtendedKey:
         return ExtendedKey.deserialize(xpub_str)
 
     @ledger_exception
-    def sign_tx(self, psbt: PSBT) -> PSBT:
+    def sign_tx(
+        self,
+        psbt: PSBT,
+        registered_descriptor: Optional[RegisteredDescriptor] = None,
+    ) -> PSBT:
```
achow101, line 196: In 07833327384f92e3fad68a5f1f187f7e25ee4bb9 "ledger: support BIP388 policy signing"

This function signature is being changed without updating any callsites or the base class. That is dangerous.

hwilib/devices/ledger.py
```diff
@@ -188,7 +188,11 @@ def get_pubkey_at_path(self, path: str) -> ExtendedKey:
         return ExtendedKey.deserialize(xpub_str)
 
     @ledger_exception
-    def sign_tx(self, psbt: PSBT) -> PSBT:
+    def sign_tx(
+        self,
+        psbt: PSBT,
+        registered_descriptor: Optional[RegisteredDescriptor] = None,
```
achow101: In 07833327384f92e3fad68a5f1f187f7e25ee4bb9 "ledger: support BIP388 policy signing"

I would prefer for this to take a set of descriptors, rather than only one, so that if a transaction involves multiple policies, they can all be signed in one call here.

test/test_device.py
```diff
@@ -810,3 +814,28 @@ def test_register_descriptor(self):
...
+        }])
+        self.assertTrue(import_result[0]["success"])
+
+        self.wpk_rpc.sendtoaddress(self.wrpc.getnewaddress(), 1)
+        self.wpk_rpc.generatetoaddress(6, self.wpk_rpc.getnewaddress())
+        psbt = self.wrpc.walletcreatefundedpsbt(
+            [],
+            [{self.wpk_rpc.getnewaddress(): 0.5}],
+            0,
+            {"includeWatching": True},
```
achow101: In 061b6d61f09b0a41c6be43bd58bf9092b0015cd4 "test: sign with registered descriptor"

`includeWatching` is no longer necessary.

achow101 (commented):
Needs rebase

570dcb0460ff6c72b2b156f07693922e7847c8e5 "coldcard: iterate simulator multisig paths correctly", 4cd34d20bed11da96103d963fcba9b85127ff7be "bitbox02: factor out BIP388 script config" can be dropped.

Sjors:
Rebased after https://github.com/bitcoin-core/HWI/pull/849 and added edge support, which uses a new helpers `PartiallySignedInput.has_signature(fingerprint)` and `has_fingerprint`.
 
Also addressed feedback.

hwilib/devices/ledger.py
Sjors: Done, with a test. Not all devices support this, but I cherry-picked 23fb64e7f28e45aa0abb14cd56ba5f8db4c61881 _coldcard: enroll Edge policies with Miniscript_ from #841 so ColdCard edge can handle multiple policies (one approval each).

hwilib/devices/ledger.py
Sjors: I wanted to avoid the churn of having `raise UnavailableActionError("... does not support BIP388 policy signing")` followed by implementing it. But I've now reordered the commits and done that.

hwilib/devices/ledger.py
Sjors: This should be better now.

One thing I didn't fix, but could be a good followup, is that if Bitcoin Core adds global xpubs, some multisig policies can be inferred and users have to approve twice.

The tests are now run once with and once without global xpubs.

Sjors:
> [570dcb0](https://github.com/bitcoin-core/HWI/commit/570dcb0460ff6c72b2b156f07693922e7847c8e5) "coldcard: iterate simulator multisig paths correctly" [...] can be dropped

No it's still needed (or maybe it wasn't, but it definitely is now). Without it all non-edge coldcard jobs fail, e.g. https://github.com/bitcoin-core/HWI/actions/runs/32497281637/job/96820811201?pr=792

hwilib/devices/ledger.py
```diff
@@ -234,6 +234,38 @@ def legacy_sign_tx() -> PSBT:
...
+            registered_descriptors or set(),
+            key=lambda registration: registration.serialize(),
+        ):
+            descriptor = registered_descriptor.descriptor
+            registered_wallet = WalletPolicy(
+                registered_descriptor.name,
+                descriptor.get_bip388_template(),
+                [p.get_bip388_key_info() for p in descriptor.get_pubkey_providers()],
+            )
+            expanded = descriptor.expand(0)
```
achow101: In b947aefe8bab1f7dbe3587f849dbea4380178eff "ledger: support BIP388 policy signing"

I don't think we should use `expand` here as it doesn't play nicely with multipath descriptors. Since that function isn't actually used anywhere else, I'm thinking to drop it. That would also make the musig descriptor implementation easier as we would not have to implement key aggregation.

I think the better way is to have a function on `Descriptor` to return an `Optional[AddressType]` since all of descriptors know their address type.

Sjors: > Since that function isn't actually used anywhere else, I'm thinking to drop it.

Done in `descriptor: remove unused script expansion`.

> function on `Descriptor` to return an `Optional[AddressType]` since all of descriptors know their address type.

Done in `descriptor: add address type helper`.

Sjors:
Rebased after #841. Dropped the multisig patch be4d753e36fa6c97bf46c17ebb3b9442a2ecaf1b because it was included in #841 via  86a86c6f26b1fcb8be44b151cead3e3ba661a40f.

Sjors:
Pushed an automation fix to dismiss `Transaction signed` and `Account registered`. Noticed while working on #852. They belong here, even though they didn't cause problems.

Sjors:
Extracted two more Ledger automations from #794 which handle transactions with longer source / destination addresses (i.e. taproot), not anything MuSig2 specific.

achow101:
ACK 97c13f1e99384816518a48095e59a8a1f9468d7a

PR #795, opened 2025-07-22
Fix CI

Sjors:
- Modernize
  - bump build.Dockerfile to Debian bookworm
  - bump https://github.com/docker/build-push-action from v5 to v6 (probably wasn't the issue)
  - drop Python 3.8: https://devguide.python.org/versions/ (some jobs were failing, though I didn't investigate)
- glibc compatility
  - downgrade build machines for Ledger, Keepkey and Coldcard simulators to fix their build
  - downgrade build machines for Trezor simulator for glibc compatibility

- Robustness / CI performance
  - revert #743
  - split device groups (see picture below)
  - abort test if any simulator fails to start
- bitcoind
  - `src/bitcoind` was moved to `bin/bitcoind` by https://github.com/bitcoin/bitcoin/pull/31161
  - build `bitcoind` with Ubuntu 22.04 for glibc compatibility
  - abort test if bitoind fails to start
- Coldcard
  - update patches
  - add `pysdl2-dll`
  - use `--headless`
- Jade
  - include #779
- Ledger
  - use LedgerX app (except for legacy)
  - update PyQt to 6 for Speculos: https://github.com/LedgerHQ/speculos/pull/565
  - add `flask-cors`
  - add `ledgerd`
- Trezor
  - add missing rust-src dependency

The simulator build jobs as well as the simulator test jobs are split in one group per device brand. This means that the failure of one simulator to build will only cancels jobs for that simulator and not the other ones. Previously this was done in #743 using `always()`, but that's a bit overkill. 

Scherm­afbeelding 2025-08-29 om 19 09 56

Issues left for a followup:

- #799
- #800


Sjors:
It seems that CI doesn't re-run if I modify a commit and force push.

Sjors:
Closing this and continuing in #796 because https://github.com/bitcoin-core/HWI/actions/runs/16440824447 has been stuck for hours.

Sjors:
CI for this PR will need to be manually stopped, though hopefully thanks to b9abebef35677f72463d18c51e159c3caf7ef34a and bfa0eb38584a0ea022a16d2b6c6619119f2ecf4e it will stop itself quickly enough once started.

Sjors:
I was able to "fix" both the Ledger Speculos and Keepkey builders by downgrading to Ubuntu 22.04. Since I don't want to needlessly downgrade the other builders, it now makes sense to pull #797 in here.

Sjors:
I'm not 100% sure if the Coldcard multisig patch update in e782954c2377953e2a92d465f90bdf02eb57fee8 is necessary, because the previous CI failure may have been due to me messing up whitespace. Still it seems a good idea to refresh it given how much the underlying file changed in recent years.

Sjors:
Ledger was missing `flask-cors`.

Sjors:
The NanoX app seems to be stuck in `test_display_address_descriptor`. E.g. https://github.com/bitcoin-core/HWI/actions/runs/16473768261/job/46570910403#step:4:5335 

It passes on the legacy (NanoS) jobs, e.g.: https://github.com/bitcoin-core/HWI/actions/runs/16473768261/job/46570910455

@bigspider any idea what could cause that?

The next push adds a timeout to proc.communicate so at least it doesn't hang forever.

Sjors:
Maybe it's this: https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/src/ui/display_nbgl.c#L353C8-L357 "Wallet policy" or "Descriptor template" depending on screen size.

Sjors:
ColdCard tests fail with "xterm: Xt error: Can't open display: %s" and "xterm: DISPLAY is not set", e.g. https://github.com/bitcoin-core/HWI/actions/runs/16476780539/job/46581762316?pr=795

Trying again by installing `xvfb`, though I wonder why this worked in the past.

Ledger is still unhappy, I might try running it locally to see what text is on screen.

Sjors:
GREEN!

contrib/build.Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN apt-get install -y \
     libudev-dev \
     faketime \
     zip \
-    pyqt5-dev \
+    python3-pyqt6 \
```
achow101, line 30: In 219aa6f4ee94dee5721cd6d27bf35ed0299fa311 "Bump pyqt5 to pyqt6"

Note that this isn't used for speculos, but rather for our own gui build.

Sjors: Ah, I'll revert that. We can upgrade the gui later.

Sjors:
Splitting the groups is very useful in debugging, from my experience so far. We can probably reduce the boiler plate.

Sjors:
Noticed while testing #804 that python-latest now pulls in Python 13 which we don't support yet.

Sjors:
@benma your fix didn't do the trick, but let's just continue debugging it here...

Sjors:
I think you just missed one instance, trying again...

benma:
> I think you just missed one instance, trying again...

TIL `ripgrep` does not search the `.github` folder :see_no_evil:

Sjors:
Not sure what's wrong with 4e0cf0ff5dc0355b15c9354b41cba05da5160754: `fatal: ambiguous argument 'HEAD~3': unknown revision or path not in the working tree` https://github.com/bitcoin-core/HWI/actions/runs/17331581629/job/49209456367#step:4:700

I'll just drop it for now and open a fresh PR later. Everything else is green. `ci.yaml` is a lot shorter, albeit at the expense of some complexity.

hwilib/devices/trezorlib/btc.py
```diff
@@ -19,7 +19,6 @@
 from decimal import Decimal
 from typing import TYPE_CHECKING, Any, AnyStr, Dict, List, Optional, Sequence, Tuple
 
-# TypedDict is not available in typing for python < 3.8
```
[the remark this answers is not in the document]
Sjors: Done. Maybe we should be using `git subtree` for these and then have a linter keep an eye on it? IIUC you can actually export a subdirectory of a repo and then import it as a subtree.

benma:
> I also need to add `libcmocka0`, any other new deps?

@Sjors bb02 master is now fixed and its simulator does not depend on libcmocka0 anymore.

achow101:
ACK 70e14aa712b04be6f1aaae14d9a6ef330a96baa0

Seems fine I guess

PR #797, opened 2025-07-22
Split CI by device group

Sjors:
Ensures the failure to build a simulator for one device doesn't abort running jobs for the others. They're still grouped by manufacturer.

It uses composite actions so it's hopefully too verbose.

Alternative approach to #743.

Builds on #795.

Scherm­afbeelding 2025-07-22 om 20 12 47


Sjors:
Folding this into https://github.com/bitcoin-core/HWI/pull/795.

Issue #817, opened 2026-01-17
JadeInterface not compatible with cbor2  5.8.0


andreasgriffin:
### Bug

`hwilib/devices/jadepy/jade.py`  `read(self, n)` is called with `n=4096`  in cbor2  5.8.0 .  
For cbor2 5.7.1 it was called with   `n=1`. This leads to a complete freezing and makes Jade unusable.

### Mitigation

-> Stay at  cbor2 5.7.1

PR #832, opened 2026-04-30
deps: relax cbor2 upper bound to permit >=5.9.0

k9ert:
## Summary

Relax the `cbor2` constraint from `>=5.4.6,<5.8` to `>=5.4.6,!=5.8.0,<6.0.0`, allowing downstreams to install HWI together with `cbor2>=5.9.0`.

## Motivation

The current `<5.8` cap was introduced in 38fae57 ("Constrain cbor2 to <5.8") in response to #817 — `cbor2==5.8.0` changes the decoder's stream-read pattern from `read(n=1)` to `read(n=4096)`, which causes Jade signing to hang on pyserial-style transports.

That fix made sense at the time, but it traps downstream users at a vulnerable version of cbor2:

- **GHSA-3c37-wwvx-h642** (high severity): DoS via uncontrolled recursion in `cbor2.loads`. Affects all `cbor2 < 5.9.0`.
- **GHSA-wcj4-jw5j-44wh** (medium): CBORDecoder reuse can leak shareable values across decode calls. Fixed in `cbor2 >= 5.8.0`.

The high-severity advisory is unfixable under the current cap. Downstream projects that pin `cbor2>=5.9.0` for security cannot install HWI 3.2.0 without overriding the resolver.

## Evidence

I built a hardware-in-the-loop matrix that exercises the full Jade signing path (`enumerate` → `getmasterxpub` → `signtx`) under HWI 3.2.0 against cbor2 versions 5.7.1, 5.8.0, and 5.9.0, with and without a candidate read-loop patch.

Test repo (with raw logs and reproduction instructions): https://github.com/al-munazzim/hwi-jade-cbor-repro

| mode \ cbor2 | 5.7.1 | 5.8.0 | 5.9.0 |
|---|---|---|---|
| stock HWI 3.2.0 | PASS | FROZEN | **PASS** |
| with read-loop patch | PASS | PASS | PASS |

Tested on macOS 15.4.1, Python 3.9.23, HWI 3.2.0, against an unlocked Jade signer. The key cell: stock HWI 3.2.0 + cbor2 5.9.0 completes Jade signing successfully without any code change in HWI. Whatever changed in cbor2 5.8.0 to break the read pattern appears resolved in 5.9.0 itself.

## Scope

This PR is intentionally minimal: cap relax only.

The `JadeInterface.read()` loop is still fragile — a future cbor2 change could in principle re-trigger the freeze. Hardening that path is a separate concern and a candidate follow-up PR (the patch from the test repo above can be the basis).

`cbor2 == 5.8.0` is excluded explicitly (`!= 5.8.0`) since the FROZEN result there is reproducible and confirmed on real hardware.

## Changes

- `pyproject.toml`: `cbor2 = ">=5.4.6,<5.8"` → `cbor2 = ">=5.4.6,!=5.8.0,<6.0.0"`
- `setup.py`: same constraint (kept in sync with pyproject)
- `poetry.lock` not regenerated in this PR. Happy to add a follow-up commit regenerating it if that is the convention; otherwise it can be regenerated at next release.

## Refs

- Reverts the workaround scope from #817 for `cbor2 >= 5.9.0`
- Original cap commit: 38fae57
- Security advisory: https://github.com/advisories/GHSA-3c37-wwvx-h642


k9ert:
The pin was IMHO done because @Sjors discovered a timeout of jade-tests and pinned in 38fae57 .
But those are now fixed for some reason, right?!

Sjors:
> But those are now fixed for some reason, right?

From that commit:

> In the Sdist and Wheel test environments, the Jade tests all timed out.

So this would be hard to miss.

k9ert:
1. I had forgotten to update poetry.lock after relaxing the cbor2 bound in pyproject.toml.
That mismatch caused early CI failures in lint / Non-device tests / Dist builder / Wine builder.
After adding the lockfile update, those jobs are green in the latest PR run:

• Dist builder ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73922819687
• Non-device tests ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73922819680
• lint ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73922819673
• Wine builder ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73922819703

Also, regarding the earlier timeout context: Jade jobs are not timing out in this run (they are green), e.g.

• Jade library ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73923444972
• Jade bindist ✅ https://github.com/bitcoin-core/HWI/actions/runs/25211462096/job/73923444980

2. The remaining red jobs do not appear to be caused by this PR change, since the same failures are present on master:

• Trezor sim builder fails on master with error[E0658]: custom test frameworks are an unstable feature
https://github.com/bitcoin-core/HWI/actions/runs/25012560729/job/73251883660
• Ledger/ledger-legacy jobs fail on master with speculos safe-directory error (fatal: detected dubious ownership ... /test/work/speculos)
https://github.com/bitcoin-core/HWI/actions/runs/25012560729/job/73253524077

So the lockfile issue is fixed in this PR, the remaining failures look like pre-existing CI/environment issues.

k9ert:
Any feedback, @Sjors ?!

trevarj:
ACK from me. 
Running into this just now with the Jade emulator. Would be good to have it patched so I don't need to monkey patch it

Sjors:
utACK 400f89c0dddcd5fed9ea4b3b075ea7335bf45702

The failing ledger tests seem unrelated. Same for the Trezor sim builder failure.

k9ert:
So we can expect a merge soon? Or what is still missing?

Sjors:
That's up to @achow101. The safest sequence would be to merge #836 first, rebase this to double-check that no test breaks, and then merge.

achow101:
please rebase

achow101:
superseded by #843

PR #819, opened 2026-01-21
Drop Python 3.9, use 3.10 for dist, bump Ledger

Sjors:
This PR first switches the deterministic builds to Python 3.10. It then drops the end-of-life 3.9.

Then it bumps Speculos and Ledger and enables the tests that were previously disabled. 

There's also a few minor cleanup commits, plus changes to make the Docker containers Podman friendly so I can easily run them locally.

PR #836, opened 2026-07-03
Fix CI ... again

Sjors:
Commits are taken from #819 (which is now based on this).

- Trezor builds started to fail when Rust nightly introduced some new rule, so we pin the nightly
- https://github.com/bitcoin/bitcoin/pull/21283 broke the Jade tests (and probably more than that)

Two minor commits drop unused CI code.

Ledger moved repos around, so also include:

- https://github.com/bitcoin-core/HWI/pull/837


bigspider:
Might want to bring in the commit from #837 (I think the CI would now fail otherwise)

Sjors:
Done!

Sjors:
The keepkey failures seem spurious.

hwilib/psbt.py
```diff
@@ -1046,8 +1046,11 @@ def cache_unsigned_tx_pieces(self) -> None:
         Does nothing if the PSBT is already v2.
         """
         # To make things easier, we split up the global transaction
-        # and use the PSBTv2 fields for PSBTv0
-        if self.tx is not None:
+        # and use the PSBTv2 fields for PSBTv0.
+        # Don't check self.tx instead of self.version: for PSBTv2 it is
+        # merely an empty placeholder, and its default version and locktime
+        # would overwrite the deserialized values.
```
achow101: In 4d974551d11878152ae4f9f239963b4b555761aa "psbt: don't overwrite PSBTv2 tx version and fallback locktime"

This much commentary is not necessary

hwilib/psbt.py
```diff
@@ -1046,8 +1046,11 @@ def cache_unsigned_tx_pieces(self) -> None:
...
+        if self.version == 0:
```
achow101, line 1050: In 4d974551d11878152ae4f9f239963b4b555761aa "psbt: don't overwrite PSBTv2 tx version and fallback locktime"

The version 2 checks are kind of split between a version check and `if self.tx.is_null()`. Could unify them under one pattern in this commit.

hwilib/psbt.py
Sjors: Deslopped.

hwilib/psbt.py
Sjors: Done.

Claude found a bunch of other issues with the v2 implementation, so I ended up adding all test vectors, and fixes, in an incremental manner. See #839.

This PR only adds the vectors for this bug fix.

Sjors:
The one ledger legacy CI failure seems spurious (and passed in #839).

Sjors:
Actually that was another case of `To`:

```
seproxyhal: received (tag: 0x5e, size: 0x2c): b'\x07\x12\x07\x00+\x00r\x00\x0c\x00\x00\x00\x00\x00\x00\x00\xff\xff\xff\x00\x00\x00\x00\x00\n\x80\x00\x00ToFjp7GLudtJ5VxU'
```

https://github.com/bitcoin-core/HWI/actions/runs/29081182723/job/86325723028?pr=836

So I'm pulling https://github.com/bitcoin-core/HWI/pull/819/changes/be87c6ebdf55ed24561a8c3b633cc488dfd08f49 in from #819.

achow101:
ACK d3e4ce7c9a4ebac1f182aae31a60a3b546b4c7fe

PR #839, opened 2026-07-10
psbt: misc fixes and add remaining BIP 174 and BIP 370 test vectors 

Sjors:
Each fix / hardening commit introduces the test vectors that cover it.

Mostly straight-forward, but `psbt: parse the global unsigned tx without witness data` is worth a closer look. IIUC it catches up with https://github.com/bitcoin/bips/pull/1099.


Sjors:
Ready now that #836 landed.

test/test_psbt.py
```diff
@@ -40,5 +40,14 @@ def test_convert_to_v0(self):
         self.assertEqual(psbt.tx.vin[0].nSequence, 0xfffffffe)
         self.assertEqual(len(psbt.tx.vout), 2)
 
+    def test_convert_to_v0_locktime(self):
```
achow101: In 629c9b83b7e404d53f2ac21cd36507d0f8fe67ad "psbt: set locktime on the transaction, not the PSBT"

Can you combine this with the other test case instead of making a new one? It's basically the same transaction.

Sjors: If dropped `test_convert_to_v0_default_sequence` and `test_convert_to_v0_locktime`, using both fields in the generic `test_convert_to_v0`.

Sjors:
Improved the test, rebased just in case.

achow101:
ACK df8c3f5d4b7e6c6b4856af8bde50c6fc9cf7431c

PR #841, opened 2026-08-02
Optionally pass BIP388 policy to displayaddress

Sjors:
Support addresses display for registered BIP388 descriptor policies.

`displayaddress` accepts:
- the `RegisteredDescriptor` registration returned by `registerdescriptor` (introduced by #842)
- an address index 
- receive/change selection (multipath index)

A generic device test registers the simple multisig policy introduced by
#842 and then displays address index 7. Device support is added one
device per commit.


Sjors:
Rebased to follow the approach in #842.
- _registerdescriptor: encode policy name in registration_: appends the policy name to registration. This is still pending discussion https://github.com/bitcoin-core/HWI/pull/842#discussion_r3710554299. The commit can hopefully be dropped here (so I didn't polish it).
- `test: display registered descriptor address` expands the existing test
- the other commits follow the same pattern as 842 (still working on cleanups)

Sjors:
This needs a rebase after #819, but I'll wait for #842 to have green CI.

Sjors:
Rebased and now using `RegisteredDescriptor` from #842. The `--desc` argument is now mutually exclusive with `--registration`, which simplifies things slightly.

Note that ColdCard edge firmware has more broad policy support, but this PR doesn't but this PR doesn't switch to that. This might be worth doing once we add MuSig2.

Sjors:
Fixed yet another automation failure, not related to this PR, this time an address starting with `Fee3bWhy6jMv2rzbRs`.

Sjors:
Rebased after #842.

hwilib/descriptor.py
```diff
@@ -354,6 +355,17 @@ def get_pubkey_providers(self) -> list['PubkeyProvider']:
             out.extend(s.get_pubkey_providers())
         return out
 
+    def derive(self, pos: int, change: bool = False) -> 'Descriptor':
```
achow101: In 98587998792d815f32310af10539a2c2456ddb1e "descriptor: add ranged descriptor derivation"

Instead of a bool, this should give the multipath index directly so it works generically for multipaths of more than 2 indexes.

hwilib/commands.py
```diff
@@ -435,8 +436,11 @@ def displayaddress(
     client: HardwareWalletClient,
     path: Optional[str] = None,
     desc: Optional[str] = None,
-    addr_type: AddressType = AddressType.WIT
-) -> Dict[str, str]:
+    addr_type: AddressType = AddressType.WIT,
+    registration: Optional[str] = None,
+    index: Optional[int] = None,
+    change: bool = False,
```
achow101: In 75f1f71f37b1749c02d16945ccdcb726ecdc5991 "displayaddress: add BIP388 policy support"

Same comment regarding `change`.

hwilib/_cli.py
```diff
@@ -198,7 +206,11 @@ def get_parser() -> HWIArgumentParser:
     group = displayaddr_parser.add_mutually_exclusive_group(required=True)
     group.add_argument('--desc', help='Output Descriptor. E.g. wpkh([00000000/84h/0h/0h]xpub.../0/0), where 00000000 must match --fingerprint and xpub can be obtained with getxpub. See doc/descriptors.md in Bitcoin Core')
     group.add_argument('--path', help='The BIP 32 derivation path of the key embedded in the address, default follows BIP43 convention, e.g. ``m/84h/0h/0h/1/*``')
+    group.add_argument('--registration', help='Registration returned by the registerdescriptor command')
     displayaddr_parser.add_argument("--addr-type", help="The address type to display", type=AddressType.argparse, choices=list(AddressType), default=AddressType.WIT) # type: ignore
+    displayaddr_policy_group = displayaddr_parser.add_argument_group("BIP388 policy")
+    displayaddr_policy_group.add_argument('--index', help='Address index to display from the policy', type=int)
+    displayaddr_policy_group.add_argument('--change', help='Display the change address for the given index', action='store_true')
```
achow101: In 75f1f71f37b1749c02d16945ccdcb726ecdc5991 "displayaddress: add BIP388 policy support"

same comment regarding `change`. This could still have `--change` for a convenience, but should also allow other indexes.

hwilib/devices/bitbox02.py
```diff
@@ -442,6 +442,24 @@ def _maybe_register_script_config(
                 xpub_type=bitbox02.btc.BTCRegisterScriptConfigRequest.AUTO_XPUB_TPUB,
             )
 
+    def _bip388_script_config(
```
achow101: In 24572febd346cc67623714952867634915efb8a6 "bitbox02: factor out BIP388 script config"

This commit seems to be purely a move, and AFAICT, it's unnecessary. The function is already a member of `Bitbox02Client`.

Sjors: Dropped. Probably a rebase leftover.

hwilib/_cli.py
Sjors: Done (along with the above).

test/test_device.py
```diff
@@ -810,3 +810,15 @@ def test_register_descriptor(self):
...
+
+        result = self.do_command(self.dev_args + [
+            "displayaddress",
+            "--index", "7",
+            "--multipath-index", "1",
+            "--registration", result["registration"],
+        ])
+        self.assertNotIn("error", result)
+        self.assertNotIn("code", result)
+        self.assertIn("address", result)
```
achow101, line 824: In 966512c932ed4f6dcee4a102c148c2c33c74de7e "test: display registered descriptor address"

This could derive the address in Bitcoin Core and check that they match.

achow101 (commented):
Seems like 966512c932ed4f6dcee4a102c148c2c33c74de7e "test: display registered descriptor address" would fail Ci on its own, so should be placed after all of the device implementations.

Sjors:
Rebased after #849 and added Edge support. Also addressed feedback (commit order and checking against core derivation).

test/test_device.py
Sjors: Done

Sjors:
Fix another automation issue. Some LLM explanation:

> Speculos accepts any integer as the button value, then encodes it as a bitmask:
>
> - `1` = left
> - `2` = right
> - `3` = `1 | 2` = both
>
> Internally it sends `button << 1`, so `3` becomes `0x06`, the Ledger “both buttons pressed” event. The release always sends zero.

History of this part of the automation:

> The lineage is:
> 
> - e5808ac2766fbd3119a0cf0b1cb6871a4cafa3ba — `tests, ledger: Have speculos do button presses internally` (2020)  
  Originally introduced the four-call sequence in the old Speculos patch for approval and message signing.
>
> - 7c24028776cf35a1d1f26f976d29cf7817237688 — `Use built in Speculos automation instead of patch` (2021)  
  Migrated that sequence into `speculos-automation.json`. This is the direct ancestor of the generic approval and final message rules.
>
> - 9ba9a17315b0dab56cce78f23f57d6444d63920d — `test, ledger: Update automation` (2025)  
  Added four-event rules for the coordinate-specific `Confirm` and `Confirm account` screens. It also extended the generic approval rule to match `Sign transaction`, making it the most relevant commit for the observed failure.
>
> - 976353eded87c8c37a56f67be3b2e3d65368b105 — `Bump Speculos and Ledger Bitcoin app` (2026)  
  Added the second coordinate-specific `Confirm` rule with the same sequence.
>
> So the pattern originated in `e5808ac`, was converted to JSON by `7c240287`, and was copied or extended by `9ba9a173` and `976353ed`.

(I checked because I want to make sure we're actually improving the automation, and not going in LLM-fix circles)

Sjors:
I split `coldcard: support registered policy address display` (23fb64e + 169fdd0) in order to share `coldcard: enroll Edge policies with Miniscript` with #792, because on Edge signing can take advantage of the registration(s).

achow101:
ACK 458a92009f5f7244863f0e3def070f9212d8b479

PR #842, opened 2026-08-04
Add `registerdescriptors` command for registering a descriptor with a device

achow101:
The Ledger, BitBox02, Jade, and Coldcard all support registering descriptors with the device to enable signing txs involving complex scripts. This PR adds a `registerdescriptor` command which takes a name and the descriptor to register. The command returns a `registration` which may contain data that must be provided to the device at a later time to remind it of the registered descriptor.

For the BiBox02, Jade, and Coldcard, this command returns an empty string for the registration as these devices store the registration and do not need a reminder.

For the Ledger, the `registration` is the HMAC that the device returns.

Some devices take a descriptor, others take a BIP 388 policy that is stuffed into their protocol. The `Descriptor` and `PubkeyProvider` classes are modified to allow produce BIP 388 compatible strings that can be provided as needed.

Each device does different validation of the descriptor it is provided. We will not do any validation - the user may provide a descriptor that a device refuses to register and any such registration errors will be propagated from the device.

The test case includes a descriptor that should work on all devices.


hwilib/devices/ledger.py
```diff
@@ -550,6 +551,17 @@ def can_sign_taproot(self) -> bool:
...
+    @ledger_exception
+    def register_descriptor(self, name: str, descriptor: 'Descriptor') -> str:
+        if isinstance(self.client, LegacyClient):
+            raise UnavailableActionError("Legacy Ledger app does not support descriptor registration")
+
+        template = descriptor.get_bip388_template()
+        keys = [p.to_string_no_deriv(hardened_char="'") for p in descriptor.get_pubkey_providers()]
+        policy = WalletPolicy(name, template, keys)
+        _, registered_hmac = self.client.register_wallet(policy)
+        return registered_hmac.hex()
```
Sjors: In 8db5557aa220ddb3c18e74c119d5949a4f301742 _ledger: Implement register_descriptor_: in #791 I added the policy name to the registration result, so you don't have to provide it again with signing and address display. 

A simple `[hmac length]hmac bytes[name length][name bytes]` serialization means Ledger can change the length later, devices that only need a name can easily omit it, and we add additional fields later.

test/test_descriptor.py
```diff
@@ -25,6 +25,8 @@ def test_parse_descriptor_with_origin(self):
         self.assertEqual(desc.pubkeys[0].deriv_path, "/0/0")
         self.assertEqual(desc.pubkeys[0].expr_index, 0)
         self.assertEqual(desc.to_string_no_checksum(), d)
+        self.assertEqual(desc.get_bip388_template(), "wpkh(@0/0/0)")
```
Sjors: In d4600926541a98bdee29462723006eff02e953b8 _descriptor: this is missing several BIP388 test vectors (and would fail those, at first glance)

hwilib/devices/bitbox02.py
```diff
@@ -962,3 +965,20 @@ def can_sign_taproot(self) -> bool:
...
+                keypath=pk.origin.path if pk.origin else None,
+                xpub=util.parse_xpub(pk.pubkey)
+            ))
+        policy = bitbox02.btc.BTCScriptConfig.Policy(
+            policy=descriptor.get_bip388_template(),
+            keys=desc_keys,
+        )
+        script_config = bitbox02.btc.BTCScriptConfig(policy=policy)
+        self._maybe_register_script_config(script_config, [], name)
+        return ""
```
Sjors: In 3f5da8c18f1cf9df63e2bc935679639914e69095 _bitbox02: Implement register_descriptor_: why not `None`? (or echo the name, see my other suggestion)

Sjors (commented):
Concept ACK

test/test_descriptor.py
achow101: Why would it fail those?

This is not a BIP 388 parser and is intentionally not validating descriptors for their BIP 388 correctness before turning things into a template. Fundamentally, I think that 388 specifies 2 things: 1) a format conversion that can be applied to any descriptor, 2) a set of rules for what wallets should accept given a descriptor. For the purposes of HWI, we only need to do 1, and we leave 2 for the devices.

Sjors: It would good to clarify in the PR description and in the code that you're only doing (1), and fully rely on the device to handle (2).

It seems better to do (2) as well, because:
1. it's easier to review a complete set of test vectors
2. devices are notoriously bad at giving useful error messages

achow101: What, exactly, do the BIP test vectors test that are not being covered by our tests? There are some for miniscript, but we have not implemented miniscript. Those test vectors don't cover cases where a descriptor cannot become a wallet policy. They have tests for wallet policies that should not be parsed, but we don't implement a wallet policy parser.

hwilib/devices/ledger.py
achow101: Still thinking on what we should actually be returning.

test/test_descriptor.py
Sjors: Miniscript, MuSig2 and `sortedmulti_a`. We could explicitly omit those and leave them for a followup.

> They have tests for wallet policies that should not be parsed

So e.g. `pkh(@0)` is not allowed.

> but we don't implement a wallet policy parser.

Maybe we should? For the reasons I mention above, and additionally as defense in depth so a device doesn't accidentally register an invalid policy, that it might then refuse to sign at the next firmware update.

hwilib/devices/ledger.py
Sjors: I just updated #841 to be based on this PR. For now it adds one commit to add the name. It's sloppy, but should go away depending on what you decide here. I do think it's nice to avoid `--policy-name`.

test/test_descriptor.py
achow101: We are not implementing a wallet policy parser.

I could change the converter to enforce BIP 388 constraints and have tests for those, but those are necessarily going to be tests that are not part of BIP 388's test vectors as they don't have tests for this direction.

> Miniscript, MuSig2 and sortedmulti_a. We could explicitly omit those and leave them for a followup.

They already are.

achow101: I've added some constraint enforcement, but left out others. These are explicitly documented.

Sjors:
This should fix the ledger tests:

```diff
diff --git a/test/data/speculos-automation.json b/test/data/speculos-automation.json
index 2b273e7..5618fef 100644
--- a/test/data/speculos-automation.json
+++ b/test/data/speculos-automation.json
@@ -40,12 +40,5 @@
         },
         {
-            "regexp": "^. of . Multisig$",
-            "actions": [
-                [ "button", 2, true ],
-                [ "button", 2, false ]
-            ]
-        },
-        {
-            "regexp": "^(Address|Review|Account name|Amount|External amounts|You spend|You receive|Fee|Confirm|The derivation|Derivation path|Reject if|The change path|Change path|Register wallet|Policy map|Key|Path|Public key|Our key|Their key|Unspendable key|Spend from|Spending policy|Primary spending path|Spending path|Transaction output|Wallet name|Wallet policy|Descriptor template|Verify [Bb]itcoin|Output|Warning).*",
+            "regexp": "^(Address|Review|Account name|Amount|External amounts|You spend|You receive|Fee|Confirm|The derivation|Derivation path|Reject if|The change path|Change path|Register wallet|Policy map|Key|Path|Public key|Our key|Their key|Unspendable key|Spend from|Spending policy|Primary spending path|Spending path|Transaction output|From account|Wallet name|Wallet policy|Descriptor template|Verify [Bb]itcoin|Output|Warning).*",
             "actions": [
                 [ "button", 2, true ],
```

This fix should prevent a double click, tested on both legacy and new app. Related: 5f0edff, 9ba9a17, #812, #818

achow101:
I've decided to create a new `RegisteredDescriptor` serialized string. It's the base64 of a compact size version number, length prefixed descriptor string, length prefixed device type string, and length prefixed registration data (hmac on the ledger, empty string for everything else).

hwilib/key.py
```diff
@@ -339,6 +332,7 @@ def parse_path(nstr: str) -> List[int]:
 
     :param nstr: path string
     :return: list of integers
+    :raises ValueError: If the path contains any invalid characters
```
Sjors: In e2bd86e88c736e51dc36b6ad7c1de2baae37e78e _descriptors: Actually parse the derivation path_: could give this a bit more teeth:

```diff
diff --git a/hwilib/descriptor.py b/hwilib/descriptor.py
index 0e85311..f590d67 100644
--- a/hwilib/descriptor.py
+++ b/hwilib/descriptor.py
@@ -168,7 +168,17 @@ class PubkeyProvider(object):
             pubkey = s[:slash_idx]
             path_str = s[slash_idx + 1:]
-            ranged = path_str.endswith("*")
-            if ranged:
+
+            if not path_str:
+                raise ValueError("Derivation path cannot be empty")
+
+            if path_str == "*":
+                ranged = True
+                path_str = ""
+            elif path_str.endswith("/*") and path_str.count("*") == 1:
+                ranged = True
                 path_str = path_str[:-2]
+            elif "*" in path_str:
+                raise ValueError("Wildcard must be the final derivation path element")
+
             if len(path_str) > 0:
                 deriv_path = parse_multipath(path_str)
diff --git a/hwilib/key.py b/hwilib/key.py
index 506fb1a..5ce2fb3 100644
--- a/hwilib/key.py
+++ b/hwilib/key.py
@@ -335,10 +335,24 @@ def _parse_path(nstr: str, allow_multipath: bool) -> List[List[int]]:

     def str_to_harden(x: str) -> int:
+        original = x
+        hardened = False
+
         if x.startswith("-"):
-            return H_(abs(int(x)))
-        elif x.endswith(("h", "'")):
-            return H_(int(x[:-1]))
-        else:
-            return int(x)
+            hardened = True
+            x = x[1:]
+        elif x.endswith(("h", "H", "'")):
+            hardened = True
+            x = x[:-1]
+
+        # Avoid accepting syntax supported by int() but not by BIP32 paths,
+        # such as whitespace, signs, underscores, and non-ASCII digits.
+        if not x or not x.isascii() or not x.isdecimal():
+            raise ValueError(f"Invalid BIP32 path index: {original}")
+
+        index = int(x)
+        if index >= HARDENED_FLAG:
+            raise ValueError(f"BIP32 path index out of range: {original}")
+
+        return H_(index) if hardened else index

     def parse_index(x: str, seen_multipath: bool) -> Tuple[List[int], bool]:
@@ -354,5 +368,8 @@ def _parse_path(nstr: str, allow_multipath: bool) -> List[List[int]]:
                 raise ValueError(f"Invalid multipath specification, less than 2 indexes specified: {x}")
             seen_multipath = True
-            return [str_to_harden(p) for p in mp], seen_multipath
+            indexes = [str_to_harden(p) for p in mp]
+            if len(set(indexes)) != len(indexes):
+                raise ValueError(f"Duplicate indexes in multipath specification: {x}")
+            return indexes, seen_multipath
         return [str_to_harden(x)], seen_multipath

@@ -373,5 +390,5 @@ def parse_path(nstr: str) -> List[int]:
     """
     Convert BIP32 path string to list of uint32 integers with hardened flags.
-    Several conventions are supported to set the hardened flag: -1, 1', 1h
+    Several conventions are supported to set the hardened flag: -1, 1', 1h, 1H

     e.g.: "0/1h/1" -> [0, 0x80000001, 1]
@@ -379,5 +396,5 @@ def parse_path(nstr: str) -> List[int]:
     :param nstr: path string
     :return: list of integers
-    :raises ValueError: If the path contains any invalid characters
+    :raises ValueError: If an index has invalid syntax or is outside the uint31 range
     """
     return [i[0] for i in _parse_path(nstr, False)]
diff --git a/test/test_bip32.py b/test/test_bip32.py
index 5a88d82..b5e3ce7 100755
--- a/test/test_bip32.py
+++ b/test/test_bip32.py
@@ -6,4 +6,5 @@
 from hwilib.key import (
     ExtendedKey,
+    parse_multipath,
     parse_path,
 )
@@ -123,4 +124,28 @@ class TestBIP32(unittest.TestCase):
                 self.assertEqual(xprv_der.to_string(), child_xpub)

+    def test_parse_path_validation(self):
+        self.assertEqual(
+            parse_path("m/0/2147483647/2147483647h/-1/1H"),
+            [0, 0x7fffffff, 0xffffffff, 0x80000001, 0x80000001],
+        )
+
+        for path in (
+            "m/",
+            "+1",
+            "1_0",
+            " 1",
+            "١",
+            "2147483648",
+            "2147483648h",
+            "-2147483648",
+        ):
+            with self.subTest(path=path):
+                with self.assertRaises(ValueError):
+                    parse_path(path)
+
+    def test_parse_multipath_validation(self):
+        with self.assertRaisesRegex(ValueError, "Duplicate indexes"):
+            parse_multipath("<0;0>")
+

 if __name__ == "__main__":
diff --git a/test/test_descriptor.py b/test/test_descriptor.py
index 828e246..ec339d8 100755
--- a/test/test_descriptor.py
+++ b/test/test_descriptor.py
@@ -4,4 +4,5 @@ from hwilib.descriptor import (
     parse_descriptor,
     MultisigDescriptor,
+    PubkeyProvider,
     SHDescriptor,
     TRDescriptor,
@@ -278,4 +279,10 @@ class TestDescriptor(unittest.TestCase):
         self.assertEqual(d, desc.to_string_no_checksum())

+    def test_invalid_derivation_path_wildcards(self):
+        for key_expr in ("xpub/", "xpub/a*", "xpub/0*", "xpub/*/0", "xpub/**"):
+            with self.subTest(key_expr=key_expr):
+                with self.assertRaises(ValueError):
+                    PubkeyProvider.parse(key_expr, 0)
+
     def test_invalid_multipath_descriptors(self):
         with self.assertRaisesRegex(ValueError, "Cannot have multiple multipath specifiers"):
@@ -289,9 +296,9 @@ class TestDescriptor(unittest.TestCase):
         with self.assertRaisesRegex(ValueError, "Invalid multipath specification, less than 2 indexes specified: <>"):
             parse_descriptor("wpkh(xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB/<>/*)")
-        with self.assertRaisesRegex(ValueError, "invalid literal for int()"):
+        with self.assertRaisesRegex(ValueError, "Invalid BIP32 path index"):
             parse_descriptor("wpkh(xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB/0>/*)")
         with self.assertRaisesRegex(ValueError, "Invalid multipath specification, missing trailing '>'"):
             parse_descriptor("wpkh(xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB/<0/*)")
-        with self.assertRaisesRegex(ValueError, "invalid literal for int()"):
+        with self.assertRaisesRegex(ValueError, "Invalid BIP32 path index"):
             parse_descriptor("wpkh(xpub661MyMwAqRbcFW31YEwpkMuc5THy2PSt5bDMsktWQcFF8syAmRUapSCGu8ED9W6oDMSgv6Zz8idoc4a6mr8BDzTJY47LJhkJ8UB7WEGuduB/<0;>/*)")
```

hwilib/devices/coldcard.py
```diff
@@ -415,6 +420,40 @@ def can_sign_taproot(self) -> bool:
...
+
+        # verify the send
+        expect = check.digest()
+        result = self.device.send_recv(CCProtocolPacker.sha256())
+        assert len(result) == 32
+        if result != expect:
+            raise DeviceFailureError(f"Wrong checksum, expected {expect.hex()}, got {result.hex()}")
+
+        # Register the descriptor
+        self.device.send_recv(CCProtocolPacker.multisig_enroll(size, expect), timeout=None)
```
Sjors, line 454: In d9f384e5c5124f40e931bdc5d398aa7b9e36e148 _coldcard: Implement register_descriptor_:

```py
if self.device.is_simulator:
           self.device.send_recv(CCProtocolPacker.sim_keypress(b'y'))
```

Sjors (approved):
ACK a7a69f8759a0f06cbc42b3d01344eb3bb34ac671

Lightly reviewed the code. I mainly tested by rebasing #841 and #792 on top of this. 

In 6a5a4961188dab2ea29e6e5a153e19fa71f729fe _bitbox02: Implement register_descriptor_: it would save me a refactor commit if you put `register_descriptor` up to the `btc.BTCScriptConfig` call in a helper method.

hwilib/devices/coldcard.py
Sjors: You'll need https://github.com/bitcoin-core/HWI/commit/12ccec2f3022e51e1014be618cdb7e9864c8220b too.

hwilib/key.py
achow101: I don't think these are necessary, and your suggestions are both too strict and deviate from the specs. Not all descriptors must have derivation paths. `H` is not an allowed hardened indicator.

Sjors:
While testing I noticed this PR emits `@0/<0;1>/*` rather than `@0/**`. The latter is a bit easier on the eyes when checking on a device, but the BIP doesn't (strongly) recommend one or the other. Annoyingly a Ledger hmac generated with one form won't work with the other form, so whatever we pick has to stay that way.

achow101:
> While testing I noticed this PR emits `@0/<0;1>/*` rather than `@0/**`.

This is intentional. I consider `/**` to be a bad syntax. Derivation paths should be explicitly given.

hwilib/key.py
Sjors: > `H` is not an allowed hardened indicator.

Oops.

And mixed notations isn't allowed either (at least in Bitcoin Core). So yeah, ignore.

The ledger hmac is based on the exact string, but it's probably better to rely on the client to be very strict about it, rather than enforce a canonical variant on the HWI side. And worst case it's easy to register the policy again.

achow101:
> In [6a5a496](https://github.com/bitcoin-core/HWI/commit/6a5a4961188dab2ea29e6e5a153e19fa71f729fe) _bitbox02: Implement register_descriptor_: it would save me a refactor commit if you put `register_descriptor` up to the `btc.BTCScriptConfig` call in a helper method.

Done

hwilib/devices/coldcard.py
achow101: Added the simulator keypress, but CI was still passing without it. We'll see if the commit is needed.

Sjors: I guess the device doesn't wait for user input before returning, but the keypress becomes needed when there's another action after the approval.

Sjors:
re-ACK 50f5781e70552cca094cba23d0d301f46d364306

PR #846, opened 2026-08-07
errors: fix UNKNWON_DEVICE_TYPE misspelling, keep it as a compat alias

### LINKED ISSUE #845: hwilib.errors: UNKNWON_DEVICE_TYPE is misspelled
[`hwilib/errors.py`](https://github.com/bitcoin-core/HWI/blob/b209f369778aae0b441cb096e4afb7a2ea3717b4/hwilib/errors.py#L19) defines error code -4 as `UNKNWON_DEVICE_TYPE`:

```python
UNKNWON_DEVICE_TYPE = -4 #: Device type is unknown
```

"UNKNWON" is a transposition of "UNKNOWN". The same spelling is used at the one other site that references it, [`UnknownDeviceError.__init__`](https://github.com/bitcoin-core/HWI/blob/b209f369778aae0b441cb096e4afb7a2ea3717b4/hwilib/errors.py#L131):

```python
class UnknownDeviceError(HWWError):
    """
    :class:`HWWError` for :data:`DEVICE_TYPE`
    """
    def __init__(self, msg: str):
        """
        :param msg: The error message
        """
        HWWError.__init__(self, msg, UNKNWON_DEVICE_TYPE)
```

That docstring's `:data:`DEVICE_TYPE`` cross-reference is also wrong on top of the typo: no such name exists in this module, so the Sphinx `:data:` role can't resolve it either way.

The numeric value (-4) and its position relative to the other codes are unaffected; this is the identifier's spelling only. Filing it as a report rather than a PR because a public, documented module-level name (`hwilib.errors` is `automodule`'d) is a compatibility question for the maintainers to decide rather than a drop-in rename: anyone importing `UNKNWON_DEVICE_TYPE` directly would break if it were renamed outright. A non-breaking fix would add the correctly-spelled name and keep the old one as an alias:

```python
UNKNOWN_DEVICE_TYPE = -4 #: Device type is unknown
UNKNWON_DEVICE_TYPE = UNKNOWN_DEVICE_TYPE  # deprecated misspelling, kept for compatibility
```

and fix the docstring reference to `:data:`UNKNOWN_DEVICE_TYPE`` at the same time.

Found while pinning this module's public surface for a subprocess adapter in another project ([btclib#468](https://github.com/btclib-org/btclib/pull/468)), which transcribes the error table as-is and had to record the misspelling rather than "fix" it in its own copy.

fametrano:
Fixes #845.

`UNKNWON_DEVICE_TYPE` (error code -4, in `hwilib/errors.py`) is a transposition of "UNKNOWN". This adds the correctly-spelled `UNKNOWN_DEVICE_TYPE` and aliases the old name to it, rather than renaming outright: it's a public, documented module-level name (the module is `automodule`'d), so an external importer of the misspelled name should not break.

`UnknownDeviceError` now raises with the correct spelling. Its docstring's `:data:`DEVICE_TYPE`` cross-reference is also fixed — it pointed at a name that didn't exist under either spelling, so the Sphinx `:data:` role couldn't have resolved it before this either.

Out of scope for this PR: `hwilib/devices/ledger_bitcoin/errors.py` carries the same misspelling in a comment-credited copy of this file ("Original version: https://github.com/bitcoin-core/HWI"), vendored as part of the `ledger_bitcoin` client subpackage. I didn't touch it, since it's a separate vendored copy and not literally this file; flagging it here in case maintainers want it addressed too, either in this PR or separately.

No behavioural change: both names resolve to the same value, `UnknownDeviceError().get_code()` still returns -4.

### DIFF
--- a/hwilib/errors.py
+++ b/hwilib/errors.py
@@ -16,7 +16,8 @@ from contextlib import contextmanager
 NO_DEVICE_TYPE = -1 #: Device type was not specified
 MISSING_ARGUMENTS = -2 #: Arguments are missing
 DEVICE_CONN_ERROR = -3 #: Error connecting to the device
-UNKNWON_DEVICE_TYPE = -4 #: Device type is unknown
+UNKNOWN_DEVICE_TYPE = -4 #: Device type is unknown
+UNKNWON_DEVICE_TYPE = UNKNOWN_DEVICE_TYPE #: Deprecated misspelling of :data:`UNKNOWN_DEVICE_TYPE`, kept for compatibility
 INVALID_TX = -5 #: Transaction is invalid
 NO_PASSWORD = -6 #: No password provided, but one is needed
 BAD_ARGUMENT = -7 #: Bad, malformed, or conflicting argument was provided
@@ -122,13 +123,13 @@ class DeviceAlreadyUnlockedError(HWWError):
 
 class UnknownDeviceError(HWWError):
     """
-    :class:`HWWError` for :data:`DEVICE_TYPE`
+    :class:`HWWError` for :data:`UNKNOWN_DEVICE_TYPE`
     """
     def __init__(self, msg: str):
         """
         :param msg: The error message
         """
-        HWWError.__init__(self, msg, UNKNWON_DEVICE_TYPE)
+        HWWError.__init__(self, msg, UNKNOWN_DEVICE_TYPE)
 
 class NotImplementedError(HWWError):
     """


hwilib/errors.py
```diff
@@ -16,7 +16,8 @@
 NO_DEVICE_TYPE = -1 #: Device type was not specified
 MISSING_ARGUMENTS = -2 #: Arguments are missing
 DEVICE_CONN_ERROR = -3 #: Error connecting to the device
-UNKNWON_DEVICE_TYPE = -4 #: Device type is unknown
+UNKNOWN_DEVICE_TYPE = -4 #: Device type is unknown
+UNKNWON_DEVICE_TYPE = UNKNOWN_DEVICE_TYPE #: Deprecated misspelling of :data:`UNKNOWN_DEVICE_TYPE`, kept for compatibility
```
Sjors, line 20: Compatibility with what?

achow101:
ACK 6e8cecafadd78d3f78ef1eaad353725b0c21348f

hwilib/errors.py
achow101: Presumably anyone that's using this as a python library and not a command line tool.

PR #794, opened 2025-07-21
MuSig2

Sjors:
This branch includes (as merge commits):
- #851 
- #852 
 
The description below covers only the additional changes.

Developed and tested together with [Sjors/bitcoin#91](https://github.com/Sjors/bitcoin/pull/91).

## MuSig2 signing

Add two-round MuSig2 signing for the Ledger Bitcoin app 2.5.0. HWI first returns the device’s public nonce and, once the other participants’ nonces are available, returns its partial signature. MuSig2 data contributed by other participants is preserved in the PSBT between calls.

Coldcard Edge is connected to the same device-test flow. Its PSBT handling now recognizes MuSig2 fields as signing data, allowing the PSBTv2 signing rounds to remain intact.

## Tests

The common device test exercises a registered policy with a MuSig2 key path and a delayed recovery leaf. It covers:

- funding the policy;
- both MuSig2 signing rounds;
- finalization, broadcast, and confirmation;
- spending through the recovery script path.

This runs for Ledger and Coldcard Edge.

A separate interoperability test exercises a no-fallback 3-of-3 MuSig2 policy between Ledger, Coldcard Edge, and Bitcoin Core. CI builds the companion Bitcoin Core branch and runs the scenario through HWI’s CLI interface.

Unit tests cover Ledger signing-round selection and MuSig2 signing results.

## Documentation

Add a walkthrough for a 2-of-2 wallet consisting of a hardware-wallet participant and a hot key managed by Bitcoin Core. It covers obtaining the BIP87 account key, registering the descriptor, displaying an address, and completing both signing rounds.

PR #847, opened 2026-08-12
[WIP] Drop Coldcard Multisig Patches

achow101:
Makes the multisig tests use only a single device in the multisig. Having the device in the multisig multiple times was originally a convenience, not a requirement.

This should let us drop the coldcard multisig patches. Possibly that will require registering a multisig policy (not implemented yet, waiting for CI) hence basing this on #842.


Sjors:
From the commit message of b9c94b6c7fd860659b175ad303b3d6909a13da4d:

> It's not a realistic test case to have the same device participate in a multisig multiple times.

It _is_ when there's a recovery path, e.g. `(A && B) || (after(1 year) && A)`. What's not realistic is `A == B`. Hopefully the former also doesn't require a patch?

achow101:
> It _is_ when there's a recovery path, e.g. `(A && B) || (after(1 year) && A)`. What's not realistic is `A == B`. Hopefully the former also doesn't require a patch?

That's miniscript, not multisig.

Sjors:
Sure, but does that mean we need to bring these patches back to support miniscript? (without which, MuSig2 is a lot less useful)

achow101:
> Sure, but does that mean we need to bring these patches back to support miniscript? (without which, MuSig2 is a lot less useful)

Probably not. We'd have to register the descriptors and not reuse the same device, but that's not that difficult to do.

Sjors:
Needs rebase :-)

I guess that's acceptable for a simulator test, as long as it works with the real one. It would just make the test scenario a bit more complicated. I'll give my MuSig2 stack a rebase to see the impact.

Sjors:
Concept ACK

Assuming you can find a way to make the tests pass here...

I added ColdCard support to #794, for now including the multisig patch (and a variant for edge). It turns out the Edge firmware already allows xpub reuse across taproot leaves, so  I don't actually need the patch (after this PR).

achow101:
Nevermind. Getting this to work is getting too complicated.

PR #857, opened 2026-09-07
descriptor: fix two parse-error messages and cover them with tests

fametrano:
Two error messages in `hwilib/descriptor.py`, now fixed and covered by tests:

- the bare-multisig error printed a literal `{}` because its `.format(len(pubkeys))` call was missing — it now reports the actual count (e.g. "Cannot have 4 pubkeys in bare multisig: only at most 3 pubkeys");
- the `wpkh` branch reported "More than one pubkey in **pkh** descriptor", copied from the `pkh` branch — now "wpkh".

`test/test_descriptor.py` adds two regression tests that exercise these two paths through `parse_descriptor`:

- `test_wpkh_rejects_multiple_pubkeys` — asserts the `wpkh` message names `wpkh`;
- `test_bare_multisig_too_many_pubkeys_reports_count` — asserts the count is interpolated.

Both fail against the pre-fix messages and pass after the fix. This mirrors the equivalent electrum change ((spesmilo\/electrum#10947, merged 2026-09-09)).


waqas878678-source (commented):
hwilib/descriptor.py

fametrano:
Added two regression tests covering both fixed messages, mirroring the equivalent electrum change (spesmilo/electrum#10947). Both fail on the pre-fix messages and pass after the fix; `test/test_descriptor.py` goes from 16 to 18 tests.

fametrano:
The electrum twin, spesmilo/electrum#10947, was merged on 2026-09-09 (87121026, by ecdsa): same two messages, same two tests.

PR #859, opened 2026-09-09
descriptor: validate public keys when parsing

fametrano:
A raw-hex public key in a descriptor is never checked to be a valid point, so `pkh(<64 hex chars that are not a curve point>)` parses. Bitcoin Core rejects an invalid key at parse time.

This validates a raw-hex key against the acceptance set Core uses in `ParsePubkeyInner`, context-aware:

- inside `tr()`: a 32-byte x-only key or a 33-byte compressed key;
- elsewhere: a 33-byte compressed key, and a 65-byte uncompressed key only at the top level or under `sh()` (rejected under `wpkh()`/`wsh()`, per BIP141);
- hybrid (`0x06`/`0x07`) and off-curve or wrong-length keys are rejected everywhere.

`hwilib/key.py` gains two small on-curve predicates (`point_on_curve`, `x_coord_on_curve`), since the existing point helpers recover a coordinate without checking the curve equation. Extended-key (xpub) expressions are unaffected. Tests cover the acceptance and rejection cases.

The equivalent change for Electrum's descriptor parser, which shares this code, is spesmilo/electrum#10951.


fametrano:
The one red check on this PR is `test-bitbox01 / Python 3.12 bitbox01 cli`, and it isn't caused by this change: the job failed before any test ran, because bitcoind never started:

```
Error: Unable to bind to 0.0.0.0:44435 on this computer. Bitcoin Core is probably already running.
Error: Failed to listen on any port. Use -listen=0 if you want this.
  File "/__w/HWI/HWI/test/./run_tests.py", line 124, in <module>
    bitcoind = Bitcoind.create(args.bitcoind)
RuntimeError: bitcoind failed with exit code 1
```

That looks like a port collision starting the daemon, not a descriptor-parsing failure. There's a single run on this head (`b5b8e25c`, run 34319824959, attempt 1), and it has never been retried, so the red has been sitting there since September 9, 2026. #857 and #858, which touch the same file, are green on all 294 jobs.

Could someone re-run that job, and take a look at the change itself? The branch is even with master, so there's no rebase that would produce a fresh run, and I can't trigger one myself.

